PR #7, opened 2016-04-02
Proposed fix for issue 6

bcpierce00:


bcpierce00:
This is easy to roll back if people propose a better fix, so I'm going to go ahead and merge it now...

PR #10, opened 2016-04-13
Preliminary Ocamlbuild support.

### LINKED ISSUE #9: Unison fails to find proper lablgtk path under an OPAM environment
Under an OPAM system, Unison fails to find the proper lablgtk2 path, as this recipe:

```
OCAMLLIBDIR=$(shell ocamlc -v | tail -1 | sed -e 's/.* //g' | tr '\\' '/' | tr -d '\r')
```

doesn't quite work under OPAM as it uses a different library layout.

My preferred solution to locate Ocaml dependencies using `ocamlfind`, however that would add a new dependency to Unison's build.

If you'd be willing to accept that new dependency I'd be happy to provide a patch, another solution would be to detect if `ocamlfind` is installed and use it, however, my Make-foo is not so good to implement that reliable.

ejgallego:
Hi, this patch adds (optional) ocamlbuild support to unison. Do
```
$ make ocb
```
in `src/` to build unison using ocambuild. This fixes #9 as now Unison can be build in an OPAM environment.

IMO it is not ready to merge yet, so I'm using this PR for review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,7 @@ TAGS
 build
 logmsg
 xcuserdata
+src/_build
+src/linkgtk2.native
+src/linktext.native
+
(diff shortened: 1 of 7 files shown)


bcpierce00:
I think we've agreed that it's better to integrate findlib support directly into the existing makefile, so I'm going to close this pull request.

PR #25, opened 2016-06-22
Change logfile location to .unison

eugenesan:
Change logfile location by prepending .unison to name of logfile in ubase/trace.ml

Bug-Debian: http://bugs.debian.org/484997
Bug-Ubuntu: https://bugs.launchpad.net/bugs/234365

### DIFF
--- a/src/ubase/trace.ml
+++ b/src/ubase/trace.ml
@@ -113,10 +113,10 @@ let logging =
 
 let logfile =
   Prefs.createFspath "logfile"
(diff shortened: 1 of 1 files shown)


bcpierce00:
I have no problem with this change (though it would be better to use fileInUnisonDir, which is more portable), but I can imagine there might be people out there with scripts that will break if the default changes.  Probably a good idea to ask on the unison-users list if there are a lot of people in that situation...  I'll close this pull request for now.

PR #46, opened 2016-11-05
In GTK interface, show full pathnames in file list

dnlopez:
A small code change to show full pathnames in the file list, instead of just an indented filename. It's always seemed to me like the indentation is supposed to suggest the shape of the directory hierarchy - ie. with subfolders and files indented more than their parents - but the exact amount of indentation has always looked erratic to me, and if you change the sorting of the list things get extra confusing.

So, I've changed my copy to just show the full paths. When I say 'full path', I mean relative to the Unison root. Here's a screenshot of what it looks like now:

unison_with_full_paths

### DIFF
--- a/src/uicommon.ml
+++ b/src/uicommon.ml
@@ -291,7 +291,7 @@ let reconItem2stringList oldPath theRI =
       (replicaContent2shortString diff.rc1,
        direction2action partial diff.direction,
        replicaContent2shortString diff.rc2,
-       displayPath oldPath theRI.path1)
+       Path.toString theRI.path1)
 
 let reconItem2string oldPath theRI status =
   let (r1, action, r2, path) = reconItem2stringList oldPath theRI in


bcpierce00:
Nice.  But I wonder whether it would be best to put this under control of either a preference or (better?) a button on the display, or both...?

dnlopez:
If it was a button then I would hope for a preference/option too so I could make it persistent, because I don't see myself ever preferring the indented view, as it currently appears.

I just took another look at the indented view to make myself understand how it works:
unison_2_indented_sortdefault

It seems there is a fixed amount of indentation, about 10 pixels maybe, for each descent into a subfolder. So the second row, 'elpa/...' is 10 pixels indented because 'elpa' is directly inside 'emacs.d'. Where then, is 'test.txt' on the next line? By looking at that indent, I find it hard to judge for sure (in fact, it's inside 'gnu').

On the next two lines, where are 'marmalade' and 'melpa'? They're almost though not quite lined up with the end of 'elpa/' above, so I find my first impression is that they're directly inside that folder, but actually the real story is that there's a 20 pixel indent from the start of 'elpa', which means they're two folders below - ie. they're inside 'archives'.

This is the first time I've understood how the indentation works, and it's worth noting that I never absorbed this from casual use - only by stopping just now to puzzle it out. I would frequently resort to just highlighting the line and looking at the full path in the bottom pane to confirm where a file actually was. Now that I understand it, I reckon I would still find the indentation hard to read - it's hard to judge those distances by eye.

I had assumed, thinking it would be logical, that the subfolders at the same level were supposed to line up, and if someone just switched that column to use a monospace font, it would look like this:
```
.emacs.d/.#ac-comphist.dat
         elpa/archives/gnu/archive-contents.signed
                           test.txt
                       marmalade/archive-contents
                       melpa/archive-contents
```
I initially aimed for that, but I struggled to find how to change the font in just that column (I'm not very good in either Gtk or Ocaml, unfortunately). That's when I changed tack to just showing the full path on every line, which turned out to be a simpler task. That was just as well, since now that I understand the indentation, I realise that a monospace font alone wouldn't have brought the kind of alignment I was looking for. I find I'm satisfied with the full path view, and I imagine I'd still prefer it to the indented view of my text example above. When I see the following, I just feel relief:

unison_2_fullpath_sortdefault

The higher volume of text doesn't bother me, and personally I'm going to be using it like that from now on.

bcpierce00:
On further reflection, I'm actually persuaded that this change is probably good for all users and doesn't need a switch.  So I'm going to go ahead and accept it.  If people disagree with this choice, please let me know.

bcpierce00:
(Monospace font is a different issue -- I don't have a good sense of whether most people would prefer this.)

acolomb:
I sometimes prefer the current display mode with only the indented base file name. It's a rather neat feature of the unison GUI which I'd miss if it were removed completely. Makes it rather easy to spot names of files which I use regularly.

Having an option to show the full path names sounds neat. In a perfect world it would be a toolbar toggle button to switch instantly.

Please keep in mind that for large changesets, the full path display will use much more memory for the TreeModel. That could become a serious problem for some people on e.g. single-board computers.

bcpierce00:
I don’t object to adding a switch for this if someone wants to implement it (even if I’d expect people on single-board computers to use the text interface!). 

    - B

PR #51, opened 2016-12-14
Add config option to allow for a time difference between archive and file system

emmenlau:


alkuzad:
Why 8 seconds? For example I use unison to sync dev pc with server and save every 3s is not unusual. Also why call it "maximum" when in fact it's minimum time delta of files that unison considers?

emmenlau:
Dear @alkuzad , please check the "files changed diff", not the individual commits. The 8 seconds was a first test, but the final version uses a config variable.

About the minimum / maximum, this is the "maximum allowed delta that files may have and still be considered equal". Its not a minimum. Check the condition in line 601: delta must be smaller or equal to maxModificationTimeDifference, so maxModificationTimeDifference is the maximum.

alkuzad:
@emmenlau Hi, sorry but on mobile interface is squeezed and I overlooked the commits over changes, sorry for that. Propably this is why some projects demands rebase before merge...

As to maximum / minimum - from user perspective unison is synchronizer, not equality checker so people look for sync options. If I see maximum time difference (in a synchronize program) I translate it to maximum time difference files has to have to be synced, example: 8 is maximum so files over that will not be synced.

Docs doesn't make it clear also:
"!maximum modification time difference (in seconds)"
("Used to compare files against the archive.")

How they are compared ? This is not clear.

emmenlau:
Hi @alkuzad ok I understand your logic for minimum / maximum. But my thinking was different, so its hard for me to suggest a better name. Maybe something like "allowed time delta" or "modification time delta" or something? Can you suggest something better?

alkuzad:
@emmenlau "modification time delta" would be better, possibly also with commentary:

"If file modification time difference between replicas is less than delta, file won't be synchronized"

emmenlau:
Dear @alkuzad I have updated the parameter and description:
```
  Prefs.createInt "modtimedelta" 0
    "!If file modification time changes by less than delta, file won't be checked for changes"
    ("Used to check if a file may have changed compared to archive.")
```
Is it good? Thanks for checking!

bcpierce00:
I apologize for losing track of this PR.  If you'd still like functionality like this in Unison, I'd like to propose a small change to how it works.  

Telling unison to ignore modtime differences up to one hour seems rather dangerous (can easily lead to missing syncs, though it won't lead to data loss).  Moreover, to address the issue that motivated the patch, it is sufficient to ignore modtime differences of _exactly_ one hour.  

So I'd suggest implementing it like this: make the preference be a _list_ of exact time differences to add to the built-in list.  With this change, I'll gladly accept the PR.

bcpierce00:
(We'd want to implement #5 first, though, to minimize compatibility problems.)

Issue #48, opened 2016-12-05
Unison segfaults on Fedora 25


alexmarkley:
I've built unison version unison-2.49.543 and unison-2.48.4 (from http://www.seas.upenn.edu/~bcpierce/unison/download/releases/ ) on Fedora 25.

I built it using simply "make UISTYLE=text" on both the client side and the server side. Both machines were recently upgraded or wiped/installed to Fedora 25.

When sync'ing a large replica (~1TB), after some time, the client always segfaults:

```
===SNIP===
Shortcut: copied /home/alex/MacBuild/unison/unison-2.48.3/path.ml from local file /home/alex/.unison.Build.7f78c97b4bc1aea68cc67a71dd2bde7f.unison.tmp/unison-2.48.3/path.ml
Shortcut: copied /home/alex/MacBuild/unison/unison-2.48.3/path.mli from local file /home/alex/.unison.Build.7f78c97b4bc1aea68cc67a71dd2bde7f.unison.tmp/unison-2.48.3/path.mli
 52%  90:58 ETASegmentation fault (core dumped)
```

The kernel buffer:

```
[ 6658.933941] unison[2475]: segfault at 0 ip 00000000004e346b sp 00007fff1efc9200 error 4 in unison[400000+121000]
```

Any ideas? I am definitely grasping at straws here.

alexmarkley:
I wonder if this is an issue with ocaml or some other dependency. The segfault is even occurring with the unison-text-240 package included in the Fedora 25 repository.

```
Shortcut: copied /home/alex/pebble-dev/pebble-sdk-4.2-linux64/pebble-tool/pebble_tool/._version.py from local file /home/alex/.unison.pebble-dev.7f78c97b4bc1aea68cc67a71dd2bde7f.unison.tmp/pebble-sdk-4.2-linux64/pebble-tool/pebble_tool/._util
Shortcut: copied /home/alex/pebble-dev/pebble-sdk-4.2-linux64/.env/bin/activate_this.py from local file /home/alex/.unison.pebble-dev.7f78c97b4bc1aea68cc67a71dd2bde7f.unison.tmp/pebble-sdk-4.2.1-linux64/.env/bin/activate_this.py
 62%  174:21 ETASegmentation fault (core dumped)
```

And in the kernel message buffer...

```
[110919.146786] unison[7839]: segfault at 1708 ip 0000000000469387 sp 00007fffb39bf2c0 error 4 in unison-text-2.40[400000+118000]
```

[...]

PR #55, opened 2016-12-22
Fix rare SIGSEGV when transferring large replicas.

alexmarkley:
Patch from here:
https://caml.inria.fr/mantis/view.php?id=7431#c17026

Related issue reports:
https://github.com/bcpierce00/unison/issues/48
https://caml.inria.fr/mantis/view.php?id=7431
https://bugzilla.redhat.com/show_bug.cgi?id=1401759


alexmarkley:
This patch fixes issue https://github.com/bcpierce00/unison/issues/48

PR #56, opened 2016-12-22
Fix a theoretical integer overflow.

alexmarkley:
Patch from:
https://caml.inria.fr/mantis/view.php?id=7431#c16962


alexmarkley:
This recommendation came from one of the OCaml developers. It emerged from an examination of the Unison code motivated by a desire to track down this issue: https://github.com/bcpierce00/unison/issues/48

PR #57, opened 2017-01-09
add Dockerfile

razic:
Hey, I've added a Dockerfile to this repository.

If someone want to use Unison via Docker, it's could be as simple as:

`docker run unison`

This pull-request is a step in the right direction. Granted, you may still want to tag images to specific versions on the Docker Hub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM ubuntu:14.04
+
+RUN apt-get -y update \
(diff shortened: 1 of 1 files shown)


bcpierce00:
I'm not a Docker user, so don't can't evaluate this addition directly, but it's a non-invasive change, so I guess no harm.

PR #71, opened 2017-03-01
Fix messy props when creating parent directory

eliemichel:
This was a bug reported by Sebastian Elsner on the user mailing list:
https://groups.yahoo.com/neo/groups/unison-users/conversations/messages/11740

It seems that `Fs.mkdir` does not correctly apply mode, and anyway file owner
and group were never copied. And furthermore the porp list was not built
correctly so both mode and owner/group were false.

In order to prevent this kind of issue, one should document in which order prop
lists in content descriptions are (parent -> child or child -> parent). I do
not do it myself because I am actually not sure.

### DIFF
--- a/src/files.ml
+++ b/src/files.ml
@@ -377,7 +377,8 @@ let rec createDirectories fspath localPath props =
           createDirectories fspath parentPath rem;
           try
             let absolutePath = Fspath.concat fspath parentPath in
-             Fs.mkdir absolutePath (Props.perms desc)
+             Fs.mkdir absolutePath (Props.perms desc);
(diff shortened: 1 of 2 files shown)


sebastianelsner:
Fixes my problem mentioned in the mailing list! Many thanks, please merge!

eliemichel:
Indeed, the original code was systematically failing. Except maybe in the very unlikely scenario in which the paths are more or less symmetrical (e.g. "foo/bar/baz/bar/foo").

Its writer must have been confused by property lists which are sometimes reversed and sometimes not among the code and whose order when passed as argument of a function is not specified. At least, I was, at first. ;)

PR #93, opened 2017-08-29
ensure sorted profile list in the macOS UI

mroi:
On APFS file systems, `NSFileManager` enumerates directories in an unspecified order. To restore the ordering at the UI level, we have to explicitly sort the profile list.


brabalan:
This looks good to me.

PR #95, opened 2017-09-27
[make] Pass correct include options to gtk2 build under OPAM env.

### LINKED ISSUE #9: Unison fails to find proper lablgtk path under an OPAM environment

ejgallego:
Fixes #9.

Makefile.OCaml hardcoded the GTK include path, which is incorrect
under an OPAM environment. The fix uses `ocamlfind`, if available, to
get the proper cmdline options.

Works for me (TM).

### DIFF
--- a/src/Makefile.OCaml
+++ b/src/Makefile.OCaml
@@ -293,8 +293,14 @@ ifeq ($(UISTYLE), gtk)
 endif
 
 # Gtk2 GUI
+OCAMLFIND := $(shell command -v ocamlfind 2> /dev/null)
+
 ifeq ($(UISTYLE), gtk2)
-  CAMLFLAGS+=-I +lablgtk2
+  ifndef OCAMLFIND
+    CAMLFLAGS+=-I +lablgtk2
+  else
+    CAMLFLAGS+=$(shell $(OCAMLFIND) query -i-format lablgtk2 )
+  endif
   OCAMLOBJS+=pixmaps.cmo uigtk2.cmo linkgtk2.cmo
   OCAMLLIBS+=lablgtk.cma
 endif


src/Makefile.OCaml
```diff
@@ -293,8 +293,14 @@ ifeq ($(UISTYLE), gtk)
 endif
 
 # Gtk2 GUI
+OCAMLFIND := $(shell command -v ocamlfind 2> /dev/null)
+
 ifeq ($(UISTYLE), gtk2)
-  CAMLFLAGS+=-I +lablgtk2
+  ifndef OCAMLFIND
```
gasche: I suppose that the assumption is that `ifndef OCAMLFIND` will *not* take the branch if the `OCAMLFIND` variable exists but is empty, right? (This is what happens with the definition above if ocamlfind does not exist in the PATH). That is a bit counter-intuitive as I would expect (not knowing bash very well) the `ifndef` to mean "if the variable is not defined", not "if it is not defined or empty".

ejgallego: @gasche I took this from [SO](https://stackoverflow.com/questions/5618615/check-if-a-program-exists-from-a-makefile) my make-fu is ⊥, but I think that it should work fine.

ejgallego: Concretely, the definition of `ifdef` in make language states that "If the variable variable-name has an empty value, the text-if-true is effective; otherwise, the text-if-false, if any, is effective. Variables that have never been defined have an empty value"

ejgallego: A key point is the use of `:=` that will expand the variable, as `ifndef` does not.

bcpierce00:
Thanks for the fix!

PR #109, opened 2017-11-28
Update for current Mac UI build. Fixes #47

### LINKED ISSUE #47: OS X build documentation is dated
Correct steps would be (roughly)

- use 'make' for text UI
- use src/uimac14/*.xcodeproj (+ build in xcode) for graphical UI

Current documentation (and also Makefiles) have e.g. references to uimac09, uimacnew, both of which seem to be gone.

alanshutko:


brabalan:
LGTM. We should also remove the old UI code (uimacnew and uimacnew09), if we remove the support for building it.

bcpierce00:
Yes.  Shall we do that as a separate step?

PR #114, opened 2018-01-02
[make] Fix detection of lablgtk2 with OPAM

erikmd:
Dear Unison developers,

I noticed a slight bug with the compilation of Unison in gtk2 mode using OPAM: if lablgtk2 is available in opam's lib folder **only** (not system-wide at the same time), up to now I could not compile Unison's GUI because `make src` behaved like `make text`.

This PR is thus a follow-up of #95.

I tested it under Debian GNU/Linux stable, in both setups (system-wide or opam-based ocaml).

Hoping this helps


bcpierce00:
Thank you!

PR #115, opened 2018-01-09
uitext: Add a "go back" option to the "proceed" prompt

g-raud:
When one mistakenly skips the last item of the "selection" prompt this
option starts again at the last item whereas the existing option "do
again" reprocesses the whole list which can be long.

Add an accumulator to the function interactAndPropagateChanges() that
passes it as a newly added accumulator of the function interact() which
simply passes it as the accumulator of its subfuntion loop().  In
interactAndPropagateChanges() additionally append the accumulator to the
main list when we want to go back to the selection.

### DIFF
--- a/src/uitext.ml
+++ b/src/uitext.ml
@@ -225,7 +225,7 @@ let displayri ri =
 
 type proceed = ConfirmBeforeProceeding | ProceedImmediately
 
-let interact rilist =
+let interact pilist rilist =
   let (r1,r2) = Globals.roots() in
   let (host1, host2) = root2hostname r1, root2hostname r2 in
   if not (Prefs.read Globals.batch) then display ("\n" ^ Uicommon.roots2string() ^ "\n");
@@ -378,8 +378,8 @@ let interact rilist =
                 ]
                 (fun () -> displayri ri)
   in
-    loop [] rilist
-
+    loop pilist rilist
+    
 let verifyMerge title text =
   Printf.printf "%s\n" text;
   if Prefs.read Globals.batch then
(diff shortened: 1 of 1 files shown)


g-raud:
You should consider picking the other commits from the branches 'proceed-goback' or 'proceed-goback-2.48' of my fork (https://github.com/g-raud/unison/tree/proceed-goback or https://github.com/g-raud/unison/tree/proceed-goback-2.48) that slightly simplify the code touched by this pull request (hopefully without introducing bugs) and add another useful menu option.

Should I create another pull request with it?

bcpierce00:
Could you make a new pull request that shows the best version of all the changes you're proposing?  I'm getting a bit confused looking at three separate sets of commits...

g-raud:
I would say the commits in the pull branch already are the best
versions I made.  However their ordering is wrong and some should be
dropped.

- uitext: Add a "go back" option to the "proceed" prompt
This is the minimal change to do what this pull request is about.

- uitext: hide the accumulator of interactAndPropagateChanges()
This is a very simple change with a lot of reindentation (to view that
call git diff --ignore-space-change) related to the minimal change in
that it hides a newly created accumulator by defining a subfunction.
This commit is optional and could be dropped because it adds
indentation to an already complex function and because the new
subfunction (as it is written) depends on the reconItemList argument
of interactAndPropagateChanges() (although I hope the compiler
would see that this dependency can be optimized away).  If the change
is dropped I should add a comment explaining that a non-empty
accumulator is only to be passed inside the function itself.

- uitext: add a "stop" option to the "selection" process
This change is rather independant from this pull request but has similar
advantages: it avoids to go through all the selection process again each
times one wants to tweak the selection from the top.

- uitext: in interactAndPropagateChanges() reuse newReconItemList
This change is independant from the pull request and was proposed only
because I came across a surprising behaviour of not re-using the
returned newReconItemList but instead the original reconItemList.  The
change has consequences however so it should be dropped from this pull
request.  Do you know why the newReconItemList is not re-used when we go
through the selection process again?  For example in doing so the ignore
patterns added by the command "I" are unignored when one starts the selection
process again; is this a bug and should I open another issue/pull request?

I will rewrite the pull branch history according to what you prefer.  Should I keep
the commit hiding an accumulator?

g-raud:
The other branches mentioned in a previous message are either backports to unison-2.48 or the pull branch itself; at the time I did not know yet that one could rewrite a pull branch of an unmerged pull request.

bcpierce00:
We can keep all your changes (with the last one in a separate PR, if you find that clearer).  I agree that the last one looks like a bug.  I'm going to close this PR for the moment so that I don't get confused; please re-open it (and/or another) when you're satisfied.

g-raud:
This is reopened as #138 as I cannot reopen this PR.

Issue #123, opened 2018-01-17
uitext: uncaught exceptions while reading from terminal


g-raud:
With unison 2.48.3-1 from Debian stretch several exceptions raised inside the function Uitext.getInput are uncaught by Unison that fails abruptly (without cleaning up and without printing a specific error message).

There are (at least) two different calls that can raise an uncaught exception: input_line() can raise End_of_file and Unix.read can raise Unix.Unix_error.

I noticed the second case by having a condition that I cannot reproduce (probably a problem with terminal input because 'rm -i filename' woud fail with message "cannot close file" and nano would fail too).  Here is what unison -ui text printed:
```
...
local          remote
new link ---->            subdir/subdir/file  [f] Uncaught unix error: read failed: Resource temporarily unavailable
Raised by primitive operation at file "unix.ml", line 248, characters 7-33
Called from file "/tmp/buildd/unison-2.48.3/uitext.ml", line 114, characters 16-42
Called from file "/tmp/buildd/unison-2.48.3/uitext.ml", line 120, characters 14-27
Called from file "/tmp/buildd/unison-2.48.3/uitext.ml", line 180, characters 6-17
Called from file "/tmp/buildd/unison-2.48.3/uitext.ml", line 545, characters 35-57
Called from file "/tmp/buildd/unison-2.48.3/uitext.ml", line 718, characters 6-47
Called from file "/tmp/buildd/unison-2.48.3/uitext.ml", line 788, characters 6-90
Called from file "/tmp/buildd/unison-2.48.3/uitext.ml", line 810, characters 19-66
Called from file "/tmp/buildd/unison-2.48.3/uitext.ml", line 870, characters 21-43
```

The first case is raised simply by redirecting stdin to /dev/null:
```
...
local          remote
new file ---->            file  [f] Uncaught exception End_of_file
Raised at file "pervasives.ml", line 336, characters 20-31
Called from file "/tmp/buildd/unison-2.48.3/uitext.ml", line 106, characters 14-30
Called from file "/tmp/buildd/unison-2.48.3/uitext.ml", line 180, characters 6-17
Called from file "/tmp/buildd/unison-2.48.3/uitext.ml", line 545, characters 35-57
Called from file "/tmp/buildd/unison-2.48.3/uitext.ml", line 718, characters 6-47
Called from file "/tmp/buildd/unison-2.48.3/uitext.ml", line 788, characters 6-90
[...]

PR #124, opened 2018-01-17
Uitext: catch exception while reading stdin and print error

g-raud:
This change is related to issue #123 but it only prints an error message without treating the error condition (and the exception is raised again and uncaught).


g-raud:
The new commit named "Uitext.selectAction: catch some exceptions then raise Util.Fatal" fixes #123.

PR #126, opened 2018-01-22
Decrement major version

bcpierce00:
Changing it to 2.51 a while back was a mis-step -- there's no need.


davewhipp:
I'm totally new to building unison from source, but I'm wondering why this was rolled back to being version 2.50 and 2.50.1. I just today compiled unison for both MacOS and Ubuntu. The Ubuntu binary lists the version as 2.51, whereas the MacOS version has 2.50. In order to sync, I finally settled on  editing the Makefile.ProjectInfo file to use version 2.51 in order to have compatibility between machines. Is there a better solution?

Any many thanks for creating an excellent file sync tool. I use it daily, and would probably never have built from source if not for having a new Mac laptop with APFS file system.

bcpierce00:
Sorry for not including an explanation here.  I had bumped from 2.50 to 2.51 a while back, but realized later that there was actually no change in the set of preferences, so no need to introduce a version incompatibility, unless there were already more than a couple folks out there running 2.51.  If there _are_ (and your message suggests there may be) then undoing the bump was a bad idea and we should put it back to 2.51.

bcpierce00:
P.S.  Your workaround is fine, in the meantime.

davewhipp:
OK, that makes sense. I'm happy with the workaround for now, and probably only represent a small fraction of the user base (Mac users running the latest version of MacOS *and* using APFS). Perhaps I'll post some installation instructions online in case others encounter the same kinds of issues I found with APFS (and that were noted in the issue tracker in this project). More than anything, it's good to help me remember how I got things working again in case I encounter this issue again :smile:. Thanks for the prompt reply!

bcpierce00:
I've changed it back to 2.51 now

PR #138, opened 2018-01-28
Uitext: add commands to "stop" to "proceed" and "go back" to "reconcile" selection

g-raud:
These new commands avoid to have to go through the entire "reconcile" selection process to change only a few paths at the top or at the bottom.

PR #140, opened 2018-01-28
Uitext: add sort commands

g-raud:
The new sort commands are similar to the ones available in the GTK2 UI, plus a "reverse sort".


g-raud:
This PR depends on the still unmerged PR #138 thus it still contains all the commits part of #138 (or part of its unmerged dependencies).

PR #143, opened 2018-01-28
Uitext: add some prompt actions

g-raud:


bcpierce00:
This seems to conflict:
```
<<<<<<< uitext-more-actions
                 (["R"],
                  ("reverse the list"),
                  (fun () ->
                     newLine();
                     loop rest (ri::prev)));
=======
                 (["s";"n"],
                  ("stop the selection"),
                  (fun() ->
                     newLine();
                     (ConfirmBeforeProceeding, Safelist.rev_append prev ril)));
>>>>>>> master
```
Also, I think the description could be clearer: "stop the selection" doesn't mean much to me.

g-raud:
This PR depends now on PR #146.

g-raud:
I will replace or revert commit "Uitext: allow the "follow" action to revert to Unison's recommendation".  I still have doubts that the default action (key "\n": do nothing but stop at conflicts) is the best but the commit change is certainly wrong to make "revert" the default.

I could define an action "next" (do nothing and go down in the list, even when the current item is a conflict) as the default (i.e. "\n").  Has anyone an opinion about this?

g-raud:
I just updated the pull branch.  The new one does not change the default
action.

PR #152, opened 2018-02-03
Uitext: some more actions and tweaks

g-raud:
This PR depends on PR #151.

PR #154, opened 2018-02-05
Uitext: add RPN boolean operations on matching conditions

g-raud:
This PR depends on PR #152.


bcpierce00:
Thank you for all these improvements!!

PR #150, opened 2018-02-02
New preference 'mergenot'

g-raud:
The preference `-batch` does not prevent `-merge` programs from running (because those could be interactive or not).  There used to be a preference `-mergebatch` that could be set (on the command line as well as in a profile) to disable unwanted merge programs (or to override them) in batch mode.   At present the merge programs have to be disabled on the command line by something like `-merge Name * -> true`:
```
$ unison -root root1 -root root2 -merge 'Name * -> mymergecmd' -batch -merge 'Name * -> echo SKIP'
Unison 2.51.2 (ocaml 4.02.3): Contacting server...
Looking for changes
Reconciling changes
new file <-M-> new file   file2  
root1        : new file           modified on 2018-02-02 at  8:43:17  size 2         rw-r-----
root2        : new file           modified on 2018-02-02 at  8:43:17  size 2         rw-r-----
Propagating updates
UNISON 2.51.2 (OCAML 4.02.3) started propagating changes at 08:45:56.02 on 02 Feb 2018
Merge command: echo SKIP
Merge result (exited (0)):
SKIP
100%  00:00 ETASKIP
No outputs detected 
No output from merge cmd and both original files are still present
100%  00:00 ETAFailed [file2]: Merge program didn't change either temp file
UNISON 2.51.2 (OCAML 4.02.3) finished propagating changes at 08:45:56.02 on 02 Feb 2018
Saving synchronizer state
Synchronization incomplete at 08:45:56  (0 items transferred, 0 skipped, 1 failed)
  failed: file2
```
Setting the preference `-mergenot` is shorter, clearer and produces a better log:
```
$ unison -root root1 -root root2 -merge 'Name * -> mymergecmd' -batch -mergenot 'Name *'
Unison 2.51.2 (ocaml 4.02.3): Contacting server...
Looking for changes
Reconciling changes
new file <-?-> new file   file2  
root1        : new file           modified on 2018-02-02 at  8:43:17  size 2         rw-r-----
root2        : new file           modified on 2018-02-02 at  8:43:17  size 2         rw-r-----
No updates to propagate
Synchronization complete at 08:46:02  (0 item transferred, 1 skipped, 0 failed)
  skipped: file2 (contents changed on both sides)
```

PR #155, opened 2018-02-06
Negative Patterns in Path Specifications

g-raud:
This change introduces the new pattern prefixes `del Regex`, `del Name`, `del Path` and `del BelowPath` that override (remove) pathspecs matched by previously given (positive) pathspec patterns, and that can be themselves overriden by following (positive) pathspec patterns.  The prefix `del` was chosen in preference to `not ` because the pattern only applies to preceding matching pathspecs in contrast with negative `not`-options that apply to all matching pathspecs of the corresponding positive option.

This is an alternative to the new preference `mergenot` from PR #150: `-merge del Name *` at the end of the command line has the same effect as `-mergenot Name *` (anywhere).

However this change allows much more than that.  For example to match only certain paths at the root of a "root", there is the preference `path` but unfortunately it does not take pathspec patterns but explicit names.  An alternative with patterns is the preference `ignore` with `-ignore Path *` overriden by `-ignorenot Path prefix-*`; however an `ignorenot` cannot be canceled anymore; with this change it can be by `-ignorenot del Path prefix-name`; or an alternative can be employed: `-ignore Path *` and `-ignore del Path prefix-*` followed by `-ignore Path prefix-name`.


g-raud:
Having "overridable" pathspecs makes it possible to split profiles in files that can be reused easily in several profiles.

g-raud:
I noticed that when the pattern `BelowPath` was added, at the same time a change in the version number was made; I don't know if it was intentional or not.

This change does not do that, because the pattern processing is backwards compatible  (in the sense that when the new patterns are not employed, the behaviour is exactly as before the change) and because a server without this change is still compatible as it does not even detect the new pattern if it is not sent to it.

PR #157 makes clearer that the server failed while attempting to parse an incompatible pattern.  It won't help in the present case but could in a future preference change.

bcpierce00:
I like this approach better than the "mergenot" approach, but I'm still not clear that we can adopt it without bumping the version number.  If a client that knows about these changes talks to a server that does not, it is possible that they will end up considering a different set of paths, and I am not sure whether the behavior is guaranteed to be safe in this case.

g-raud:
If the client sends the new patterns to a server that does not know about them, my tests suggests that the server always fails with exception Prefs.IllegalValue while recompiling the regexes (exception that PR #157 makes the client display without the full backtrace).  The backtrace is at the end of this message if that helps.

I do not fully understand the client-server code but I am nonetheless conviced that an older server cannot behave badly. Here's why.

The Pred.t type has a string list of patterns and a compiled regex (the other fields make no difference or the same following reasoning applies).  Either the sever receives the compiled regex and it uses it: in that case it will behave as the client and that's fine.  Or the server attempts to recompile the patterns into a regex: in that case it fails with the exception Prefs.IllegalValue so there is no risk in this case either.  If the server does something else I would consider it crazy, and not increasing the version number is a good chance to detect that craziness.

Here is the backtrace of a failing server:
```
Unison failed: Uncaught exception Prefs.IllegalValue("bad pattern: DelName *\nA pattern must be introduced by one of the following keywords:\n    Name, Path, BelowPath or Regex.")
Raised at file "/tmp/buildd/unison-2.48.3/pred.ml", line 92, characters 26-60
Called from file "/tmp/buildd/unison-2.48.3/pred.ml", line 73, characters 6-947
Called from file "/tmp/buildd/unison-2.48.3/ubase/safelist.ml", line 63, characters 22-25
Called from file "/tmp/buildd/unison-2.48.3/ubase/safelist.ml", line 67, characters 23-36
Called from file "/tmp/buildd/unison-2.48.3/pred.ml", line 164, characters 2-23
Called from file "/tmp/buildd/unison-2.48.3/globals.ml", line 264, characters 2-26
Called from file "/tmp/buildd/unison-2.48.3/update.ml", line 2026, characters 34-61
Called from file "/tmp/buildd/unison-2.48.3/ubase/safelist.ml", line 92, characters 18-28
Called from file "/tmp/buildd/unison-2.48.3/update.ml", line 2023, characters 4-480
Called from file "/tmp/buildd/unison-2.48.3/update.ml", line 2048, characters 18-66
Called from file "/tmp/buildd/unison-2.48.3/remote.ml", line 764, characters 4-24
Called from file "/tmp/buildd/unison-2.48.3/lwt/lwt.ml", line 75, characters 20-23
Re-raised at file "/tmp/buildd/unison-2.48.3/lwt/lwt.ml", line 135, characters 12-13
Called from file "list.ml", line 73, characters 12-15
Called from file "/tmp/buildd/unison-2.48.3/lwt/lwt.ml", line 31, characters 2-37
Called from file "/tmp/buildd/unison-2.48.3/lwt/lwt.ml", line 83, characters 17-46
Called from file "/tmp/buildd/unison-2.48.3/lwt/generic/lwt_unix_impl.ml", line 147, characters 6-40
Called from file "/tmp/buildd/unison-2.48.3/main.ml", line 202, characters 6-24
Called from file "/tmp/buildd/unison-2.48.3/main.ml", line 131, characters 4-9
```

bcpierce00:
OK, that's not as bad as it could be.  :-)

But I wonder if there's a way for the client to detect when the server fails in this way.  (I suspect no, in which case we should at least clearly document the fact that this can happen, and that these options should not be used with older servers.)

g-raud:
The only way I see to detect an older server would be to parse what the server prints.

I don't think documenting it is necessary as the current version is beta according to the homepage and the next stable release would have a version number incompatible with the previous stable release (and perhaps with the current beta version).  Documenting it would be good if the negative patterns are backported (to a stable version).

bcpierce00:
Good point -- yes, we get the server's version number back when we first connect.  It would be pretty easy to check it and give a sensible error message at some point in the startup sequence.

My plan is actually _not_ to bump the major version number when we promote the beta release to stable.  Since Unison is not very good at inter-version compatibility (see #5), every time it gets bumped a lot of people have problems.

g-raud:
I added a commit that appends a warning about failures potentially caused by negative pathspec patterns.

I also changed the source code to use the sign variants with a regex argument.

g-raud:
I just pushed a more elegant implementation.

g-raud:
I changed the prefix from `Del` to `del ` (with a space); it makes the profile files more readable I think.

brabalan:
I'll have a look at this, but it will have to wait till the end of the month.

bcpierce00:
Sorry -- I merged this before Alan and I noticed that the handling of version incompatibilities is not quite right.  We are reverting it and will try to reopen this one for further discussion.

bcpierce00:
Actually, re-opening the PR seems not to be possible.  However, @brabalan or I can create a new PR later with the same contents (or you can, @g-raud), if we find a good way to resolve the issue.

The issue we are concerned about is that, at the moment, the message about "maybe the server can't deal with these new patterns" will appear _every time_ the client loses the connection with the server, whether this was the cause or not.  Since major version number changes happen very rarely (there may never be another :-), this seems unfortunate.  

More unfortunately, we're not sure what a better fix for this issue should look like.  What we _want_ is for the client to be able to check whether the server supports these extensions and not allow them to be used if it does not.  We should be able to do this by looking at the patchlevel of the server.

PR #157, opened 2018-02-08
Improve server failure message

g-raud:
This makes unison print "Unison server failed" when it is the server that failed.

It also makes the client catch an exception Prefs.IllegalValue from the server (even if the exception is normally avoided by incompatible version numbers).

### DIFF
--- a/src/main.ml
+++ b/src/main.ml
@@ -131,7 +131,7 @@ let catch_all f =
     f ();
     (* Util.msg "Done catch_all...\n"; *)
   with e ->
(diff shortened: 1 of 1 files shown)


g-raud:
> @bcpierce00 requested changes on this pull request.
> 
> No need for nested try blocks -- a single try block can have multiple handlers
> (separated by |)

I know.  Here the first catch transforms the exception and the second
one catches it (plus all others) to print it.  So nested try blocks
avoid code duplication in this case.

bcpierce00:
Ah, ok.

PR #165, opened 2018-03-01
Fixed String Pathspec Patterns

g-raud:
This PR depends on PR #155; merging the latter one first would mke the history clearer.

This change improves globbing pattern documentation, makes the parser fail when the any pattern starts by a slash and adds patterns matching fixed strings: `NameString`, `String` and `BelowString` that take a string optionally double quoted.  Example:
```
  mergenot = String 'path/odd * 'file' , {name}   '
```


g-raud:
The pattern `String` can be useful when generating profile files programmatically because it allows not to reimplement `Uicommon.quote()`.

g-raud:
I just updated the branch.  Now the quotes around a fixed string pattern are single quotes.  This way double quotes can be used to implement C-style escapes to be able to specify filenames containing newlines.

hjmjohnson:
@g-raud  Conflicts now occur.  Is this PR still relevant, or have these changes been incorporated in another way?

L3P3:
@g-raud would be nice to be able to close this PR. Could you do the merging or can this PR be closed?

g-raud:
PR #155 that has been merged has been removed afterwards if I remember correctly, so that it does not make much sense to merge this one now.  The reason for not accepting PR #155 was because of (potential in my opinion) non backward compatibility.  I still think PR #155 should be merged beacause it is a very beneficial improvement for unison that cannot introduce backward incompatibilities related to the protocol since it does not modify the preferences that are passed to the unison server (only if one uses the new config file "patterns" then an older unison (frontend) will fail and even then without any risk of data loss or of cryptic error message).

If you wish you can close this PR now and I will reopen the related PRs this summer when time permits.

gdt:
I am just barely keeping up as maintainer, and my preference is for PRs to be either marked draft if they are being actively worked on, or open in a condition which is ready to merge (rebased etc.) and should  be merged in the view of the submitter, or closed and resubmitted when ready.   I very much look forward to you submitting things agan as you have time; my desire to close is just about reducing cognitive load on my end when evaluating 'what should I do to the repo".

PR #171, opened 2018-03-03
Include options for the command line

g-raud:
This change adds the command line options `-include` and `-source`.  They are not defined as new preferences but as extra parser spec.  They appear in the usage message.

These options can be used to compose or tweak profiles on the command line.


g-raud:
This PR now depends on PR #167.

PR #174, opened 2018-03-12
New preference mergenot

g-raud:
This preference is client-local so it does not create incompatibilities.

One use is to disable all merges on the command line with `-mergenot "Regex .*"`, for example to avoid interactive merges in batch mode.

Another use is to selectively disable merge options in a profile after generic merge patterns at the start of the file.


hjmjohnson:
@g-raud Please rebase on master branch to ensure that this is still relevant and clean.

PR #176, opened 2018-03-13
Pred: fix detection of empty pathspecs

g-raud:
The previous test would never match anything.


g-raud:
Hello, this is approved but still pending.

PR #178, opened 2018-03-14
Assoc only Pathspec Patterns

g-raud:
A "pattern" `assoc <PATTERN>` to record the associated string without setting the preference.  Example:
```
merge = assoc Name *.txt -> mergecmd CURRENT1 CURRENT2                                                                             
merge = Path prefix*.txt
```

The patterns `add <PATTERN>` equivalent to `<PATTERN>` are also added for completeness but left undocumented.


g-raud:
This PR depends on PR #165 (Fixed String Pathspec Patterns).

PR #181, opened 2018-03-17
Pred: parse a single mapSeparator separated by spaces

g-raud:
This PR depends on #176 (Pred: fix detection of empty patterns).

With this change a pathspec can contain the separator if a (possibly dummy) associated string is appended.  For example to specify a fixed string pattern (PR #165):
```
ignore = String "downloads/ -> strange filename <- " ->
```

A separator not surrounded by spaces is not recognized anymore to make it less likely to accidentally have it appear in a pathspec.  This should be backward compatible as the manual shows only examples of a separator surrounded by spaces.  It is also what the preference rootalias requires.  A separator followed by an empty string is recognized but this is wittingly not documented.


src/pred.ml
```diff
@@ -61,16 +61,18 @@ let mapSeparator = "->"
...
+        (* Actually find "(^| )mapSep( |$)" (by surrounding [clause] by spaces
+           possibly removed by previous trimming) to detect an empty pattern
+           and/or an empty string *)
+      ("", _)     -> raise (Prefs.IllegalValue "Empty pattern")
+    | (p, None)   -> (p, None)
+    | (p, Some v) -> (p, Some (Util.trimWhitespace v)) in
   let compiled =
     begin try
-      select p
+      select (String.sub p 1 ((String.length p)-1)) (* Remove prepended space *)
```
brabalan: I don't understand where the prepended space comes from

g-raud: The space comes from the line 65 in `(" "^clause^" ")`; it is added to be able to detect the separator at the start or at the end of the pattern since `Util.trimWhitespace` has already been applied to the pattern.

brabalan: It makes sense, thanks.

g-raud:
Hello, this change is approved but still pending.

PR #190, opened 2018-04-05
Uitext: fix more newline printing and clear infos before failing

g-raud:


g-raud:
Hello, this is approved but still pending.

PR #201, opened 2018-05-24
Move ExternalSettings creation to makefile rather than Xcode project

### LINKED ISSUE #199: Bad binaries on github
File:
https://github.com/bcpierce00/unison/releases/download/v2.51.2/Unison-2.51.2.OS.X.zip

does not contain correct build. I found that Unison.app has no version (key CFBundleShortVersionString is not set in Info.plist file). I don't know if there are other issues.

I've build from source latest version and result Unison.app is correct. It's probably just the build uploaded that is bad.

alanshutko:
xcodebuild was not picking up settings if run as a target.

This fixes #199

### DIFF
--- a/tests/travis-ocaml.sh
+++ b/tests/travis-ocaml.sh
@@ -90,6 +90,7 @@ install_on_osx () {
   # sudo hdiutil attach XQuartz-2.7.6.dmg
   # sudo installer -verbose -pkg /Volumes/XQuartz-2.7.6/XQuartz.pkg -target /
   brew update >/dev/null
+  brew upgrade python
   case "$OCAML_VERSION,$OPAM_VERSION" in
     3.12,1.2.2) OPAM_SWITCH=3.12.1; brew install opam ;;
     4.00,1.2.2) OPAM_SWITCH=4.00.1; brew install opam ;;


brabalan:
Thank you, the patch looks good. But does this mean we no longer can build Unison directly from Xcode? (I'm fine with that, but we should document it if it's the case.)

alanshutko:
I didn't think you could ever build it directly from Xcode without running the make, because it needed to compile all the ocaml files.

brabalan:
Of course, I forgot about that.

@bcpierce00 I recommend that we merge this PR.

bcpierce00:
Do we know why the Travis build is failing?

alanshutko:
It looks like it has been failing for a couple months on Mac: https://travis-ci.org/bcpierce00/unison/pull_requests <https://travis-ci.org/bcpierce00/unison/pull_requests>

I am not even seeing the logs for 

sh -x ./tests/travis-ocaml.sh &>travis.log

Is there a way to look at the workspace and see what is in travis.log? It looks like it’s failing before the make.

brabalan:
It seems that there is a problem with Travis and OS X. I found this workaround https://github.com/commontk/AppLauncher/commit/5fd1ccc32cff5950297ba3de07b81b343cc5f387 but I don't know anything about Travis.

alanshutko:
Looks like this is it.  I’m working on debugging this on my branch. Please do not merge yet.

https://github.com/Homebrew/homebrew-core/issues/26358

PR #233, opened 2018-11-28
[build] Preliminary support for building with Dune.

ejgallego:
Dune is a domain-specific build system that has become the standard
in the OCaml ecosystem.

This PR adds preliminary support for building Unison with Dune; the PR
itself is quite simple and doesn't modify the make-based system.

It has been tested in Linux and does work correctly. Among the new
features this PR provides are:

- automatic developer tools setup [merlin, etc...]
- Unison can now become supported in the OCaml platform,
- builds tested from 4.02.3 to 4.07.1,
- automatic generation of OPAM package / install files,
- automatic generation of API documentation [if so wished],
- massive linting of the codebase suggested [see warnings],
- compositional builds, easy library vendoring,
- hygienic builds.

I do believe this PR is important to bring Unison closer to the modern
OCaml toolchain. Additionally, some important platform changes are
upcoming, for example lablgtk2 is deprecated, and this PR will allow
to compose the Dune build with lablgtk3, greatly facilitating the
porting.

This PR opens up the question of what the supported OCaml version
are. 4.02.3 -- 4.07.1 seems like a sensible choice as of today.
Depending on the answer, we could then address:

- `-safe-string`: unison should be ported to -safe-string, however
  this usually places a lower-bound on OCaml 4.02.3.

- warnings: the codebase could benefit from the massive linting
  suggested by the compiler, however to make the warning system happy
  precise compiler version bounds have to be established.

TODO:

- Update CI
- Documentation ?
- OSX support:
  it shouldn't be too hard using the foreign sandboxing described at
  https://dune.readthedocs.io/en/latest/foreign-code.html#foreign-build-sandboxing
  unfortunately, I am not a Mac user.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,7 @@ TAGS
 build
 logmsg
 xcuserdata
+
+_build
+.merlin
+*.install
--- a/dune-project
+++ b/dune-project
@@ -1,0 +1,2 @@
+(lang dune 1.3)
+(name unison)
--- a/src/dune
+++ b/src/dune
@@ -1,0 +1,27 @@
+(library
+ (name unison_lib)
+ (wrapped false)
+ (modules :standard \ linktext linkgtk linkgtk2 uigtk2 uimacbridge uimacbridgenew test)
+ (modules_without_implementation ui)
+ (flags :standard
+        -w -3-6-9-10-26-27-32-34-35-38-39-50-52
+        -warn-error -3-6-9-10-26-27-32-34-35-39-50-52
+        -unsafe-string -no-strict-sequence)
+ (c_names bytearray_stubs osxsupport pty)
+ (c_library_flags -lutil)
+ (libraries str unix lwt_lib bigarray))
+
+(include_subdirs unqualified)
+
+(executable
+ (name linktext)
+ (public_name unison)
+ (modules linktext)
+ (libraries unison_lib))
+
+(executable
+ (name linkgtk2)
+ (public_name unison-gtk2)
+ (flags :standard -w -3-6-9-27-32-52 -unsafe-string)
+ (modules linkgtk2 uigtk2)
+ (libraries threads unison_lib lablgtk2))
--- a/src/fsmonitor/dune
+++ b/src/fsmonitor/dune
@@ -1,0 +1,1 @@
+(include_subdirs no)
--- a/src/fsmonitor/linux/dune
+++ b/src/fsmonitor/linux/dune
@@ -1,0 +1,16 @@
+(copy_files# ../watchercommon.ml{,i})
+
+(library
+ (name fswatcher)
+ (wrapped false)
+ (modules :standard \ watcher)
+ (flags :standard -w -3-27-39 -unsafe-string)
+ (c_names inotify_stubs)
+ (libraries unix lwt_lib))
+
+(executable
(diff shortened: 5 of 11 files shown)


bcpierce00:
> This PR opens up the question of what the supported OCaml version
are. 4.02.3 -- 4.07.1 seems like a sensible choice as of today.
Depending on the answer, we could then address:

I don't see a strong reason to maintain compatibility with super-old versions of OCaml.  @brabalan should also weigh in on this question, though.

> -safe-string: unison should be ported to -safe-string, however
this usually places a lower-bound on OCaml 4.02.3.

This seems reasonable.

> warnings: the codebase could benefit from the massive linting
suggested by the compiler, however to make the warning system happy
precise compiler version bounds have to be established.

Since the code currently seems to work fine, I'm disinclined to spend a lot of anybody's effort cleaning up warnings.  Would rather concentrate energy on keeping the codebase working as OCaml and the rest of the world evolve.

ejgallego:
Thanks @bcpierce00 !

> I don't see a strong reason to maintain compatibility with super-old versions of OCaml. @brabalan should also weigh in on this question, though.

Cool, I guess I'll give a try to -safe-string. GTK3 is a problem tho; lablgtk2 is going to be removed from major distros soon, and I think lablgtk3 only supports OCaml >= 4.05.0 . This is not urgent but at some point will have to be addressed.

> Since the code currently seems to work fine, I'm disinclined to spend a lot of anybody's effort cleaning up warnings.

That seems reasonable. I would say that maybe the ones that signal a dangerous or compatibility situation are worth cleaning.

brabalan:
I'm also in favor of using a recent OCaml. Regarding safe strings, I looked at it and got scared when I had to change unicode manipulating functions (in unicode.ml).

bcpierce00:
Again on safe strings, I also don’t see why this is a critical change — it’s true that safe strings are better, but do we have any reason to suspect that there’s actually anything wrong about the way strings are handled in Unison now?

ejgallego:
> Again on safe strings, I also don’t see why this is a critical change — it’s true that safe strings are better, but do we have any reason to suspect that there’s actually anything wrong about the way strings are handled in Unison now?

The main issue is that I understand OCaml plans to drop support for `-unsafe-string` not too far in the future.

ejgallego:
There is nothing wrong about how Unison handles strings, indeed this is more of a case of OCaml upstream having chosen a rough migration path.

bcpierce00:
Ah.  Why???

ejgallego:
Mainly the desire to have "pure" [or immutable if you want to call it that way] string type. This is convenient in some backends [for example in Javascript scripts are immutable].

There has been a lot of discussion over this, see for example https://github.com/ocaml/ocaml/pull/1274

Thus IMO it is worth porting things to work with `-safe-string` on. In my experience it is not a lot of work, [you can just `s/String/Bytes/`.

bcpierce00:
It would certainly be worth trying this to see if things still work.

brabalan:
> Thus IMO it is worth porting things to work with `-safe-string` on. In my experience it is not a lot of work, you can just `s/String/Bytes/`.

I remember trying this, but for some reason it was not trivial in unicode.ml. I don't have the time to work on this now but I can review a patch.

PR #236, opened 2018-12-21
enable macOS Mojave dark mode

mroi:
I updated the XIB file for the macOS client to look better in macOS Mojave’s new dark mode. These are just basic fixes for obvious flaws, I did not apply and fine tuning to (for example) the icon artwork.

Unfortunately, building the macOS client on Mojave requires a minimum deployment target of macOS 10.6. This was previously configured to 10.5, which I had to raise. Even without dark mode, this is needed to build on Mojave.

### DIFF
--- a/src/Makefile.OCaml
+++ b/src/Makefile.OCaml
@@ -173,7 +173,7 @@ else
   buildexecutable:: $(NAME)$(EXEC_EXT)
 endif
 
-MINOSXVERSION=10.5
+MINOSXVERSION=10.6
 # XCODEFLAGS=-sdk macosx$(MINOSXVERSION)
 ifeq ($(OSARCH),osx)
   CAMLFLAGS+=-ccopt -mmacosx-version-min=$(MINOSXVERSION)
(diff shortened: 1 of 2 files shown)


brabalan:
As I'm no longer on macOS, I won't be able to test this. I think raising the minimal target version to 10.6 is fine, as it's 6 years old if I'm not mistaken.

bcpierce00:
Agreed!

bcpierce00:
Can someone remind me how to clone a copy of this PR so I can try compiling before I merge it?

alanshutko:
> git clone https://github.com/mroi/unison.git

I swear there used to be an easier way to do this in the UI.

bcpierce00:
Thanks much.  Seems to work fine here.

mroi:
I don’t want to abuse this PR as a forum, but allow me one question:

Would there be interest in more visual refinements? I could see how to improve the hardcoded colors for dark mode and look into the options for hires artwork. But before I start, I wanted to check if that would be OK or if there is a Mac UI person I should collaborate with.

I cannot test on 10.6 though (which is now 9 years old), but can make an effort to keep the minimum target at least from an API availability perspective.

mroi:
Thanks, I will take this discussion to unison-hackers then. My development for the Mac UI will happen in [my uimac-update](https://github.com/mroi/unison) branch.

PR #252, opened 2019-03-05
cleanup macOS build system

mroi:
* enable builds on Xcode versions from 3.2 (macOS 10.6) up to current Xcode 10 (macOS 10.14)
* rename `uimac14` to `uimac`, because it is the only macOS frontend


bcpierce00:
Works for me.  @brabalan, what about you?

bcpierce00:
Actually, maybe there's no need to double-check this one, Alan: if it's OK on my machine, it will very likely also be OK on yours!

I'll just merge it.

bcpierce00:
Thank you, Michael!

brabalan:
I cannot test it, but I read through the commit and it looks good.

mroi:
Thanks for merging. My progress with the cleanup of the macOS client is currently quite slow, but I keep pushing. I still have the things on my todo list as mentioned earlier on the mailing list.

PR #255, opened 2019-03-21
Compatibility with OCaml 4.08

jaapb:
In order to make unison compile with 4.08, I needed to eta-expand the link function (there's a new optional parameter in 4.08's Unix.link function) and replace functions from the removed Sort module with their equivalents from List.


brabalan:
It looks good to me.

G-Rath:
Any chance of getting this released?

I'm trying to update my docker-sync, which requires ocaml 4.08, but still recommends "the latest unison release" which is 2.51.2 & fails when I attempt to built it due to "The implementation system.ml does not match the interface system.cmi".

I'm not an ocaml dev, so this might not be the fix - let me know if you'd like me to make a new issue or provide more info :)

PR #315, opened 2020-02-10
Fix warnings raised by OCaml 4.08.1

glondu:


glondu:
I've rebased my branch onto current master. Pull request #320 introduced a new warning (illegal escape sequence), I've fixed it too.

Note: this pull request precedes another one that fixes compilation without -unsafe-string (available under my [safe-string](https://github.com/glondu/unison/tree/safe-string) branch), which is apparently [needed](https://github.com/bcpierce00/unison/issues/316) with OCaml 4.10.

bjodah:
Just wanted to report that I tried this branch and warnings went away for 4.08.1.

bcpierce00:
Hmm -- I merged this PR without asking myself about the consequences of making 4.08.1 the minimum required OCaml version to compile Unison. Aren't there still a lot of people out there with older versions of OCaml?

h8ghbzc9bbr8c7:
Yes, even the (not very old) Ubuntu LTS 18.04 includes only OCaml 4.05. Maybe it is too early to raise the minimum required version so high.

glondu:
IMHO, those who care to have the latest Unison with an old OCaml in their distribution can use opam...

PR #320, opened 2020-02-20
Update clroot.ml

theegala:
Currently unison not supporting ipv6 in url "ssh://tjim@[fd00:4888:1::10]/hello/world".

Added regex to extract ip which is passed in "[]".

with this change users can use urls with ipv6 address as follows.
"ssh://tjim@[fd00:4888:1::10]:22/hello/world".
"ssh://tjim@[ipv6]:port/hello/world".


bcpierce00:
Thanks!
   - B

PR #326, opened 2020-03-23
Ignore unison.log file by default?

kirkins:
Sorry for incomplete PR, this is more of an idea.

My friend created an infinite loop by accidentally backing up `unison.log`, would it be worth ignoring files named `unison.log` by default or do you think that's too much of an edge case?


bcpierce00:
This seems like a good idea, but I think it would work better to do it at the OCaml level, so that (a) it will work across user interfaces, and (b) we can use the actual path to the log file (which the user might have changed) by looking at the appropriate setting.

    - B

kirkins:
That sounds like a better approach. I'm not too familiar with OCaml so I'm going to close this PR. It sounds like you have an idea of how it can be implemented.

PR #351, opened 2020-06-23
Compile without -unsafe-string

glondu:
In recent versions of the OCaml compiler, `-unsafe-string` is no longer available. It is time to fix Unison to not use it.


gliptak:
@bcpierce00 Travis PR build is green https://travis-ci.org/github/bcpierce00/unison/builds/701165634

Issue #353, opened 2020-07-03
OCaml 4.10 support


chenrui333:
During the ocaml 4.10 regression build, it looks like the unison does not build with OCaml 4.10, I have attached the build failure log below.

<details>
<summary>build failure log</summary>

```
 Last 150 lines from /Users/brew/actions-runner/_work/homebrew-core/homebrew-core/bottles/logs/unison/01.make:
   -impl <file>  Compile <file> as a .ml file
   -inline <n>|<round>=<n>[,...]  Aggressiveness of inlining (default 1.25, higher numbers mean more aggressive)
   -inline-toplevel <n>|<round>=<n>[,...]  Aggressiveness of inlining at toplevel (higher numbers mean more aggressive)
   -inline-alloc-cost <n>|<round>=<n>[,...]  The cost of not removing an allocation during inlining (default 7, higher numbers more costly)
   -inline-branch-cost <n>|<round>=<n>[,...]  The cost of not removing a conditional during inlining (default 5, higher numbers more costly)
   -inline-call-cost <n>|<round>=<n>[,...]  The cost of not removing a call during inlining (default 5, higher numbers more costly)
   -inline-prim-cost <n>|<round>=<n>[,...]  The cost of not removing a primitive during inlining (default 3, higher numbers more costly)
   -inline-indirect-cost <n>|<round>=<n>[,...]  The cost of not removing an indirect call during inlining (default 4, higher numbers more costly)
   -inline-lifting-benefit <n>|<round>=<n>[,...]  The benefit of lifting definitions to toplevel during inlining (default 1300, higher numbers more beneficial)
   -inlining-report  Emit `.<round>.inlining' file(s) (one per round) showing the inliner's decisions
   -insn-sched  Run the instruction scheduling pass (default)
   -intf <file>  Compile <file> as a .mli file
   -intf-suffix <string>  Suffix for interface files (default: .mli)
   -keep-docs  Keep documentation strings in .cmi files
   -no-keep-docs  Do not keep documentation strings in .cmi files (default)
   -keep-locs  Keep locations in .cmi files (default)
   -no-keep-locs  Do not keep locations in .cmi files
   -labels  Use commuting label mode
   -linkall  Link all modules, even unused ones
   -inline-max-depth <n>|<round>=<n>[,...]  Maximum depth of search for inlining opportunities inside inlined functions (default 1)
   -alias-deps  Do record dependencies for module aliases
   -no-alias-deps  Do not record dependencies for module aliases
[...]

PR #354, opened 2020-07-03
Bump OCaml to v4.10

chenrui333:
relates to #353

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,12 +14,11 @@ cache:
   - /usr/local
 
 env:
-  # - OCAML_VERSION=4.01
-  - OCAML_VERSION=4.08
+  - OCAML_VERSION=4.10
 
 os:
   - linux
-#  - osx
+  - osx
 
 # BCP and brabalan, 1/19: We took away the OSX build because it is
 # horribly slow and 98% redundant...


gdt:
This pull request is not linked to an issue (maybe that's ok), but it does not explain what it does and why, and why it would safe to merge, or good to merge.

It seems that unison needs to support the current ocaml release and several previous ones.  By "support", I mean one can build unison with that ocaml version and it will work.   Right now it seems 4.08 and 4.09 are ok.  I don't see an issue "building unison with ocaml 4.10 fails", and I can't parse "bump" in terms of whether building with any older version that works now will still work.  Also, I can't  be sure that there are no other adverse consequences.

bcpierce00:
Agreed that we should keep Unison limping along on old OCaml versions until it becomes inconvenient to do so, and “bump” only then.

gdt:
I will let @chenrui333 explain, but I am inclined to close this after the 30-day feedback timeout (that I have heavy-handedly declared).

gdt:
If anything, we are working on making unison work with even older ocaml.

chenrui333:
Thanks @gdt, @bcpierce00, it might be also a good idea to do the cross compilation (if not too much hassle).

But I am OK with what is right now.

PR #359, opened 2020-08-12
Add ACL synchronization support (NFSv4 ACL, limited POSIX-draft ACL and NTFS ACL)

tleedjarv:
### Description

This adds support for synchronizing ACLs on files and directories. Primary intention is to support NFSv4 ACLs on Unix-like platforms but there is limited POSIX draft ACL support as well.
Windows ACLs are are now also supported.
Since synchronization is done via text serialization of ACL then theoretically support for any platform/ACL type could be added.

### Platform support

Currently supported platforms are Solarish - all illumos-based OS (OpenIndiana, SmartOS, OmniOS, ...), FreeBSD, NetBSD, Darwin (macOS) and Windows.
Linux? ...unfortunately, no. Maybe once https://github.com/openzfs/zfs/pull/9709 becomes reality.

### Testing

I have done basic testing in following scenarios:

- OmniOS <-> OmniOS  (NFSv4 ACL on ZFS)
- OmniOS <-> OpenIndiana 151a9  (NFSv4 ACL on ZFS)
- OmniOS <-> OpenIndiana 151a9  (NFSv4 ACL on ZFS <-> POSIX draft ACL)
- OmniOS <-> FreeBSD  (NFSv4 ACL on ZFS)
- OmniOS <-> FreeBSD  (NFSv4 ACL on ZFS <-> POSIX draft ACL)
- OpenIndiana 151a9 <-> OpenIndiana 151a9  (NFSv4 ACL on ZFS)
- OpenIndiana 151a9 <-> OpenIndiana 151a9  (NFSv4 ACL on ZFS <-> POSIX draft ACL)
- OpenIndiana 151a9 <-> FreeBSD  (NFSv4 ACL on ZFS)
- OpenIndiana 151a9 <-> FreeBSD  (NFSv4 ACL on ZFS <-> POSIX draft ACL)
- FreeBSD <-> FreeBSD  (NFSv4 ACL on ZFS)
- FreeBSD <-> FreeBSD  (POSIX draft ACL)
- Darwin (macOS) <-> Darwin (macOS) (extended ACL)
- Windows 7 32-bits <-> Windows 7 32-bits
- Windows 10 64-bits <-> Windows 10 64-bits
- Windows 10 64-bits <-> Windows 7 32-bits

(_NFSv4 <-> POSIX draft synchronization is not intended to work on non-illumos platforms_)

### DIFF
--- a/doc/unison-manual.tex
+++ b/doc/unison-manual.tex
@@ -2174,10 +2174,72 @@ propagated.  The values of the other bits are set to default values
 Unix system).
 \item For security reasons, the Unix \verb|setuid| and \verb|setgid|
 bits are not propagated.
-\item The Unix owner and group ids are not propagated.  (What would
-this mean, in general?)  All files are created with the owner and
-group of the server process.
+\item The Unix owner and group ids can be propagated (see \verb|owner|
+and \verb|group| preferences) by mapping names or by numberic ids (see
+\verb|numericids| preference).
 \end{itemize}
+
+
+\SUBSECTION{Access Control Lists - ACLs}{acls}
+
+Unison allows synchronizing access control lists (ACLs) on platforms
+and filesystems that support them. In general, synchronization makes
+sense only in case both replicas support the same type of ACLs and
+recognize same users and groups. In some cases you may be able to
+go beyond that and synchronize ACLs to a replica that couldn't fully
+use them---this may be be useful for the purpose of preserving ACLs.
+
+If one of the replicas does not support any type of ACLs then
+Unison will not attempt ACL synchronization. If the other replica
+does support ACLs then those will remain intact.
+
+If both replicas support ACLs of any supported type then you can
+request Unison to try ACL synchronization (\verb|acl| preference).
+Success of synchronization depends on permissions of the owner and
+group of Unison process (Unison must have permissions to set ACL)
+and the compatibility of ACL types on both replicas.
+
+An ACL is propagated as a single unit, with all ACEs. There is no
+merging of ACEs from the replicas.
+
+{\em Caveat}: ACE inheritance may in certain scenarios cause synchronization
+inconsistencies. In Windows, only explicit ACEs are synchronized; inherited
+ACEs are not actively synchronized, but Windows will propagate ACEs from parent
+directories (unless inheritance is explicitly prevented on a file or a
+directory---this prevention is also synchronized). Due to inheritance, the
+ultimately effective ACL may be different, or provide different access, even
+after synchronization.
+
+Unison currently supports the following platforms and ACL types:
+\begin{itemize}
+  \item Windows (Windows XP SP2 and later)
+  \begin{itemize}
+    \item NTFS ACL (discrete ACL (DACL) only)
+  \end{itemize}
+\item Solaris, OpenSolaris and illumos-based OS (OpenIndiana, SmartOS,
+  OmniOS, etc.)
+  \begin{itemize}
+  \item NFSv4 ACL (ZFS ACL)
+  \item POSIX-draft ACL
+  \item Some NFSv4 ACL (ZFS ACL) cross-synchronization with
+  POSIX-draft ACL
+  \item Full cross-synchronization with other platforms that support
+  NFSv4 ACLs; limited cross-synchronization with POSIX-draft ACLs
+  \end{itemize}
+\item FreeBSD, NetBSD
+  \begin{itemize}
+  \item NFSv4 ACL (ZFS ACL)
+  \item Limited POSIX-draft ACL (access ACL only; not default ACL)
+  \item Full cross-synchronization with other platforms that support
+  NFSv4 ACLs
+  \end{itemize}
+\item Darwin (macOS)
+  \begin{itemize}
+  \item Extended ACL
+  \end{itemize}
+\end{itemize}
+Not all filesystems on the listed platforms support all ACL types
+(or any ACLs at all).
 
 
 \SUBSECTION{Extended Attributes - xattrs}{xattrs}
--- a/src/Makefile.OCaml
+++ b/src/Makefile.OCaml
@@ -161,6 +161,10 @@ else
         CLIBS+=-cclib -lutil
       endif
     endif
+    ifeq ($(OSARCH),solaris)
+      # ACL functions
+      CLIBS+=-cclib -lsec
+    endif
     buildexecutable::
 	@echo Building for Unix
   endif
@@ -247,7 +251,7 @@ OCAMLOBJS+=main.cmo
 OCAMLLIBS+=unix.cma str.cma
 INCLFLAGS+=-I +unix -I +str
 
-COBJS+=osxsupport$(OBJ_EXT) pty$(OBJ_EXT) bytearray_stubs$(OBJ_EXT) hash_compat$(OBJ_EXT) props_xattr$(OBJ_EXT)
+COBJS+=osxsupport$(OBJ_EXT) pty$(OBJ_EXT) bytearray_stubs$(OBJ_EXT) hash_compat$(OBJ_EXT) props_xattr$(OBJ_EXT) props_acl$(OBJ_EXT)
 
 ########################################################################
 ### User Interface setup
--- a/src/TODO.txt
+++ b/src/TODO.txt
@@ -285,14 +285,6 @@ should strip symbols from binary files in 'make exportnative'
     offer to delete them *for* the user, rather than forcing the user to
     delete them manually.
 
-*** A switch to include NTFS ACE/ACL file permissions to be copied when
-    copying from one NTFS location to another NTFS location. As I
-    mentioned this is less generic, but of fundamental usefullness in
-    Windows usage, as NTFS permissions are absolutely essential in many
-    backup/replication situations in Windows systems. Robocopy has the
-    /SEC switch, but Unison is a far better tool, and I was hoping in that
-    light that Unison could implement the rights/permissions stuff also.
-
 *** There is no command-line argument to tell Unison where the .unison
     directory is; Unison finds it in the environment or not at all.  I was
     able to workaround this with a symbolic link to put .unison where it was
--- a/src/dune
+++ b/src/dune
@@ -9,7 +9,7 @@
         -no-strict-sequence)
  (foreign_stubs
   (language c)
-  (names bytearray_stubs osxsupport pty hash_compat props_xattr))
+  (names bytearray_stubs osxsupport pty hash_compat props_xattr props_acl))
  (c_library_flags -lutil)
  (libraries str unix lwt_lib bigarray))
 
--- a/src/fs.ml
+++ b/src/fs.ml
@@ -79,6 +79,11 @@ let xattrUpdatesCTime = System.xattrUpdatesCTime
 
 (****)
 
+let acl_get_text f = System.acl_get_text (path f)
+let acl_set_text f acl = System.acl_set_text (path f) acl
+
+(****)
+
 let fingerprint f = System.fingerprint (path f)
 
 let hasInodeNumbers () = System.hasInodeNumbers ()
--- a/src/fsmonitor/windows/Makefile
+++ b/src/fsmonitor/windows/Makefile
@@ -13,7 +13,7 @@ FSMOCAMLOBJS = \
 FSMCOBJS = \
    bytearray_stubs$(OBJ_EXT) \
    system/system_win_stubs$(OBJ_EXT) lwt/lwt_unix_stubs$(OBJ_EXT) \
-   props_xattr$(OBJ_EXT)
+   props_xattr$(OBJ_EXT) props_acl$(OBJ_EXT)
 FSMOCAMLLIBS=unix.cma
 
 ifeq ($(NATIVE), true)
--- a/src/props.ml
+++ b/src/props.ml
@@ -1119,7 +1119,31 @@ module ACL : sig
   include S
   val get : Fspath.t -> Unix.LargeFile.stats -> t
   val check : Fspath.t -> Path.local -> Unix.LargeFile.stats -> t -> unit
+  module Data : sig
+    include Propsdata.S
+    val keep : t -> unit
+  end
 end = struct
+
+module Data = struct
+  include Propsdata.ACL
+
+  let keep = function
+    | None | Some "" -> ()
+    | Some s -> keep s (* [keep] of Propsdata.ACL *)
+end
+
+(* The result value of this function must be deterministic for its input
+   (over both roots, and over time, as long as it is the same archive). *)
+let deflate acl =
+  if acl = "" then acl
+  else begin
+    let key = Digest.string acl in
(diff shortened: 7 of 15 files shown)


tleedjarv:
Darwin support is now also compiling and works. I have done basic testing on macOS 10.15

bcpierce00:
> There is a remaining problem with the dontverifyacl preference. The issue stems from that the propagated ACL is not always set and/or read back 100% the same way as in the originating replica. That may be completely acceptable to the user as there may have been no semantic loss or the loss is acceptable. So, I added the dontverifyacl preference to skip the Props.check on the ACL after synchronization. However, the archive still records the ACL from the originating replica. Now the archive considers the file/dir to be up to date, yet it actually has a different ACL on disk. The next time Unison is run, it will detect it as an update, of course. Which it is not. I don't know what is the best way to handle this situation. @bcpierce00, could you suggest a way to update the archive with the ACL that is read back from the file/dir, instead of the one received from the other replica?

The invariant that the archives are the same on both sides is deeply built into Unison's DNA: loosening it is probably not a good idea.  However, there are already situations where Unison accepts a "small difference" between what it sees in the filesystem and what it sees in the archive -- e.g., modification times are allowed to be slightly different (to accommodate, for example, the fact that on some windows filesystems the granularity of mod times is 2 seconds).  Could we go this route?

tleedjarv:
Thank you for the clarification.
I had overlooked the fact that archives **must** be the same on both sides.

In that case I will consider what differences can be accepted similar to the modification time scenario. As for the remaining cases, I guess that means the user will have to turn off ACL synchronization completely.

tleedjarv:
I have now resolved the remaining problem by tapping into the "numericids" preference and removing the "dontverifyacl" preference completely.
In conclusion, the archive equality invariant is kept, with the following small differences for ACLs allowed:
- some cosmetic differences are covered already in the C code
- user/group names are ignored and only numeric ids compared if the user sets "numericids" preference; without this preference, user/groups names and numeric ids both have to match

If there are other cosmetic or non-significant differences coming up later, those can be covered in a similar manner.

tleedjarv:
I have done some primitive non-functional testing with the code in this PR.

**Test setup:**
- ca. 42 000 files per replica
- rotating disks (can sustain probably around 160-180 IOPS max)
- both replicas on the same machine and same disks - no network overhead, but same CPU and disks process both replicas: so, in total ca. 84 000 files to process in a single test run
- each file has on average 6 ACEs in ACL (ca. 260 bytes in text representation)
- empty cache means that none of the files' metadata was cached
- cached runs means that 100% of the files' metadata was cached: the entire run required 0 disk accesses
- averages are taken from 10 runs, but all the results where extremely close (deviations no more than 1%)

**Results:**

<table>
<thead>
<tr>
<th></th><th>Unison 2.51.2 baseline</th><th>Unison PR359 ACL off</th><th>delta</th><th>Unison PR359 ACL enabled</th><th>delta</th>
</tr>
</thead>
<tbody>
<tr>
<td>Empty cache run avg</td><td><pre>real 0m31.738s
user 0m0.511s
sys  0m2.254s</pre></td><td><pre>real 0m32.845s
user 0m0.620s
sys  0m2.305s</pre></td><td><pre>1.107s  (3.5%)
0.109s (21.3%)
0.051s  (2.3%)</pre></td><td><pre>real 0m45.583s
user 0m2.491s
sys  0m7.411s</pre></td><td><pre>12.738s  (39%)
 1.871s (300%)
 5.106s (221%)</pre></td>
</tr>
<tr>
<td>Cached run avg</td><td><pre>real 0m1.387s
user 0m0.404s
sys  0m0.972s
</pre></td><td><pre>real 0m1.498s
user 0m0.492s
sys  0m0.993s
</pre></td><td><pre>0.111s  (8.0%)
0.088s (21.8%)
0.021s  (2.2%)
</pre></td><td><pre>real 0m11.621s
user 0m2.307s
sys  0m5.840s
</pre></td><td><pre>10.123s (675%)
 1.815s (369%)
 4.847s (488%)
</pre></td>
</tr>
<tr>
<td>Archive size</td><td>3.25 MB</td><td>3.52 MB</td><td>+0.27 MB (8.3%)</td><td>14.65 MB</td><td>+11.13 MB (315%)</td>
</tr>
</tbody>
</table>

**Discussion:**

- I can't explain the difference between 2.51.2 baseline and this PR baseline, other than that Unison has been compiled on different machines with different OCaml versions and this PR also includes all other code changes since 2.51.2.
- The delta numbers look terrible, but absolute numbers with this replica size are so small that it makes no real difference in practice. Those who need to sync ACLs are well prepared to accept this time and size increase.
- It seems like the time increase with ACL enabled comes equally from time spent in syscalls, OCaml-C interface and string manipulations. Any optimizations, if attempted, would need to focus on the latter two.
- Even though the text representation of ACL requries about 260 bytes per file/directory, with this replica size, there is absolutely no reason to optimize this (e.g. by storing a fingerprint in the archive).
- I do not have properly measured sync times because the data size in replica is huge and it is not easy to run repeated attempts. It seems that enabling ACL sync may increase the sync times somewhat, but it is really inconclusive as with so much data reading and writing on same rotating disks the results are really unpredictable and are not comparable run per run. I may eventually attempt smaller test cases with only a subset of files updated/synced.
- My main Unison replica is around 700 000 files with archive size 55 MB. Extrapolating these results to this replica would result in archive size of about 250 MB. I don't think this is any issue at all on the disk. It would be interesting to measure run-time memory usage, though.

gdt:
What happens when synchronizing between a replica that supports ACLs and one that does not?  In the case when it has ACLs, and when it doesn't?

(Agreed that this needs to be deferred until we have format bumps.)

tleedjarv:
> What happens when synchronizing between a replica that supports ACLs and one that does not?

In this case it is almost as if the `acl` preference was not given. Files in replica that doesn't support ACL will not be affected. New files in replica that supports ACL will be created with whatever inherited/default ACL is assigned by the fs. Updating existing files in this replica from the other replica will not strip files of their ACL.

> In the case when it has ACLs, and when it doesn't?

If ACL is supported in both replicas then syncing is done fully, i.e. "no ACL" is a valid state and will be synced. If a file is synced from a replica that does not have an ACL to the other replica then syncing will strip the ACL in other replica and vice versa - when syncing from replica with ACL, the ACL will be set on file in the other replica.

tleedjarv:
Rebased to current master.

**Some recent memory usage figures**
**Test setup**

- Sync between directories on the same machine.
- 200 000 files in one replica; other replica empty - full initial synchronization.
- ACL text representation size 230 bytes per file (6 ACEs).

**Results**
<table>
<thead>
<tr>
<th></th><th>2.51.3.70 baseline</th><th>PR359 ACL off</th><th>PR359 ACL enabled</th>
</tr>
</thead>
<tbody>
<tr>
<td>Archive size</td><td>14.4 MB</td><td>14.6 MB</td><td>62.4 MB</td>
</tr>
<tr>
<td>RSS (max)</td><td>272 MB</td><td>273 MB</td><td>408 MB</td>
</tr>
</tbody>
</table>

- Archive size increase: 251 bytes per file.
- Memory usage increase: 708 bytes per file (~3 x ACL length per file).
- Baseline memory usage per file: 1426 bytes.

tleedjarv:
> It seems like the time increase with ACL enabled comes equally from time spent in syscalls, OCaml-C interface and string manipulations. Any optimizations, if attempted, would need to focus on the latter two.

I've done some additional testing now and it appears that this assumption was plain wrong. > 99.5% of the increased run time is spent in system library functions + syscalls. Unison code itself (both OCaml and C together) is responsible for less than 0.5% of run time increase.

I also experimented with declaring external functions [@@noalloc] but did not see any tangible differences compared to code without [@@noalloc] (testing with ~400 000 function calls, which is probably too little to show any benefits).

tleedjarv:
I've added ctime checking in order to make metadata (what is not provided by `stat`) updates checking faster; the idea from https://github.com/bcpierce00/unison/pull/368#issuecomment-691532473
ctime checking code is universal, not specific to ACL or xattr, but it does not provide any benefits for metadata that is available from `stat`.

In my testing, this has worked out very well. Time taken scanning updates with ACL syncing is now on par with time taken without ACL syncing. Another positive side-effect is the reduction of memory usage. as the copies of ACL are no longer kept in memory twice (once in archive and once the current one).

The caveat? Archive size is increased by 14 bytes per file.

tleedjarv:
Repeating tests from [above](https://github.com/bcpierce00/unison/pull/359#issuecomment-691281847), now with ctime checking, the results are much improved.

<table>
<thead>
<tr>
<th></th><th>Unison 2.51.2 baseline</th><th>Unison PR359 ACL off</th><th>delta</th><th>Unison PR359 ACL enabled</th><th>delta</th>
</tr>
</thead>
<tbody>
<tr>
<td>Cached run avg</td><td><pre>real 0m1.387s
user 0m0.404s
sys  0m0.972s
</pre></td><td><pre>real 0m1.389s
user 0m0.482s
sys  0m0.894s
</pre></td><td><pre> 0.002s  (0.1%)
 0.078s (19.3%)
-0.078s   (-8%)
</pre></td><td><pre>real 0m1.680s
user 0m0.600s
sys  0m1.053s
</pre></td><td><pre>0.291s (21%)
0.118s (24%)
0.159s (18%)
</pre></td>
</tr>
<tr>
<td>Partial props update avg</td><td></td><td><pre>real 0m3.024s
user 0m1.715s
sys  0m1.293s</pre>
Actual updates: 0.100 s</td><td></td><td><pre>real 0m3.606s
user 0m1.877s
sys  0m1.581s</pre>
Actual updates: 0.254 s</td><td><pre>0.582s (19%)
0.162s  (9%)
0.288s (22%)
</pre>
Actual updates: +0.154 s (154%)</td>
</tr>
<tr>
<td>Archive size</td><td>3.25 MB</td><td>3.86 MB</td><td>+0.61 MB (19%)</td><td>15.38 MB</td><td>+11.52 MB (300%)</td>
</tr>
<tr>
<td>RSS</td><td></td><td>47 MB</td><td></td><td>95 MB</td><td>+48 MB (102%)</td>
</tr>
</tbody>
</table>

tleedjarv:
I've added a commit which implements ACL deduplication. As most ACL values will be the same for many files (for example, all files in a directory typically have the same ACL) and a single ACL text representation can be rather large (average 200-400 bytes), deduplication results in significant reduction in memory need and archive size. This is achieved in exchange to slightly increased processing overhead, but that should remain non-significant.

All unique ACLs are put in a lookup table with an MD5 hash as the key. The lookup table is stored in the archive file in the archive properties list.

Updated results for tests in https://github.com/bcpierce00/unison/pull/359#issuecomment-722539953 are now:

<table>
<thead>
<tr>
<th></th><th>2.51.3.70 baseline</th><th>PR359 ACL off</th><th>PR359 ACL enabled</th>
</tr>
</thead>
<tbody>
<tr>
<td>Archive size</td><td>14.4 MB</td><td>17.1 MB (increase caused by ctime commit)</td><td>20.3 MB (-42.1 MB with this commit, with ctime commit causing ~2.5 MB increase, so -44.6 MB only with this commit)</td>
</tr>
</tbody>
</table>


**Test setup 2 results**
<table>
<thead>
<tr>
<th></th><th>PR359 ACL off<br/>initial sync</th><th>PR359 ACL off<br/>next sync</th><th></th><th>PR359 ACL enabled<br/>initial sync</th><th>PR359 ACL enabled<br/>next sync</th>
</tr>
</thead>
<tbody>
<tr>
<td>RSS (max)</td><td>301 MB (the increase here is caused by the ctime commit)</td><td>198 MB</td><td></td><td>326 MB (-120 MB with this commit)</td><td>220 MB (-220 MB with this commit)</td>
</tr>
</tbody>
</table>

tleedjarv:
Repeating tests from [above](https://github.com/bcpierce00/unison/pull/359#issuecomment-691281847), now with ctime checking and ACL deduplication, the results are as follows.

<table>
<thead>
<tr>
<th></th><th>Unison 2.51.2 baseline</th><th>Unison PR359 ACL off</th><th>delta</th><th>Unison PR359 ACL enabled</th><th>delta</th>
</tr>
</thead>
<tbody>
<tr>
<td>Empty cache run avg</td><td><pre>real 0m31.738s
user 0m0.511s
sys  0m2.254s</pre></td><td><pre>real 0m25.519s
user 0m0.583s
sys  0m1.797s</pre></td><td><pre>-6.219s (-20%)
 0.072s  (14%)
-0.457s (-20%)</pre></td><td><pre>real 0m31.120s
user 0m0.624s
sys  0m1.946s</pre></td><td><pre>5.601s (22%)
0.041s (7%)
0.149s (8%)</pre></td>
</tr>
<tr>
<td>Cached run avg</td><td><pre>real 0m1.387s
user 0m0.404s
sys  0m0.972s
</pre></td><td><pre>real 0m1.389s
user 0m0.482s
sys  0m0.894s
</pre></td><td><pre> 0.002s  (0.1%)
 0.078s (19.3%)
-0.078s   (-8%)
</pre></td><td><pre>real 0m1.536s
user 0m0.515s
sys  0m1.006s
</pre></td><td><pre>0.147s (11%)
0.033s (7%)
0.112s (13%)
</pre></td>
</tr>
<tr>
<td>Partial props update avg</td><td></td><td><pre>real 0m3.024s
user 0m1.715s
sys  0m1.293s</pre>
Actual updates: 0.100 s</td><td></td><td><pre>real 0m3.515s
user 0m1.812s
sys  0m1.567s</pre>
Actual updates: 0.260 s</td><td><pre>0.491s (16%)
0.097s  (6%)
0.274s (21%)
</pre>
Actual updates: +0.16 s (160%)</td>
</tr>
<tr>
<td>Archive size</td><td>3.25 MB</td><td>3.86 MB</td><td>+0.61 MB (19%)</td><td>4.40 MB</td><td>+0.54 MB (14%)</td>
</tr>
<tr>
<td>RSS</td><td></td><td>47 MB</td><td></td><td>52 MB</td><td>+5 MB (10%)</td>
</tr>
</tbody>
</table>

tleedjarv:
Added code for upcoming ACL support in NetBSD.

tleedjarv:
Rebased, did some minor refactoring, and added Windows ACL synchronization support.

On NTFS, discrete ACL (DACL) can be synchronized. Only explicit ACE are synchronized, the inherited ACE (whether inherited from within synchronization root or outside it) are not synchronized, but are propagated by Windows by the regular propagation rules.

There is currently a bug with remote sync, but it is not related to Windows code.

tleedjarv:
Rebased and improved the code. This feature no longer breaks backwards compatibility.

It is ready for review and testing.

The PR is currently based on the xattrs PR.

One note about backwards compatibility. Once the archive file has been upgraded, it can no longer be opened by an older version. I did not try to do anything about this because I don't think this is a common usage pattern. If there really is a need to downgrade then doing a rescan is completely acceptable (or if doing an immediate downgrade, having a backup of archive files is to be expected).

tleedjarv:
Ready to be merged.

PR #360, opened 2020-08-14
Add ANSI color to text UI

### LINKED ISSUE #345: Add color output for terminal
For the terminal version of Unison, I think it would be useful to have a color option.  With this option, in the final summary...

1 item transferred, 0 skipped, 1 failed

...red text would be used for the "failed" report if there are one or more failed transfers.  I think most terminal programs support ANSI text colors these days.  I'm sure there are other situations where color would be helpful to call attention to exceptional situations.

tleedjarv:
Closes #345 
Respects `NO_COLOR` environment variable.
In addition to the synchronization status, also syncrhonization actions and output of unified diff are colored. If using a PAGER then users may have to change pager options, e.g. `less -R`.

### DIFF
--- a/src/uitext.ml
+++ b/src/uitext.ml
@@ -86,6 +86,68 @@ let setupTerminal() =
     with Unix.Unix_error _ ->
       restoreTerminal ()
 
+let colorMode =
+  Prefs.createBoolWithDefault "color" ~local:true
+    "!use color output for text UI (true/false/default)"
+    ("When set to {\\tt true}, this flag enables color output in "
+     ^ "text mode user interface. When set to {\\tt false}, all "
+     ^ "color output is disabled. Default is to enable color if "
+     ^ "the {\\tt NO\\_COLOR} environment variable is not set.")
+
+let colorEnabled = ref false
+
+let setColorPreference () =
+  let envOk = try let _ = System.getenv "NO_COLOR" in false
+    with Not_found -> true
+  and termOk = try System.getenv "TERM" <> "dumb" with Not_found -> true
+  and ttyOk = (Unix.isatty Unix.stdin) && (Unix.isatty Unix.stderr) in
+  let colorOk = envOk && termOk && ttyOk && not (Prefs.read dumbtty) in
+  colorEnabled :=
+    match Prefs.read colorMode with
+    | `True    -> true
+    | `False   -> false
+    | `Default -> colorOk
+
+let color t =
+  if not !colorEnabled then "" else
+  match t with
+    `Reset       -> "\o033[0m"
+  | `Focus       -> "\o033[1m"
+  | `Success     -> "\o033[1;32m"
+  | `Information -> "\o033[1;34m"
+  | `Warning     -> "\o033[1;33m"
+  | `Failure     -> "\o033[1;31m"
+  | `AError      -> "\o033[31m"
+  | `ASkip       -> "\o033[1;35m"
+  | `ALtoRf      -> "\o033[1;32m"
+  | `ALtoRt      -> "\o033[1;33m"
+  | `ARtoLf      -> "\o033[1;34m"
+  | `ARtoLt      -> "\o033[1;33m"
+  | `AMerge      -> "\o033[1;36m"
+  | `DiffHead    -> "\o033[1m"
+  | `DiffAdd     -> "\o033[32m"
+  | `DiffDel     -> "\o033[31m"
+  | `DiffLoc     -> "\o033[36m"
+  | _            -> ""
+
+let lineRegexp = Str.regexp "^"
+
+let colorDiff text =
+  let result = Buffer.create (String.length text) in
+  let a s = Buffer.add_string result s in
+  let p = Str.full_split lineRegexp text in
+  Safelist.iter (fun t ->
+              match t with
+                Str.Delim s -> a s
+              | Str.Text s -> (let lineSt = s.[0] in
+                               match lineSt with
+                               | '+' -> a (color `DiffAdd); a s; a (color `Reset)
+                               | '-' -> a (color `DiffDel); a s; a (color `Reset)
+                               | '@' -> a (color `DiffLoc); a s; a (color `Reset)
+                               | _   -> a s)
+            ) p;
+  Buffer.contents result
+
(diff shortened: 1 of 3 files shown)


src/uitext.ml
```diff
@@ -86,6 +86,63 @@ let setupTerminal() =
...
+  | `DiffLoc     -> "\o033[36m"
+  | _            -> ""
+
+let lineRegexp = Str.regexp "^"
+
+let colorDiff text =
+  let result = ref "" in
+  let p = Str.full_split lineRegexp text in
+  Safelist.iter (fun t ->
+              result := !result ^
```
tleedjarv, line 134: I know this is not the OCaml way. Initially, I was concerned about the recursion because the diffs could be very long. But I hadn't realized then that Safelist implements tail-recursive functions.
What is the preferred way? Is `String.concat "" (Safelist.map (...) p)` reasonable?

gdt:
It seems that sending color codes should happen if requested, vs being opt-out.

tleedjarv:
> It seems that sending color codes should happen if requested, vs being opt-out.

Are you thinking of something breaking or just user preference?

The code tries to be a bit smart, so if you're running from a script or have redirected the output then it won't enable color automatically.

I can't speak for everybody, but I think most modern software enables color automatically, or is being configured so by OS (e.g. typically by Linux distros), and active opt-out is rather the norm. It would be great to hear more from Unison users.

acolomb:
As a user I wouldn't mind color being enabled by default.

gdt:
Some GNU/Linux flavors may have a color norm, but I find that things being in color by default to be highly irregular.  We have a text UI and a GUI, and I don't  think we should default the text UI towards GUIish behavior.

As a datapoint, I haven't tried to enable color,  on NetBSD and macOS, and I basically never encounter it.

Are you saying that these Linux systems have some global "should text programs colorize output" setting, sort of like papersize?  If so, that seems fine to respect.  It's really turning on color absent 1) user request or 2) systemwide "I want color" setting that I object to.

acolomb:
Git is a prominent example of opt-out color text UI. APT on Debian / Ubuntu as well. They honor some heuristics to automatically disable it when not connected to a console directly, just as this PR does (plus NO_COLOR preference). 

Given that enabling the feature is probably hard to discover (unless you want ugly "did you know" tips on every invocation) and problems being very unlikely with the auto-disable code in place, I think it brings moderate gain to many users with very low potential to cause problems for users with very special setups / needs. Those who might run into trouble probably know enough to avoid it.

It's a **user** interface meant for humans and therefore should be as easy to parse for them as possible IMHO.

tleedjarv:
> As a datapoint, I haven't tried to enable color, on NetBSD and macOS, and I basically never encounter it.

Yes, I admit that this may be a bit more of a Linux distro thing for core system utilities (ls, for example, and default shell prompt). Even on illumos, which does not have much color by default (if any), I can think of prominent color-by-default examples like git, as pointed out by @acolomb, vim (not vi), opam, I think, too.

> Are you saying that these Linux systems have some global "should text programs colorize output" setting, sort of like papersize? If so, that seems fine to respect. It's really turning on color absent 1) user request or 2) systemwide "I want color" setting that I object to.

No, I don't think there is a global setting like this. I think it is a combination of two things: 1) most programs default to color on their own, and 2) shell prompt and some basic utils are colored in the default profile (e.g. by adding `alias ls='ls --color=auto'`).

avollmerhaus:
As a unison user with about 10 years of almost daily usage under my belt, I'd really appreciate the colours.
Having the output coloured would make it much easier to quickly eyeball the proposed changes.

I'm voting for having them on by default.
If they were disabled by default, I might not even be aware that I could enable them without following release notes or github discussions.
The absolute vast majority of terminals can handle colours with ease nowadays, and many Linux CLI programs enable them by default.
For the odd terminal without colour support or some users simply preferring b/w output it's easy to opt out, and the first case is likely already covered by automatisms.

tleedjarv:
Rebased and squashed commits.

gdt:
Coming back to this since after deferring for release.

I didn't notice before, but this is "ANSI" color codes, vs using terminfo/termcap which is what I'd expect.   Do git/emacs/etc. and other things that do color all have ANSI hard-wired in, vs using  some kind of terminfo/termcap?   Having used pre-ANSI terminals, that would surprise me, but perhaps that is how it is.   Also, I would expect that if someone had a termcap setting without color (on purpose), then no color would happen.   (My sole non-ANSI terminal does not have color.)

I don't mean to be extremely difficult; if this is doing what the other 100 programs that do color do, I don't have a problem with it.  A quick look suggests that git has this bug: "TERM=vt52 git log" still outputs color sequences.

tleedjarv:
I will investigate around terminfo/termcap and see how other implementations look like.

gdt:
If 90% of existing practice is hard-coded ANSI I don't object.   I don't actually have any clue about what is normal, just what I'd expect.

tleedjarv:
I've now looked at a few programs (ocaml itself, gcc, git, vim) and it does seem to be pretty much the way it's being done. The exception is Vim (and I would assume the same of Emacs). It does use terminfo/termcap as it needs this for its basic operation, not specifically for color. In fact, gcc has this comment in the color-related source
``` c
/* It would be impractical for GCC to become a full-fledged
   terminal program linked against ncurses or the like, so it will
   not detect terminfo(5) capabilities.  */
```

I can't judge whether it is correct or not but it seems to be at least somewhat common in such ubiquitous software.

After this admittedly brief research, one thing I will add is to check that TERM is not dumb.

tleedjarv:
Rebased and added further conditions to default color mode.

One new condition is that the `dumbtty` preference must not be enabled. This means that running in a shell inside Emacs the color will not be enabled by default.

mintsoft:
@tleedjarv in addition to the environment variable, is there a command line flag for this? I only ask as I've often had a number of issues with using ANSI colour outputting tools on Windows and ending up with horrendously corrupted/confused output and I like to be able to explicitly disable the colour output on the tool itself if possible as it's the most-clear way of doing it in my opinion.

tleedjarv:
Yes, there is a command line flag to explicitly control this.

PR #361, opened 2020-08-15
Add profile selection to text UI

tleedjarv:
### Description

If profile or roots are not specified on the command line, provide interactive profile selection to user. Also respects the `key` preference used in GUI, but only for initial profile selection.
Moved `scanProfiles` and `provideProfileKey` from `uigtk2.ml` to `uicommon.ml`

Based on the ANSI color pull request #360. This PR can be rebased once that one is merged.

### DIFF
--- a/src/TODO.txt
+++ b/src/TODO.txt
@@ -867,7 +867,7 @@ The scroll bar is not usable during transport: every time a line changes
   user to control the colors of all the arrows individually.
 
 Text mode user interface should be brought up to date with graphical
-  interface (it should prompt for profile selection, creation, root
+  interface (it should prompt for profile creation, root
   entry, etc.; command characters should be the same; ...)
 
 Since the manual is pretty big, it would be nice if the on-line version
--- a/src/uicommon.ml
+++ b/src/uicommon.ml
@@ -70,7 +70,7 @@ let profileKey =
   Prefs.createString "key" ""
     "!define a keyboard shortcut for this profile (in some UIs)"
     ("Used in a profile to define a numeric key (0-9) that can be used in "
-     ^ "the graphical user interface to switch immediately to this profile.")
+     ^ "the user interface to switch immediately to this profile.")
 (* This preference is not actually referred to in the code anywhere, since
    the keyboard shortcuts are constructed by a separate scan of the preference
    file in uigtk.ml, but it must be present to prevent the preferences module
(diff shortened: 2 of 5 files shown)


tleedjarv:
Removed dependency to ANSI color pull request. This PR is now rebased on master so that it can be merged on its own. Once merged to master, color code can be added back with a separate PR.

gdt:
I don't understand the motivation for this.  Right now unison prints a very easy to understand usage message that is aligned with command-line norms.   One can then just run it again and give a profile, which is what the person would have done in the first place had they read the documentation.

I think it's a negative for a program to start prompting for input in what is basically an error case.

tleedjarv:
The intention is to be a tiny bit more user-friendly, without being intrusive or prescriptive.

There is no impact when running from a script or the user already provides a profile and/or roots on the command line. The prompt will be created only when no profile and no roots are given. In this case if the user passes through the prompt (e.g. by just pressing Enter) or gives invalid input then the usual usage message is provided.

My use case is this. I have many profiles that I sync manually, some of which I sync very seldom. Without the prompt, I have to type each profile name manually every time I sync them, must remember the profile names and the roots they sync. Sure, I could use workarounds to reduce the manual typing, like creating very short profile names or creating aliases (or scripts) for executing Unison with each profile, but now I'd have to maintain all of that and still remember the non-descriptive aliases or short-names.

gdt:
I undestand where you are coming from in having the ability to specify these interactively.   It's just that I don't like turning the no-argument invocation into this.  I would be completely fine with "unison -query" starting this mode, and adding that to the usage.  That way you can have unisonq as an alias and get the mode you want (and that others probably find helpful), without changing the default behavior.

But I don't feel like I get to decide, and will defer to @bcpierce00 here.

bcpierce00:
I agree with Greg that the no-argument behavior should not change.

tleedjarv:
Adding an argument for starting this mode is a good idea. In that way it becomes actually discoverable.

But I would still make a case for no-argument invocation as well.

Currently this PR starts profile selection if you have at least one profile. This means that executing Unison for the first time, the user would still get the usage message, but not after that (because then the "default" profile gets created). I will make a change so that profile selection is started if you have more than one profile. Then you will never see profile selection unless you have explicitly created a profile.

Current no-argument behavior serves only two purposes - to display the usage message, and to automatically select the default profile (unless I'm mistaken). I think that with the abovementioned change, this will be a good compromise solution. It will not change anything for active users (who specify the profile or roots explicitly) and it will make sure that first-time users or infrequent-users who only have default profile still get the usage message and the default profile selected automatically.

bcpierce00:
I’m afraid I still disagree: running unison with no arguments to be reminded of the command-line format is useful no matter how many profiles you have.

gdt:
Benjamin opined that the behavior with no arguments should not change (which is what I think too).  So it seems you should follow that :-)

I am very surprised that this would create  a default profile.   I operate with no profiles, and have scripts that invoke unison.   I can see creating a named profile via interactive prompts, but not default.

Perhaps my bias is that having a default profile is not really a good idea, which is perhaps separate.

tleedjarv:
Well, at least I tried...

Just to clarify, creating the default profile is not part of this PR. It's the current behavior, and I suspect it is actually a side-effect of code changes when the first GUI was created. Or maybe not. If you always specify a profile or roots on command line then I think you avoid creation of the default profile.

tleedjarv:
Rebased and changed default behavior. Now a command-line pref is needed to invoke the interactive behavior.

I've chosen "i" as the command-line pref name (as in `unison -i`) and hinted at a "profile manager" even though it is currently just profile selection. It's because one day someone may implement more interactive profile management features (it's even in TODO.txt).

I'll take recommendations on the pref name but I prefer something short and simple and memorable.

gdt (approved):
This has been hanging out for a while with no more comments, so I think it's getting to be time to merge it.  I'll do that >=48h from now if there are no objections.

glondu (approved):

PR #362, opened 2020-08-18
Change diff ordering to compare from older to newer file

tleedjarv:
### Description

Currently the diff function works by always comparing files in the same order, e.g. from root2 to root1 with the default preference.
This PR makes it possible to compare files from older to the most recently modified one, no matter which root has the latter.
The previous way of working is retained for compatibility, but the default behaviour is changed to comparing from older to newer.

### DIFF
--- a/src/files.ml
+++ b/src/files.ml
@@ -634,14 +634,17 @@ let copy
 let (>>=) = Lwt.bind
 
 let diffCmd =
-  Prefs.createString "diff" "diff -u CURRENT2 CURRENT1"
+  Prefs.createString "diff" "diff -u OLDER NEWER"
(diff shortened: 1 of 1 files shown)


gdt:
It seems that the new feature is good and I don't expect any objections.

How do people feel about the change of default?   (I think the number of people that are being confused by reverse diffs probably outweighs the number that like it how it is now.)

I will wait for some positive comments before hitting merge, since I am not sure if the default change is controversial or not.

gdt (commented):
Seems ok to me.  Given the default change, I would like to hear from others.   I'll merge >= 18Z Sunday, earlier if comments convince me.

gdt (approved):

PR #364, opened 2020-08-21
Allow setting file modification time to 0

### LINKED ISSUE #223: -times fails to sync time of 0 (ie 1970-01-01 00:00:00)
When using -times, if a file has an actual modification time of unix epoch 0, unison will fail to synchronize the time. So eg:

    mkdir a b
    touch -d "1970-01-01 00:00:00 +0000" test
    unison -root a -root b -times

will fail with:

> Failed: Failed to set modification time of file /home/foo/b/.unison.test.ec13104661b0c45e358568993443eacc.unison.tmp to 1970-01-01 at  1:00:00 (0.000000): the time was set to 2018-10-10 at  4:02:27 (1539140547.897847) instead
Failed [test]: Failed to set modification time of file /home/foo/b/.unison.test.ec13104661b0c45e358568993443eacc.unison.tmp to 1970-01-01 at  1:00:00 (0.000000): the time was set to 2018-10-10 at  4:02:27 (1539140547.897847) instead

Epoch time 0 is fairly common for temporary or immutable files (eg Rust's package manager cargo uses it a lot for its internal package database). 

(I tried looking at the source because I assumed something like, unison uses 0 as the null value and tries to detect a failure to adjust the time by checking if the file has something besides the null value, but unfortunately the null value is itself a valid value, so unison should skip that check if the desired time value is 0 anyway. But I can't make sense of the time syncing logic, so I don't know if that's true. :/)

tleedjarv:
Closes #223 

`Unix.utimes` arguments `atime` and `mtime` must not be set to 0 at the same time due to OCaml (arguably not POSIX) behavior.
Setting the access time to just about different from zero allows to get the desired result, which is to sync all mtime values, including 0, precisely.

### DIFF
--- a/src/props.ml
+++ b/src/props.ml
@@ -534,7 +534,8 @@ let set fspath path kind t =
                Util.finalize
                  (fun()->
                     Fs.chmod abspath 0o600;
-                    Fs.utimes abspath v v)
+                    Fs.utimes abspath (if v = 0. then 1e-12 else v) v)
+                    (* See comment about this statement further below *)
                  (fun()-> Fs.chmod abspath oldPerms)
              end
            else if false then begin
@@ -557,7 +558,14 @@ let set fspath path kind t =
              let (r,_) = Lwt_unix.run (External.runExternalProgram cmd) in
              if r<>(Unix.WEXITED 0) then raise (Util.Transient "External time-setting command failed")
            end else
-             Fs.utimes abspath v v)
+             Fs.utimes abspath (if v = 0. then 1e-12 else v) v)
+             (* If atime and mtime arguments are both 0 then Unix.utimes
+                will set actual atime and mtime on the file to be the
+                current timestamp, which is not the desired result.
+                To sync the exact mtime value of 0, atime must be non-zero.
+                Setting atime to be different from zero by less than a
+                nanosecond allows to achieve the desired result.
+                https://github.com/bcpierce00/unison/issues/223 *)
   | _ ->
       ()
 


gdt:
This PR does not explain what it actually does and why that's the right thing.

tleedjarv:
The explanation is in the discussion of #223. Basically, it is a workaround for OCaml behavior, which seems to be incorrect. The idea is to prevent passing 0 access time and 0 modification time to Unix.utimes as that triggers the incorrect behavior. Since 0 modification time is desired, i.e. it is an actual use case, then access time must be set to non-zero.

gdt:
ok - but then you have to somehow invert this transform when you sync and compare back to some other system, with a lookaside database of real mtime vs fs mtime, which is what syncthing used to do.  There needs to be an argument of correctness for various operations, and this feels pretty scary to get right.   I suppose using system to call a C helper from the build might be a strategy.

glondu:
It seems that replacing `0.` by `Float.succ 0.` does the job, I would suggest to do that in this PR.

gdt:
Is the idea that by using Float.succ 0, we are passing a number that will fail the test of being = to 0, but will convert to 0 in the assignment to the C struct, and thus we work around the test in the ocaml glue, and achieve exactly the right bits in the filesystem?  And therefore we do not need to do any fixups?   If so, that sounds good to me, assuming it's explained well (reason for code as committed in comment, reason for change in commit message).

glondu:
Yes exactly, that's my idea.

tleedjarv:
> you have to somehow invert this transform when you sync and compare back to some other system, with a lookaside database of real mtime vs fs mtime, which is what syncthing used to do.

Not sure I follow. This workaround does not change mtime at all, nothing is transformed. It only prevents atime and mtime being 0 at the same time, and changes atime to achieve this guarantee. Unison has never synced atime anyway, so I don't find this to be an issue.

tleedjarv:
Using `Float.succ 0.` for atime seems like a reasonable idea (can be `Float.succ v` then, to drop the `if` completely). But I'm not sure it is better because it would set the atime to the epoch. Is that desirable?

`Float.succ` is available since OCaml 4.08.0. I guess older-OCaml equivalent would be `Float.min_float`, or potentially `0. +. Float.epsilon`, right?

glondu:
> Using `Float.succ 0.` for atime seems like a reasonable idea (can be `Float.succ v` then, to drop the `if` completely).

Maybe. But one has to be sure that `Float.succ v` doesn't return `0.`! (I don't think it would, but a convincing argument should be made.)

> But I'm not sure it is better because it would set the atime to the epoch. Is that desirable?

Setting atime to the epoch wouldn't come out of nowhere in this case: the user asked for synchronization of times, and the time on one side is epoch, so it is correct to propagate to the other side (if the user chose so).

> Unison doesn't sync atime anyway, so whatever value is set as atime is probably just as good.

Oh, I don't know the details of this.

> Are there any special considerations to take regarding atime, like some software depending on it?

No idea.

> `Float.succ` is available since OCaml 4.08.0. I guess older-OCaml equivalent would be `Float.min_float`, or potentially `0. +. Float.epsilon`, right?

I think Unison already requires OCaml >= 4.08.0 because of `Stdlib`.

gdt:
I think it would be good to avoid increasing the ocaml requirement.  That should be clear in a readme anyway.

I think it's best to apply the workaround only when needed, because this is all irregular.   So checking for 0 and adding a tiny bit makes sense.   Maybe even just hard-code 0.1 ns, to avoid the language issue, which will round to 0 even on machines with timespec for fs times.

tleedjarv:
Unison does not sync atime currently. It only syncs mtime. Whatever way this fix eventually looks like, one thing is common: mtime is not changed or transformed in any way. It continues to be synced exactly as it has always been, and now also successfully syncing value 0.

Since atime was never synced, but `Unix.utimes` does require atime as a mandatory argument, Unison has always set the atime to be the same as mtime, i.e. the current code is literally this:
``` ocaml
Fs.utimes abspath v v
```

This means that all files synced by Unison have a bogus atime anyway, and have always had. It doesn't really matter what atime they have after this fix; it was and still will be bogus. The only argument of trying to keep the atime same or as close to mtime as possible would be consistency.

In generic terms, the fix is this:
``` ocaml
Fs.utimes abspath anyBogusNonZeroValue v
```

>  But one has to be sure that `Float.succ v` doesn't return `0.`!

Actually, no. The entire fix rests on one very simple premise - atime and mtime arguments to `Unix.utimes` can't be zero at the same time AND we only care about correct value for mtime (as there is no correct atime value, unless Unison starts syncing it).
Therefore, something like this will work:
``` ocaml
Fs.utimes abspath (Float.succ v) v
```

> I think it's best to apply the workaround only when needed, because this is all irregular. So checking for 0 and adding a tiny bit makes sense. Maybe even just hard-code 0.1 ns, to avoid the language issue, which will round to 0 even on machines with timespec for fs times.

This makes a lot of sense because of consistency and simplicity. I will push an update with this change.

gdt:
It would also be nice if someone filed a bug with ocaml about this behavior.   I see no good reason for the full semantics supplied by POSIX to be unavailable in the same obvious, unconfusing way, in a language binding.

tleedjarv:
Basic test case

``` sh
$ mkdir test1 test2

$ touch -d "1970-01-01T00:00:00Z" test1/a.txt
$ touch test1/b.txt

$ stat test1/{a,b}.txt
  File: test1/a.txt
  Size: 0               Blocks: 1          IO Block: 131072 regular empty file
Device: 11200010021h/1176821104673d     Inode: 723197      Links: 1
Access: (0644/-rw-r--r--)
Access: 1970-01-01 00:00:00.000000000 +0000
Modify: 1970-01-01 00:00:00.000000000 +0000
Change: 2020-09-16 16:36:39.276223013 +0000
 Birth: 2020-09-16 16:36:39.276167344 +0000
  File: test1/b.txt
  Size: 0               Blocks: 1          IO Block: 131072 regular empty file
Device: 11200010021h/1176821104673d     Inode: 723202      Links: 1
Access: (0644/-rw-r--r--)
Access: 2020-09-16 16:36:44.652595654 +0000
Modify: 2020-09-16 16:36:44.652595654 +0000
Change: 2020-09-16 16:36:44.652602688 +0000
 Birth: 2020-09-16 16:36:44.652538243 +0000

$ unison test{1,2} -times

$ stat test2/{a,b}.txt
  File: test2/a.txt
  Size: 0               Blocks: 1          IO Block: 131072 regular empty file
Device: 11200010021h/1176821104673d     Inode: 723203      Links: 1
Access: (0644/-rw-r--r--)
Access: 1970-01-01 00:00:00.000000000 +0000
Modify: 1970-01-01 00:00:00.000000000 +0000
Change: 2020-09-16 16:37:04.067907494 +0000
 Birth: 2020-09-16 16:37:04.067655182 +0000
  File: test2/b.txt
  Size: 0               Blocks: 1          IO Block: 131072 regular empty file
Device: 11200010021h/1176821104673d     Inode: 722212      Links: 1
Access: (0644/-rw-r--r--)
Access: 2020-09-16 16:36:44.652595000 +0000
Modify: 2020-09-16 16:36:44.652595000 +0000
Change: 2020-09-16 16:37:04.092234971 +0000
 Birth: 2020-09-16 16:37:04.068050346 +0000
```

tleedjarv:
I used `1e-9` which I just realized might not be small enough, but... actually it's ok.
Currently OCaml utimes supports microsecond precision at best, and on top of that, atime has always been a non-synced bogus value. So, even keeping 'now' would be just as correct in my view.
Case in point:
``` sh
$ touch -m -d "1970-01-01T00:00:00Z" test1/c.txt
$ stat test1/c.txt 
  File: test1/c.txt
  Size: 0               Blocks: 1          IO Block: 131072 regular empty file
Device: 11200010021h/1176821104673d     Inode: 724289      Links: 1
Access: (0644/-rw-r--r--)
Access: 2020-09-16 17:01:16.256656590 +0000
Modify: 1970-01-01 00:00:00.000000000 +0000
Change: 2020-09-16 17:01:16.256711224 +0000
 Birth: 2020-09-16 17:01:16.256656590 +0000

$ unison test{1,2} -times

$ stat test2/c.txt 
  File: test2/c.txt
  Size: 0               Blocks: 1          IO Block: 131072 regular empty file
Device: 11200010021h/1176821104673d     Inode: 724064      Links: 1
Access: (0644/-rw-r--r--)
Access: 1970-01-01 00:00:00.000000000 +0000
Modify: 1970-01-01 00:00:00.000000000 +0000
Change: 2020-09-16 17:01:39.822929213 +0000
 Birth: 2020-09-16 17:01:39.822680266 +0000
```

gdt:
I realize you think that because unison does not have a correctness statement to sync atime, that this is sort of undefined.  But sort of undefined is a long way from "it's a good plan to have gratuitously different behavior because of how we are choosing to work around a bug in ocaml, when it would have been just as easy to choose a plan  that results in the behavior that would have happened if ocaml weren't buggy."  Basically, as little magic as possible should leak.

However, if the ocaml implemenntation is rounding to integer microseconds, then it seems like your code will achieve the "behave exactly as if ocaml didn't have this bug", which is great!

gdt:
I don't get the failing mac build but the patch looks good for me, so @glondu if you want to hit merge I think it's ready.

tleedjarv:
I believe the failing build is a temporary issue with GitHub. At the workflow level there are some error messages visible that read:
```
GitHub Actions has encountered an internal error when running your job.
GitHub.Services.WebApi.VssServiceResponseException: Service Unavailable
```

PR #365, opened 2020-08-22
Add signal handler for SIGUSR1 to close logfile

thingsconnected:
The logfile will be re-opened automatically as soon as something is
logged. This allows for log rotation, by first renaming existing log
file(s) and then performing:

    killall -USR1 unison

In case no log file has been opened (yet), the signal is ignored.

This feature is intended for cases where unison is running continuously
as a background (daemon) process, e.g.

    unison -repeat watch

### DIFF
--- a/src/ubase/trace.ml
+++ b/src/ubase/trace.ml
@@ -121,6 +121,16 @@ let logfile =
 
 let logch = ref None
 
+let closelog _ =
+  match !logch with
+    None -> ()
+  | Some(ch,file) ->
+      close_out ch;
+      logch := None;
+  ;;
+
+Sys.signal Sys.sigusr1 (Signal_handle closelog);;
+
 let rec getLogch() =
   Util.convertUnixErrorsToFatal "getLogch" (fun() ->
   match !logch with
--- a/src/ubase/trace.ml
+++ b/src/ubase/trace.ml
@@ -117,7 +117,9 @@ let logfile =
     "!logfile name"
     "By default, logging messages will be appended to the file
      \\verb|unison.log| in your HOME directory.  Set this preference if
-     you prefer another file.  It can be a path relative to your HOME directory."
+     you prefer another file.  It can be a path relative to your HOME directory.
+     Sending SIGUSR1 will close the logfile; the logfile will be re-opened (and
+     created, if needed) automatically, to allow for log rotation."
 
 let logch = ref None
 
--- a/src/strings.ml
+++ b/src/strings.ml
@@ -1738,7 +1738,9 @@ let docs =
       \032         By default, logging messages will be appended to the file\n\
       \032         unison.log in your HOME directory. Set this preference if you\n\
       \032         prefer another file. It can be a path relative to your HOME\n\
-      \032         directory.\n\
+      \032         directory. Sending SIGUSR1 to the unison process will close the\n\
+      \032         logfile; the logfile will be re-opened (and created, if needed)\n\
+      \032         automatically, to allow for log rotation.\n\
       \n\
       \032  maxbackups n\n\
       \032         This preference specifies the number of backup versions that\n\


gdt:
This doesn't change the documentation.   I'm not sure if that's a fair comment.

Is there  a norm (in the POSIX world, presumably) that programs behave this way on SIGUSR1, or is this custom behavior?

thingsconnected:
As far as I know you are free to use SIGUSRn in any way you like. E.g. nginx anb mongodb use it for logfile rotation (says google, page 1 of search results). Other projects use SIGHUP for this purpose, but (existing) users may expect unison to quit on SIGHUP, which is the default action, so I thought it would be best to use SIGUSR1.

The alternative is to perform a restart of unison (like e.g. Apache does when rotating logs), which I think is a bit clunky, especially when continuously monitoring a filesystem for changes.

src/ubase/trace.ml
```diff
@@ -15,6 +15,10 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 *)
 
+(* ignore SIGUSR1 by default, until the logfile is opened, where a signal
+   handler is installed *)
+open Sys;;
+signal sigusr1 Signal_ignore;;
```
tleedjarv, line 21: Why do you need to ignore the signal at all? What happens if you don't?

src/ubase/trace.ml
```diff
@@ -121,7 +125,16 @@ let logfile =
...
+let closelog _ =
+  match !logch with
+    None -> ()
+  | Some(ch,file) ->
+      close_out ch;
+      logch := None;
+  ;;
+
 let rec getLogch() =
+  ignore(signal sigusr1 (Signal_handle closelog));
```
tleedjarv, line 137: Why set the signal handler over and over again instead of setting it once (e.g. like you currently set Signal_ignore)?

src/ubase/trace.ml
```diff
@@ -15,6 +15,10 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 *)
 
+(* ignore SIGUSR1 by default, until the logfile is opened, where a signal
+   handler is installed *)
+open Sys;;
```
tleedjarv, line 20: Don't open `Sys`, use the `Sys.` notation instead.

src/ubase/trace.ml
thingsconnected: The signal handler is only installed as soon as the logfile is opened. If you don't explicitly ignore the signal, it will call the default handler and terminate the program.

tleedjarv: Ok, I see. But why don't you install the handler immediately? There is no dependency on the logfile being opened, as far as I can see.

src/ubase/trace.ml
```diff
@@ -121,7 +125,16 @@ let logfile =
...
 let logch = ref None
 
+let closelog _ =
+  match !logch with
+    None -> ()
+  | Some(ch,file) ->
+      close_out ch;
+      logch := None;
+  ;;
+
```
tleedjarv, line 135: Just to illustrate what I had in mind.
```suggestion
      logch := None

let sigusr1_behavior = signal sigusr1 (Signal_handle closelog)
```
And then of course remove the `Signal_ignore` and installing the handler in function `getLogch`.

src/ubase/trace.ml
tleedjarv: Also, just so you know, you can use `Sys.set_signal`. Then you don't have to use `ignore` here.

thingsconnected: That would be fine as well. Given my total lack of experience in OCaml, I just went with this solution, as it compiled and worked after spending a bit too much time on it.

I have revisited the code and changed it to install the signal handler simply once. The commit has been pushed onto this pull request.

thingsconnected:
> This doesn't change the documentation. I'm not sure if that's a fair comment.

I added explanation in (I hope) all the right places.

src/ubase/trace.ml
thingsconnected: Before reading this I was already busy patching and testing, please check if my latest tweaks match your expectations.

gdt:
Needs rebasing and then I think there is agreement it's ok.

Whether it goes in 2.51.3 or not is a different manner.  My own bias is that 2.51.3 is so overdue that only outright bugs with clear fixes should be fixed (and this doesn't qualify), and otherwise we should be heading to RC.   I would like to then see a 2.51.4 with as many non-breaking improvements merged, before making the breaking changes.

thingsconnected:
> Needs rebasing and then I think there is agreement it's ok.

I rebased it onto unison/master and squashed the two overlapping commits.

> Whether it goes in 2.51.3 or not is a different manner. My own bias is that 2.51.3 is so overdue that only outright bugs with clear fixes should be fixed (and this doesn't qualify), and otherwise we should be heading to RC. I would like to then see a 2.51.4 with as many non-breaking improvements merged, before making the breaking changes.

That's fine by me. It would be great to have this feature merged upstream some day.

bcpierce00:
One might also consider maintaining both 2.51 and 2.52 branches, at least for a while…

src/ubase/trace.ml
tleedjarv: No worries. Looks good to me.

gdt:
I propose to hit merge after the release.  Speak up for or against by Sunday, please.

Issue #291, opened 2019-10-18
unison does not handle extended attributes


rmrnc:
## Problem

`unison` doesn't pick up changed extended attributes.

## Reproducing steps

On macOS

1. create file
1. assign multiple (!) tags and/or a finder comment
![image](https://user-images.githubusercontent.com/8008083/67095055-4fb2c000-f1b5-11e9-8e1f-76442bb9e14c.png)
1. check corresponding extended attributes on replicas , i.e., `com.apple.metadata:_kMDItemUserTags` and `com.apple.metadata:kMDItemFinderComment` with 
```bash
> xattr -l <file>
```

```
> xattr -l <src>/A/a.txt 
com.apple.FinderInfo:
00000000  00 00 00 00 00 00 00 00 00 04 00 00 00 00 00 00  |................|
00000010  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  |................|
00000020
com.apple.metadata:_kMDItemUserTags:
00000000  62 70 6C 69 73 74 30 30 A2 01 02 55 52 65 64 0A  |bplist00...URed.|
00000010  36 57 47 72 65 65 6E 0A 32 08 0B 11 00 00 00 00  |6WGreen.2.......|
00000020  00 00 01 01 00 00 00 00 00 00 00 03 00 00 00 00  |................|
00000030  00 00 00 00 00 00 00 00 00 00 00 19              |............|
0000003c
com.apple.metadata:kMDItemFinderComment:
00000000  62 70 6C 69 73 74 30 30 5F 10 16 48 45 52 45 20  |bplist00_..HERE |
00000010  49 53 20 41 20 43 4F 4D 4D 45 4E 54 21 21 21 21  |IS A COMMENT!!!!|
00000020  21 08 00 00 00 00 00 00 01 01 00 00 00 00 00 00  |!...............|
00000030  00 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00  |................|
00000040  00 21                                            |.!|
00000042

# replica
xattr -l <dst>A/a.txt 
com.apple.FinderInfo:
00000000  00 00 00 00 00 00 00 00 00 04 00 00 00 00 00 00  |................|
00000010  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  |................|
00000020
```

Note: `rsrc = true`

bcpierce00:
[...]

PR #368, opened 2020-08-23
Add extended attributes (xattr) synchronization support

tleedjarv:
### Description

xattr synchronization support. This would also solve #291.

I've used the same method as for ACL synchronization, i.e. use `Props`. This is easy to do and it works. To prevent a potentially large number of potentially large-sized attributes from consuming memory and increasing archive size on disk, a fingerprinting and on-demand reading approach is used (see discussion below in comments).

### Platform support

Almost all platforms support extended attributes in some form. The OCaml code in `props.ml` is essentially universal synchronization of a list of name=value pairs. Platform-specific code is in `props_xattr.c`. Platforms don't have to understand other platforms' xattrs, as long as the names don't clash and other constraints are not violated. This is useful for hub-spoke synchronization.

### DIFF
--- a/doc/unison-manual.tex
+++ b/doc/unison-manual.tex
@@ -2180,6 +2180,70 @@ group of the server process.
 \end{itemize}
 
 
+\SUBSECTION{Extended Attributes - xattrs}{xattrs}
+
+Unison allows synchronizing extended attributes on platforms and
+filesystems that support them. System attributes are not synchronized.
+What exactly is considered a system attribute is platform-dependent.
+Synchronization is possible cross-platform, but see caveats below.
+
+If one of the replicas does not support extended attributes then
+Unison will not attempt attribute synchronization. If the other
+replica does support extended attributes then those will remain intact.
+
+If both replicas support extended attributes then you can request
+Unison to try attribute synchronization (\verb|xattrs| preference).
+Extended attributes from both replicas will not be merged, all extended
+attributes are propagated as a set from one replica to another.
+
+Unison currently supports extended attributes on the following platforms:
+\begin{itemize}
+\item {\em Linux}
+Attributes in user, trusted and security namespaces. Synchronization of
+the latter two namespaces depends on \verb|unison| process privileges
+and is disabled by default. To sync one or more attributes in the security
+namespace, for example, you can set the preference
+\verb|xattrignorenot| to \verb|Path !security.*| (for all) or to
+\verb|Path !security.selinux| (for one specific attribute).
+\item {\em Solaris, OpenSolaris and illumos-based OS (OpenIndiana, SmartOS,
+  OmniOS, etc.)}
+\item {\em FreeBSD, NetBSD}
+Attributes in user namespace.
+\item {\em Darwin (macOS)}
+\end{itemize}
+Not all filesystems on the listed platforms may support extended attributes.
+
+\noindent {\it Caveats:}
+\begin{itemize}
+\item Some platforms and file systems support very large extended attribute
+values. Unison synchronizes only up to 16 MB of each attribute value.
+\item Attributes are synchronized as simple name-value pairs. More complex
+extended attribute concepts supported by some platforms are not synchronized.
+\item On Linux, attribute names always have a fully qualified form
+(\texttt{namespace.attribute}). Other platforms do not have the same constraint.
+The consequence of this is that Unison will sync the attribute names on Linux
+as follows: an \verb|!| is prepended to the namespace name, except for the
+\verb|user| namespace; the \verb|user.| prefix is stripped from attribute names
+instead. This allows syncing extended attributes from Linux to other platforms.
+These transformations are reversed when syncing {\em to} Linux, resulting in
+correct fully qualified attribute names.
+The \verb|xattrignore| and \verb|xattrignorenot| preferences work on the
+transformed attribute names. This means that any patterns for the user
+namespace must be specified without the \verb|user.| prefix and any patterns
+intended for other namespaces must begin with an \verb|!|.
+\end{itemize}
+
+The \verb|xattrignore| preference can be used to filter the names of extended
+attributes that will be synchronized. The most useful ignore patterns can
+be constructed with the \verb|Path| form (where shell wildcards \verb|*| and
+\verb|?| are supported) and with the \verb|Regex| form. The
+\verb|xattrignorenot| preference can be used to override \verb|xattrignore|.
+
+Disabling the security and trusted namespaces on Linux is achieved by setting
+a default \verb|xattrignore| pattern of
+\texttt{Regex !(security|trusted)[.].*}.
+
+
 \SUBSECTION{Cross-Platform Synchronization}{crossplatform}
 
 If you use Unison to synchronize files between Windows and Unix
--- a/src/.depend
+++ b/src/.depend
@@ -817,6 +817,7 @@ props.cmi : \
     uutil.cmi \
     ubase/umarshal.cmi \
     ubase/prefs.cmi \
+    pred.cmi \
     path.cmi \
     osx.cmi \
     fspath.cmi
--- a/src/.depend
+++ b/src/.depend
@@ -787,6 +787,7 @@ props.cmo : \
     ubase/umarshal.cmi \
     system.cmi \
     ubase/safelist.cmi \
+    propsdata.cmi \
     ubase/prefs.cmi \
     pred.cmi \
     path.cmi \
@@ -803,6 +804,7 @@ props.cmx : \
     ubase/umarshal.cmx \
     system.cmx \
     ubase/safelist.cmx \
+    propsdata.cmx \
     ubase/prefs.cmx \
     pred.cmx \
     path.cmx \
@@ -821,6 +823,15 @@ props.cmi : \
     path.cmi \
     osx.cmi \
     fspath.cmi
+propsdata.cmo : \
+    ubase/util.cmi \
+    ubase/safelist.cmi \
+    propsdata.cmi
+propsdata.cmx : \
+    ubase/util.cmx \
+    ubase/safelist.cmx \
+    propsdata.cmi
+propsdata.cmi :
 recon.cmo : \
     uutil.cmi \
     ubase/util.cmi \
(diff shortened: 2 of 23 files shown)


tleedjarv:
I have not tested cross-platform xattr synchronization but in theory it should work on all platforms, except for Linux. Also, if binary data is stored in xattrs, the platform endianness may become an issue (but won't stop the synchronization).

bcpierce00:
> @bcpierce00 what is your suggestion here? I can think of a couple of options. Continue using Props, but set reasonable limits (potentially somewhat tunable by user preferences). Continue using Props, but change update and transport code so that only xattr fingerprints need to be stored in archive. Make current Osx code more generic so that it can be used for generic additional data, like resource forks, extended attributes or alternative data streams. Or yet some other way.

I don't have a clear sense of the tradeoffs here, but keeping the changes fairly local is obviously attractive.  My first thought was the same as your second one: use Props, but store only fingerprints in the archive.

tleedjarv:
I will investigate and experiment a bit more. But I doubt that I can generate large enough dataset to actually distinguish any advantages/disadvantages.

tleedjarv:
I've been thinking about this issue and summarized my thoughts in this comment.

The current solution just reads the xattrs and their values for every file every single time. All of that data is stored in the archive for update detection. My concern with this approach was excessive IO (to both read and store the values) and excessive memory usage.

**Excessive memory usage**
For 100 000 files:
16 byte fingerprints total to 1.5 MB, regardless of xattr sizes.
64 bytes for xattrs (perhaps an SELinux context) per file total to 6.5 MB.
4 KB for xattrs per file (which I find very unlikely in practice) total to around 400 MB.
For 1 000 000 files, same totals: 15 MB, 65 MB and 4 GB, respectively.

All of the numbers above are for one copy of data. In reality, worst case, up to 4 copies of data are kept in memory: the archive and the current fs state for each replica.

Totals in memory then: 6 MB, 26 MB, 1.6 GB for 100 000 files and 60 MB, 260 MB and 16 GB for 1 000 000 files.

**Discussion:** It doesn't seem like 100 000 files are a concern at all. With 1 000 000 files, up to few hundred bytes per file for xattrs is also of no real concern. It is only the real extremes that may cause problems: very large xattr values and very large number of files. Probably not your average Unison user.

As a comparison, I've guesstimated an average of 250-300 bytes to keep an ACL for each file, which would come out at 1 GB, worst case (1 000 000 files, 4 data copies in memory).

**Excessive IO**
Total IO size is the same as above, to read xattrs from files. Double that, to read the archive. And double once more, if both replicas are on the same computer/disks.
But there is a bit of a catch here. Xattr data has probably been cached already when `stat`ing the file (in the most common case all the xattrs fit in one filesystem block). Total IO operations will probably not increase. Worst case, total IO operations increase by 1 per file. This may present an issue with 1 000 000 files on rotating drives.

**Discussion:** Based on numbers above, and best case scenario, the increase in IO would be next to none. Worst case, one or more IO ops per file is added and that may quickly become a bottleneck on rotating media. Theoretically.


### Is optimization needed or not?

_In theory_, there is not enough reason to change the current solution. Up to a few hundred bytes for xattrs on 1 000 000 files should be easily managed. The only change: current build-time limit of 4 KB should probably be brought down to around 200-300 bytes with a user preference to increase it "if you know what you are getting into".

But, I don't have any empirical data on the average number or size of extended attributes nor real-life Unison repository sizes.


### What optimizations are possible?

I figured out a way to, with hopefully very minimal code changes outside `Props` module, separate the metadata for update detection and actual xattr data for synchronization.

For metadata, my thinking goes like this:
(_I will leave Solaris/illumos out of this discussion, because on these platforms xattrs are literally regular files and don't suffer from same issues._)
- The first bit of info we get is the attribute names. That alone is not sufficient for update detection.
- Next, we get the lengths of attribute values. That still is not sufficient.
- Next, we have to read the entire attribute value and compare it directly (if stored in the archive) or fingerprint it for comparison.
- As the only option to limit having to read in all xattr values, I could think of `ctime`. Anything could have changed besides xattrs (even properties Unison does not track at all), but `ctime` can be used to identify files that clearly don't have xattrs (and other props) changed. I don't know if this is true on all platforms.
- Filtering by `ctime` change will limit files for which xattrs must be read.
- As far as I can tell, there are no more shortcuts to take.

These are the two potential optimizations/metadata options, to be used alone or in combination. How do they measure up in light of the original concerns?

- **Filtering by `ctime` change**
This by itself would not reduce archive size at all. On the contrary, `ctime` now also needs to be recorded in the archive.
It greatly reduces the needed IO and runtime memory needs during update detection. Assuming only 10% of the files have `ctime` changed, the potential extra IO (which may well be 0 anyway, for most common cases) and memory for current fs state would be reduced by 90%; but no practical change for the archive.

- **Fingerprinting**
Fingerprint will require 16 bytes for each file. A bit of a trade-off is that this 16 bytes must always be stored, so it is an increase for files that don't have any xattrs (could be remedied with an empty variant). Assuming larger xattrs, fingerprinting will reduce the size/memory needs.
By itself, fingerprinting will not reduce IO requirements for reading current fs state.

`ctime` change optimization will not have any benefits on first synchronization (where all files are "updated"), but fingerprinting has the chance to reduce memory requirements if implementation is done similarly to file synchronization so that xattr values are not kept in memory all the time.

Combination of `ctime` and fingerprinting will result in having to scan fewer files and having potentially less information in memory and in archive.

IO and memory, fine, but what about actual running time? This is very difficult to reason about without actual testing, but with xattrs less than 4 KB or even 8 KB per file, I don't expect any major disadvantages or advantages when it comes to Unison running time.


### Conclusion

I am currently conservative about starting to optimize anything here. There seems to be no real issue for common cases, unless I'm not seeing something.

More empirical data is needed to understand real replica and xattrs sizes.

_Anyone interested in syncing extended attributes can help by providing some real world figures about their replicas and xattrs._

@bcpierce00 do you have an idea of typical replica sizes (number of files and directories) of Unison users?

tleedjarv:
Rebased and refactored a bit. Uses the `ctime` optimization added in the ACL PR. Reduced the max size of attribute values.

pamolloy:
I've been using unison for years and the lack of xattr support has definitely been a pain point. I finally got a chance to test this PR. I was able to synchronize an existing directory with an empty one and the xattrs were copied.

When I tried to synchronize between two existing directories (for the first time with `-ignorearchives`) unison took about 7 minutes. Every time I run with `-xattr` it skips all files because `properties changed on both sides`.

tleedjarv:
@pamolloy thank you for testing.

I'm in the process of rebasing this code on the latest (umarshal+feature flags) version, which also makes it fully backwards compatible. I'm rewriting quite a bit during the rebase, and that includes better error reporting. It'll take some time to complete, though.

> I'm not sure what "You can debug "props+" to see more details" means.

It means to add a `debug` preference with value `props+`. On the command line `-debug props+` or in a profile `debug = props+`. I'll improve this message.

pamolloy:
> I'm in the process of rebasing this code on the latest (umarshal+feature flags) version, which also makes it fully backwards compatible. I'm rewriting quite a bit during the rebase, and that includes better error reporting. It'll take some time to complete, though.

Thanks for the reply! And I really appreciate the work you're doing. Feel free to ping me whenever you have something that could use some testing.

> > I'm not sure what "You can debug "props+" to see more details" means.
> 
> It means to add a `debug` preference with value `props+`. On the command line `-debug props+` or in a profile `debug = props+`. I'll improve this message.

Thanks!

pamolloy:
@bcpierce00 I'd be happy to put a bounty on this feature. However I imagine wading through the sea of bounty websites and then integrating it into the repo may not be worth the effort. In case you're interested in setting that up, I'd be happy to contribute.

pamolloy:
Having a bit of trouble with building for my backup server that is running Bullseye i686. But I was able to test between a desktop and laptop running Fedora 36.

Running unison with `-ignorearchives` took about 20 minutes, if I recall correctly. And there were several conflicts caused by inconsistent properties, which seems normal. I manually tried to resolve them with `setfattr`, but unison still complains about differences.

```
props    <-?-> props      posts/archive.md  
local        : changed props      modified on 2022-07-30 at 16:03:48  size 1813      rw-rw-r-- 2 xattrs (67 bytes)
remote       : changed props      modified on 2022-07-30 at 17:07:47  size 1813      rw-rw-r-- 2 xattrs (67 bytes)
```

I tried a `rsync -E` as well with no luck.

tleedjarv:
> Running unison with `-ignorearchives` took about 20 minutes, if I recall correctly. And there were several conflicts caused by inconsistent properties, which seems normal.

Yes, syncing existing data will do that, unless already equal.

> I manually tried to resolve them with `setfattr`, but unison still complains about differences.
> 
> ```
> props    <-?-> props      posts/archive.md  
> local        : changed props      modified on 2022-07-30 at 16:03:48  size 1813      rw-rw-r-- 2 xattrs (67 bytes)
> remote       : changed props      modified on 2022-07-30 at 17:07:47  size 1813      rw-rw-r-- 2 xattrs (67 bytes)
> ```

Have you checked with `getfattr` that the xattrs on this file in both replicas are actually equal? You can also try to sync with the `xattrs` preference disabled to see if the props conflict is caused by xattrs or by some other properties (I see that times are different, for example).

> I tried a `rsync -E` as well with no luck.

You mean `rsync -X`?

tleedjarv:
Ready to be merged.

pamolloy:
Glad that this was merged! Just rebuilt on all my hosts including the following two:

```
$ unison -version
unison version 2.52.1 (ocaml 4.11.1)
$ ssh server unison -version
unison version 2.52.1 (ocaml 4.11.1)
```

When I run normally:

```
Fatal error: Archive file seems damaged (sum2: invalid tag): use the -ignorearchives option, or throw away archives on both machines and try again
```

When I run with `-ignorearchives`:

```
Fatal error: Internal error: New archives are not identical.
Retaining original archives.  Please run Unison again to bring them up to date.
```

tleedjarv:
@pamolloy Could you please open a new issue about it?

It is likely that the versions on the machines are still not the same. Even though there is now compatibility between different versions, this only applies for release versions and may not apply between development versions. So make absolutely sure that the version that you want to run is the one actually executed.

Next step is to run with `-debug verbose`. But please collect any additional information into a new issue and include details like platform/OS, CPU arch.

pamolloy:
See #787

PR #371, opened 2020-09-07
Improve actions workflow configuration

smorimoto:
This PR removes some unnecessary configuration and adds some changes.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -1,8 +1,8 @@
 name: Main workflow
 
 on:
-  - push
   - pull_request
+  - push
 
 jobs:
   build:
@@ -13,11 +13,10 @@ jobs:
           - macos-latest
           - ubuntu-latest
         ocaml-version:
-          - 4.10.0
+          - 4.10.1
           - 4.09.1
           - 4.08.1
-          - 4.08.0
-          - 4.07.0
+          - 4.07.1
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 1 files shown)


smorimoto:
There is no specific command, but we can run `opam upgrade --fixup` to check the consistency of installed packages. At this point, we're not caching the opam switch in a very sound way, so having this step is a good idea.

.github/workflows/workflow.yml
```diff
@@ -13,41 +13,39 @@ jobs:
           - macos-latest
           - ubuntu-latest
         ocaml-version:
-          - 4.10.0
+          - 4.10.1
           - 4.09.1
           - 4.08.1
-          - 4.08.0
-          - 4.07.0
+          - 4.08.1
```
glondu, line 19: This adds 4.08.1 a second time. I think this is an error.

smorimoto: Indeed... I will fix it.

.github/workflows/workflow.yml
```diff
@@ -13,41 +13,38 @@ jobs:
...
-        if: runner.os != 'Windows'
+        id: cache-opam
         with:
           path: ~/.opam
-          key: ${{ matrix.os }}-ocaml_${{ matrix.ocaml-version }}-opam
-
-      - name: Set up opam on macOS
-        if: steps.cache-opam.outputs.cache-hit == 'true' && runner.os == 'macOS'
-        run: /usr/local/bin/brew install opam
+          key: v1-${{ matrix.os }}-opam-${{ matrix.ocaml-version }}
```
smorimoto, line 32: > Caches are immutable so it is useful to start all your cache keys with a version prefix, for example `v1-...` This enables you to regenerate all of your caches by incrementing the version in this prefix. [Caching Dependencies: Managing Caches - CircleCI](https://circleci.com/docs/2.0/caching/#managing-caches)

PR #369, opened 2020-08-29
Prepare release 2.52.0

glondu:
I intend to merge this PR in two weeks, if nobody objects. When merged, I intend to tag the resulting master as 2.52.0 and make a "release" on GitHub with the head of src/NEWS (letting GitHub generate the tarballs).

PR #376, opened 2020-09-14
Prepare release 2.51.3

glondu:
This replaces #369.


glondu:
I intend to merge and tag as v2.51.3 this weekend unless some objection is raised.

bcpierce00:
Great.

gdt:
Are we sure there are no interop problems?  Merging the fix is great, but I'd like to see an RC people can test, and that packagers can draft update a package too.   I don't mean to be super difficult, but the pain of protocol breaks is enormous for people that don't build unison from sources all the time themselves.

That said, if it's "released" and trouble, we can then say "don't use 2.51.3 - use 2.51.4".    My point is really just that we need to get to safe, stable point -- that we can demand all bug report submitters update to -- before the big changes.

glondu:
> Are we sure there are no interop problems?

As far as I can tell, no.

> I'd like to see an RC people can test

So are you suggesting to wait a bit before tagging "v2.51.3" after the merge? Maybe tagging a "v2.51.3_rc1"? And advertising it on unison-{users,hackers} mailing-lists, and wait, say, 1 month before making the actual release?

> That said, if it's "released" and trouble, we can then say "don't use 2.51.3 - use 2.51.4".

I would prefer that, but of course I am biased as I am eager to merge my "big changes".

> we need to get to safe, stable point -- that we can demand all bug report submitters update to -- before the big changes

I totally agree.

bcpierce00:
If enough people are testing the release candidate, then even a week or two is probably enough to flush out most problems.

gdt:
Yes, I mean to merge the fix, and tag it as v2.51.3rc1 -- just like it was a release, but with that rc1 tag.  And then on the list, ask everyone to test.  I don't think we need to wait a month, but at least a week, if we get a bunch of "I tried it and it works fine with my other systems that are running v2.51.2" and two weeks if we don't.

It would be great if we had some kind of regression test suite that was able to test against the previous release, but given that we aren't talking about whether that theoretical test suite passes or not it seems we don't (and I get it that it's hard to create that).

glondu:
If we go the release candidate route, then maybe I can do this (merging, tagging rc and announcing on the lists) earlier than this weekend (maybe today)?

gdt:
I'm fine with an immediate RC.  That's the whole point - if it's not ok, we just make rc2, no harm done.

tleedjarv (commented):
@glondu, I've been thinking what needs to be done to correct and re-apply the reverted patch. I've based the below on assumption that Unison only supports Latin1 and Unicode.

- **case.ml** and **ubase/util.ml**: Here, `lowercase_ascii` is not correct. Since `lowercase` has been deprecated in Stdlib then the proposal is to copy code from https://github.com/ocaml/ocaml/blob/trunk/stdlib/char.ml#L49-L54 directly to Unison codebase. This is correct if the Latin1 assumption is maintained.
- **transport.ml**: I don't get why `uppercase` is used here in the first place. I think it should be `capitalize`. Either way, `*_ascii` is perfectly ok here.
- **uigtk2.ml**: `capitalize_ascii` is perfectly ok here.
- **unicode.ml**: Just looking at the code, I can't determine if `lowercase_ascii` would be correct here. The comment does mention ASCII characters. The expected input is Unicode and this code path is only a fallback for invalid Unicode strings, as I understand it. In that case, it seems that `lowercase_ascii` is actually correct. I can't determine what would happen to Latin1 strings fed into these Unicode functions. They'd probably yield incorrect results outside ASCII characters range. Is that an issue? Probably not, as this code is executed on invalid Unicode input and it has no idea what character set it may be. At the same time, to keep the previous behavior, same solution as for **case.ml** above might be thinkable.

PR #380, opened 2020-09-29
Fix builds

rivy:
- adds working windows builds (both text and GTK+2)
- adds working builds for multiple platform/OCaml-variant combinations (including static linux/ubuntu text)
- adds *automatic* (no-touch) packaging/github-release of all builds for any commit tagged by the regex form of `[Vv]?[0-9]*`
  - eg, see [GH:rivy/unison/releases](https://github.com/rivy/unison/releases)

I've used `unison` extensively for years on multiple windows and linux machines. But, I've been stuck on version 2.40.102 for a very long time. I've needed matched versions which run on windows (with a GUI) and various linux machines (including busybox-based hosts). It has been almost impossible to find that combination of executables.

So, after noticing some recent activities which made it seem that a possible solution was obtainable, I spent a couple of weeks designing this solution (composing well-specified build recipes and testing them, making required modifications to the avsm/setup-ocaml github action ([PR in-progress](https://github.com/avsm/setup-ocaml/pull/43)), and then creating a GHA workflow for `unison`).

It does work, but it's not perfect. I have extremely limited knowledge of OCaml; so the code modifications I've made to allow windows builds just ape other similar modifications in prior commits and might reasonably be described as "ugly hacks". And I don't have knowledge (or an implementation) of a working build recipe for certain build combinations, eg static gui builds for any platform.

I'm happy to refactor the changes in this PR into a more reasoned and pleasing form.

Some items about the PR:

#### Notes

- `make src ...` is used because `make` is not propagating STATIC and UISTYLE down to the subordinate 'src' makefile(s)

#### Problems

  - for windows platforms, `unison-fsmonitor.exe` (either built with UISTYLE=text or UISTYLE=gtk2) causes `unison.exe` built for UISTYLE=gtk to hang at "Looking for changes"
  - macos builds with UISTYLE=mac fails
  - linux/ubuntu builds with UISTYLE=gtk2 STATIC=true fails
  - linux/ubuntu static build (with ocaml-version X+musl+...) fails when attempting to build unison-fsmonitor
    - a makefile gate was added [based on STATIC] to avoid building unison-fsmonitor for that case

#### Questions

  - how can a static Win32 GTK2 executable be built?
  - how can a static unison-fsmonitor for linux (? macos) be built?

### `EDIT`

automation separated into new PR by request


gdt:
Please don't go on a refactor rampage until you hear from someone else too.   I have read this and it seems ok, but my ocaml knowledge is quite limited.   There are three separate things going on in this PR:
 * windows packaging
 * updating code for more modern ocaml
 * release builds/CI setup

Were I in charge, I would want to see the code fixes/updates separately, as I don't see them as related.  This is partly about cleanliness, and partly to make it easier for people to revert parts if they have to debug.

gdt:
Two testing questions:
  Did you verify that this interoperates with the current RC?

  Did you verify that with the ocaml code changes, this still compiles with the same degree of old compiler that it did before?

src/lwt/win/lwt_unix_impl.ml
```diff
@@ -351,7 +354,10 @@ if !d then Format.eprintf "Start reading@.";
 if !d then Format.eprintf "Reading started@.";
   res
 
-let write ch s pos len =
+let read ch s pos len =
+  _read_string ch (Bytes.to_string s) pos len
```
glondu, line 358: This doesn't make sense to me... I believe `read` should not be touched here, and should directly take a bytes as argument.

src/lwt/win/lwt_unix_impl.ml
```diff
@@ -363,6 +369,12 @@ if !d then Format.eprintf "Start writing@.";
 if !d then Format.eprintf "Writing started@.";
   res
 
+let write ch s pos len =
+  _write_string ch (Bytes.to_string s) pos len
```
glondu, line 373: I think it's better let `write` untouched (and take a bytes), and to add a new function `write_substring` whose contents is the same as `write` except for the needed ajustments to make it take a string.

src/system/system_win.ml
```diff
@@ -38,10 +38,13 @@ let fspathAddSuffixToFinalName f suffix = f ^ suffix
...
-    if f.[i] = '/' then Bytes.set f i '\\'
-  done;
-  f
+  let length = String.length f in
+  let buffer = Bytes.create length in
+    String.blit f 0 buffer 0 length;
+    for i = 0 to length - 1 do
+      if Bytes.get buffer i = '/' then Bytes.set buffer i '\\'
+    done;
+    Bytes.to_string buffer
```
glondu, line 47: The indenting is misleading here; please un-indent this block to make it align with its surrounding `let ... in`.

glondu (changes requested):
This has been done for the unix part in acbce4327d40842f10d286aa61579338613c91dd, you can take inspiration from it.

glondu (approved):
Looks good

glondu (approved):
OK

glondu (commented):
The commit about GHA workflow looks like a house of cards to me, but I'm not really competent to judge...

glondu:
@rivy BTW, thank you for tackling this!

I'm willing to postpone the 2.51.3 release another week or two to have this included. @gdt What do you think?

glondu:
About static linking and GTK2: I wouldn't be surprised this is not possible (or very difficult). Indeed, GTK2 itself would need to be available as a static library, which is probably frowned upon by packagers.

gdt:
> About static linking and GTK2: I wouldn't be surprised this is not possible (or very difficult). Indeed, GTK2 itself would need to be available as a static library, which is probably frowned upon by packagers.

If it's for Unix, I'm opposed.  If it's for Windows only, my advice is not to run Windows, and if you do, I really don't know what to say, so I abstain.

gdt:
> @rivy BTW, thank you for tackling this!
> 
> I'm willing to postpone the 2.51.3 release another week or two to have this included. @gdt What do you think?

I think we should discuss that on the mailinglist.

glondu:
@rivy Could you make separate PRs for 1) the OCaml fixes and 2) the other build fixes and 3) the GitHub Actions part?

src/system/system_win.ml
rivy: Just the last line, correct?

rivy (commented):
> Please don't go on a refactor rampage until you hear from someone else too. I have read this and it seems ok, but my ocaml knowledge is quite limited. There are three separate things going on in this PR:
> 
> * windows packaging
> * updating code for more modern ocaml
> * release builds/CI setup

I see it as two-ish:

1. Three fixes (in logically separate commits) that are required to get windows builds to compile for the previously specified 4.07 thru 4.10 compilers. I may be mis-remembering, but I believe that all fixes were for hard compiler stops. I wasn't trying to modernize the code per se, especially given that I no very little about OCaml.
2. Adding a more capable CICD workflow that:
  a.  builds all platforms (and extra variants, such as a STATIC linux text build) and 
  b. automatically packages the built binaries, when needed for releases
 
> Were I in charge, I would want to see the code fixes/updates separately, as I don't see them as related. This is partly about cleanliness, and partly to make it easier for people to revert parts if they have to debug.

I understand, and I can do that, but, if merged as separate commits, the commit separation gives you essentially both of those things. If I pull out the first three commits into a separate PR, the "CICD" commit will have to be based upon that set of commits to succeed. Again do-able, but I thought all of this to be logically entangled.

rivy:
> Two testing questions:
> Did you verify that this interoperates with the current RC?

The original CICD workflow testing still succeeds, and I did a couple of limited test runs on my platforms, but that's the total testing that I've been able to do with the testing that is present in the project.

I did try to make a minimally working set of code changes so as to minimize possible unintended consequences. I'm happy to modify things if anyone sees a problem, or can guide me to a better solution set.
 
> Did you verify that with the ocaml code changes, this still compiles with the same degree of old compiler that it did before?

The associated CICD builds show pass for all the prior tested versions, as well as the newly available variants (see <https://github.com/bcpierce00/unison/actions/runs/277695413>). Is that what you mean?

rivy:
> The commit about GHA workflow looks like a house of cards to me, but I'm not really competent to judge...

Well, fair point, but, in my defense, this is hard. I tried to create a well-specified build process that works for all supported platforms. And, if it falls over, it'll be obvious and, anyway, it's just the CICD. Changing it or even ripping it out won't really much affect anyone but the devs. 

Although, it bears repeating, this gives the project *automatic* release packaging for all platforms.

rivy:
> > About static linking and GTK2: I wouldn't be surprised this is not possible (or very difficult). Indeed, GTK2 itself would need to be available as a static library, which is probably frowned upon by packagers.

Static builds are needed for lower capability appliances, such as NASes (QNAP, Synology, etc) or IoT devices, which mostly don't have the ability to install much of any support packages. GUI would be nice on those as well, but it's not a deal breaker. I had come across some static GTK-style builds, but it may have only been for windows (via MSVC compiles?).

> If it's for Unix, I'm opposed. If it's for Windows only, my advice is not to run Windows, and if you do, I really don't know what to say, so I abstain.

Well, I'm a linux/windows polyglot, using both quite regularly, like, I expect, literally *tons* of other users. I'm hoping with this PR to make the Windows story more complete and available for users, while, simultaneously, making things easier on the devs, with more automatic functionality.

rivy:
> @rivy Could you make separate PRs for 1) the OCaml fixes and 2) the other build fixes and 3) the GitHub Actions part?

I'll look at it later today. Though, as I just mentioned in a prior response, the CICD commit is predicated on fixed windows builds and fails without them, so the PR would really need to be based on-top of the other commits.

src/lwt/win/lwt_unix_impl.ml
rivy: The CICD portion took the most mental time in the last couple of weeks, so I'm fuzzy about this choice. But I believe it was a combination of minimizing code duplication (especially for `write` and `write_substring`), seeing relatively easy conversion of Bytes to Strings (`(Bytes.to_string s)`) vs the scary reverse conversion using `unsafe_blit_to_bytes`, and getting the functional interface correct for all the used locations.

I'll will look at it and see if I can refactor it into a form more closely aligned with your suggestion.

src/lwt/win/lwt_unix_impl.ml
rivy: ... see reponse to `write` below. I'll look at it and see if I can refactor it to a form more in line with your suggestion.

gdt:
I see your point.  But do the compiler adjustments affect people that aren't running windows?  If they are solely in code that is #ifdef windows, then I don't mind.    And it's not a big deal.

Part of where I'm coming from is to sort things into "affects unix" and "doesn't affect unix" and not really worry about the "doesn't  affect unix part".   The ocaml changes didn't in my mind fall into that bucket, but I could be misreading.

gdt:
> > Two testing questions:
> > Did you verify that this interoperates with the current RC?
> 
> The original CICD workflow testing still succeeds, and I did a couple of limited test runs on my platforms, but that's the total testing that I've been able to do with the testing that is present in the project.
> 
> I did try to make a minimally working set of code changes so as to minimize possible unintended consequences. I'm happy to modify things if anyone sees a problem, or can guide me to a better solution set.

OK - but it's critical that we not break interop, so I think someone needs to run a build of this against a build fo master without this, just syncing some things, and verifying that the new build can read the old archive files.  I am NOT demanding that CI test this.

> > Did you verify that with the ocaml code changes, this still compiles with the same degree of old compiler that it did before?
> 
> The associated CICD builds show pass for all the prior tested versions, as well as the newly available variants (see https://github.com/bcpierce00/unison/actions/runs/277695413). Is that what you mean?

sounds good, thanks.

gdt:
> > > About static linking and GTK2: I wouldn't be surprised this is not possible (or very difficult). Indeed, GTK2 itself would need to be available as a static library, which is probably frowned upon by packagers.
> 
> Static builds are needed for lower capability appliances, such as NASes (QNAP, Synology, etc) or IoT devices, which mostly don't have the ability to install much of any support packages. GUI would be nice on those as well, but it's not a deal breaker. I had come across some static GTK-style builds, but it may have only been for windows (via MSVC compiles?).

The notion of lower-capabilty and GUI is funny to me, but yet I believe people want it.

I guess if there is a default-off option to static link, then I don't mind.   From a packaging viewpoint, it's irregular and a security concern, so I think it should be limited to those who opt in.   I think though that this is what you've done.

> > If it's for Unix, I'm opposed. If it's for Windows only, my advice is not to run Windows, and if you do, I really don't know what to say, so I abstain.
> 
> Well, I'm a linux/windows polyglot, using both quite regularly, like, I expect, literally _tons_ of other users. I'm hoping with this PR to make the Windows story more complete and available for users, while, simultaneously, making things easier on the devs, with more automatic functionality.

Sure, I think that's all fine.  DIdn't mean to imply otherwise.  What I meant is that I really do no use windows, and don't consider myself competent to give an opinion.

rivy:
> I see your point. But do the compiler adjustments affect people that aren't running windows? If they are solely in code that is #ifdef windows, then I don't mind. And it's not a big deal.
> 
> Part of where I'm coming from is to sort things into "affects unix" and "doesn't affect unix" and not really worry about the "doesn't affect unix part". The ocaml changes didn't in my mind fall into that bucket, but I could be misreading.

Gotcha, I don't think the commits make any changes to unison code for linux builds. The code changes (in the second two "fix" commits) are localized to files which only build for windows (*src/lwt/win/lwt_unix_impl.ml* and *src/system/system_win.ml*). The first "fix" does modify a commonly used file (*src/Makefile.OCaml*), but the changes are generally gated to Cygwin/windows builds. The resource compilation does look to be in a communal spot, but it should only happen for windows compilations, and probably could actually be gated for just windows.

I'll go over the three fix commits and try to make sure the changes really are a minimal set for a working windows build.

I do see a spot that gated out compilation of `unison-fsmonitor` for linux STATIC=true builds, because of the build failures that I noted above in the original posting... Thoughts about that?

gdt:
If the changes are only in windows files, then that's ok.  Would be good for the commit messages to say that.

As for turning off things that fail, I'm basically opposed.  I'd rather not hide problems.

rivy:
> > Static builds are needed for lower capability appliances, such as NASes (QNAP, Synology, etc) or IoT devices, which mostly don't have the ability to install much of any support packages. GUI would be nice on those as well, but it's not a deal breaker. I had come across some static GTK-style builds, but it may have only been for windows (via MSVC compiles?).
> 
> The notion of lower-capabilty and GUI is funny to me, but yet I believe people want it.

Yes, I can see where that would be a bit ironic. 😄  ... But I have had to manually guide a few syncs between devices which involved a lot of files, and the GUI was invaluable (especially for NASes). `rust` (default to fully static builds) has gotten me used to being able to use full-feature utilities on just about any device/platform; it's really awesome!

rivy:
> > I do see a spot that gated out compilation of unison-fsmonitor for linux STATIC=true builds, because of the build failures that I noted above in the original posting... Thoughts about that?
>
> As for turning off things that fail, I'm basically opposed. I'd rather not hide problems.

I am also a fan of fix over bypass.

In this case, I'm not sure that `STATIC=true` actually compiles for linux builds. For me, using ubuntu 18.04 and ocaml-v4.10.0, `opam exec -- make UISTYLE=text STATIC=true` fails with:

```text
/usr/bin/ld: dynamic STT_GNU_IFUNC symbol `ceil' with pointer equality in `/usr/lib/x86_64-linux-gnu/libm-2.27.a(s_ceil.o)' can not be used when making an executable; recompile with -fPIE and relink with -pie
collect2: error: ld returned 1 exit status
File "caml_startup", line 1:
Error: Error during linking
Makefile.OCaml:419: recipe for target 'unison' failed
```

Has anyone seen a `STATIC=true` configuration that compiles (besides the construction I'm including here)?

If it has never compiled, I'm really not the one to fix it. I need a lot more OCaml language knowledge before I'd attempt it.

glondu:
> In this case, I'm not sure that `STATIC=true` actually compiles for linux builds. For me, using ubuntu 18.04 and ocaml-v4.10.0, `opam exec -- make UISTYLE=text STATIC=true` fails with:
> 
> ```
> /usr/bin/ld: dynamic STT_GNU_IFUNC symbol `ceil' with pointer equality in `/usr/lib/x86_64-linux-gnu/libm-2.27.a(s_ceil.o)' can not be used when making an executable; recompile with -fPIE and relink with -pie
> collect2: error: ld returned 1 exit status
> File "caml_startup", line 1:
> Error: Error during linking
> Makefile.OCaml:419: recipe for target 'unison' failed
> ```
> 
> Has anyone seen a `STATIC=true` configuration that compiles (excluding the construction I'm including here)?

On Debian testing, it compiles and I get static binaries, but the build (rightfully) complains about the use of "dynamic" features of libc like getpwnam. These functions inherently need dynamic linking (think about NSS plugins), so statically linking Unison is broken (at least on Linux) IMHO. I don't know about other unixes, though.

Note that the compiled text version of Unison only depends on libc (at least on Linux (Debian)), I don't really see the point of static linking.

rivy:
Refactored based on PR feedback.

The first two "fix" commits are windows-only code changes and are required to successfully build windows binaries.

The last "fix" commit is a gate added *Makefile.OCaml* avoiding a fsmonitor build for any MUSL-type ocaml variant. It appears that fsmonitor uses some functions which are incompatible/not-included when linking to the MUSL library.

src/system/system_win.ml
rivy: Done.

src/lwt/win/lwt_unix_impl.ml
```diff
@@ -340,8 +343,9 @@ external start_read :
 external start_write :
   file_descr -> buffer -> int -> int -> int -> unit = "win_write"
 
-let read ch s pos len =
+let read ch b pos len =
 if !d then Format.eprintf "Start reading@.";
+  let s = (Bytes.to_string b) in
```
glondu, line 348: I don't see how this can work. You are creating a copy of b, reading into that copy and then discarding the copy. You have to use directly `read`'s argument.

Looking at the wider picture, it seems you have to change the string argument of `blit_buffer_to_string` to bytes (blitting to a string does not make sense with safe strings). And, possibly, rename that function to `blit_buffer_to_bytes` to avoid confusion.

src/lwt/win/lwt_unix_impl.ml
```diff
@@ -363,6 +367,9 @@ if !d then Format.eprintf "Start writing@.";
 if !d then Format.eprintf "Writing started@.";
   res
 
+let write ch b pos len =
+  write_substring ch (Bytes.to_string b) pos len
```
glondu, line 371: Here, you are copying `b`, which can be costly and avoided. I would put into `write`'s body the same as `write_substring`, with the call to `blit_string_to_buffer` replaced with `blit_bytes_to_buffer`.

src/system/system_win.ml
```diff
@@ -38,10 +38,13 @@ let fspathAddSuffixToFinalName f suffix = f ^ suffix
...
-  for i = 0 to String.length f - 1 do
-    if f.[i] = '/' then f.[i] <- '\\'
-  done;
-  f
+  let length = String.length f in
+  let buffer = Bytes.create length in
+    String.blit f 0 buffer 0 length;
+    for i = 0 to length - 1 do
+      if Bytes.get buffer i = '/' then Bytes.set buffer i '\\'
+    done;
```
glondu, line 46: The indentation is confusing here, all these lines should be indented the same way.

glondu (changes requested):

src/system/system_win.ml
rivy: I pushed a "fix!" commit changing this... if it's not the correct indentation, please post a "suggested change" and I'll update it.

src/lwt/win/lwt_unix_impl.ml
rivy: I'll attempt this, but I will probably need some help.
I know essentially zero about OCaml syntax, so seeing the bigger picture is basically by guess and inference for me.

rivy: It does seem to be working as currently written. Synchronizations are completing correctly.

src/lwt/win/lwt_unix_impl.ml
rivy: I don't know how to construct `blit_bytes_to_buffer`. Can you suggest the correct code change?

src/lwt/win/lwt_unix_impl.ml
glondu: It might work, but it's not correct or future-proof.

src/lwt/win/lwt_unix_impl.ml
glondu: > I don't know how to construct `blit_bytes_to_buffer`.

You can construct `blit_bytes_to_buffer` by copying and adapting `blit_string_to_buffer`. For now, it is an external, and maybe just adapting the type works. But this is not future-proof. Ideally, the copying/adapting should be done with the C functions as well.

> Can you suggest the correct code change?

Are there instructions on how to reproduce the build environment locally?

src/system/system_win.ml
glondu: It is good now.

src/lwt/win/lwt_unix_impl.ml
rivy: > It might work, but it's not correct or future-proof.

Not arguing that point, but I don't have the OCaml knowledge to design a correct replacement.

src/lwt/win/lwt_unix_impl.ml
rivy: Yes, I've got a detailed recipe that I'm working into a PR for the Windows PC READMEs ...

- install Cygwin

```batch (CMD shell)
@rem # CMD shell

@rem # configuration/customization (user may alter at their discretion)
set "CYGWIN_ARCH=x86_64" &@rem # [x86,x86_64] ; x86/32-bit can be used for 32-bit build machines (if needed), but the x86_64/64-bit version can build either 32-bit or 64-bit binaries
set "CYGWIN_ROOT=c:\cygwin"
set "CYGWIN_URL=http://cygwin.mirror.constant.com"

@rem # basic set of default packages for the Cygwin/OCaml toolchain
set "PACKAGES=curl,diffutils,dos2unix,git,m4,make,mercurial,patch,perl,rsync,unzip,mingw64-i686-gcc-core,mingw64-x86_64-gcc-core"
if /i "%CYGWIN_ARCH" == "x86_64" set "PACKAGES=%PACKAGES%,mingw64-x86_64-gcc-core"

@rem # 'lablgtk' prerequisite packages
set "PACKAGES=%PACKAGES%,mingw64-i686-glib2.0,mingw64-i686-gtk2.0,mingw64-i686-pango1.0,libgtk2.0_0,libgtk2.0_devel,ocaml-lablgtk2"
if /i "%CYGWIN_ARCH" == "x86_64" set "PACKAGES=%PACKAGES%,mingw64-x86_64-glib2.0,mingw64-x86_64-gtk2.0,mingw64-x86_64-pango1.0"

@rem # install Cygwin + packages
@rem # * `setup-%CYGWIN_ARCH%.exe` can be downloaded from https://cygwin.com/install.html
setup-%CYGWIN_ARCH%.exe --no-shortcuts --quiet-mode --root "%CYGWIN_ROOT%" --site "%CYGWIN_URL%" --packages %PACKAGES%

@rem # start Cygwin bash shell
"%CYGWIN_ROOT%"\Cygwin.bat
```

- install/configure OCaml, download unison source, and build unison

```shell (Cygwin/bash)
# Cygwin/bash shell

# configuration/customization (user may alter at their discretion)
BUILDER_ARCH=64 ## [32,64] ; note: must match 'bitness' of CYGWIN_ARCH
DOWNLOADS_DIR=/tmp/ocaml.x${BUILDER_ARCH}-downloads
OCAML_VERSION=4.08.1
PROJECTS_DIR=~/Projects
TARGET_ARCH=32 ## [32,64]
TARGET_BASEDIR=${PROJECTS_DIR}
TOOLSET=mingw ## [mingw,msvc] ; note: 'MSVC' is not working because of GTK+/lablgtk installation blockers
UNISON_REPO_URL=https://github.com/bcpierce00/unison.git
UNISON_REPO_COMMIT=master
UNISON_REPO_DIR=${PROJECTS_DIR}/unison

UNISON_FETCH=https://github.com/rivy/unison fix.builds+publishing ## remove when merged into master

# * OCAML installation/setup

SWITCH=${OCAML_VERSION}+${TOOLSET}${TARGET_ARCH}

BUILDER_ARCH_ID="i686"
case "${BUILDER_ARCH}" in "64") BUILDER_ARCH_ID="x86_64" ;; esac ;
TARGET_ARCH_ID="i686"
case "${TARGET_ARCH}" in "64") TARGET_ARCH_ID="x86_64" ;; esac ;

mkdir -p "${DOWNLOADS_DIR}"
pushd "${DOWNLOADS_DIR}"
curl -#L https://github.com/fdopen/opam-repository-mingw/releases/download/0.0.0.2/opam${BUILDER_ARCH}.tar.xz -o "${DOWNLOADS_DIR}/opam${BUILDER_ARCH}.tar.xz"
tar xvf "opam${BUILDER_ARCH}.tar.xz"
bash opam${BUILDER_ARCH}/install.sh
popd

OPAM_REPOSITORY_URL=https://github.com/fdopen/opam-repository-mingw.git#opam2
opam init --disable-sandboxing --enable-completion --enable-shell-hook --auto-setup default "${OPAM_REPOSITORY_URL}"
opam update

opam switch create ${SWITCH} ${SWITCH}c ## using '...c' pre-compiled version decreases switch installation time by approximately 50%
opam switch ${SWITCH}
eval "$(opam env)"
case "${TOOLSET}" in "mingw")
  ocaml-env exec -- opam install --yes depext-cygwinports depext
  ocaml-env exec -- opam install --yes lablgtk
  ;; esac
case "${TOOLSET}" in "msvc")
  ocaml-env exec -- opam install --yes depext
  ##ocaml-env exec -- opam install --yes lablgtk ## doesn't work
  ;; esac

# * install/build `unison`

mkdir -p "${PROJECTS_DIR}"
mkdir -p "${UNISON_REPO_DIR}"
mkdir -p "${TARGET_BASEDIR}"

## * unison source
git clone "${UNISON_REPO_URL}" "${UNISON_REPO_DIR}"
pushd "${UNISON_REPO_DIR}"
git checkout ${UNISON_REPO_COMMIT} -f ## optional
if [ -n "${UNISON_FETCH}" ] ; then git fetch ${UNISON_FETCH} && git checkout FETCH_HEAD ; fi

## * build unison UISTYLE=text
VERSION_TEXT=$(git describe)
TARGET_DIR="${TARGET_BASEDIR}/unison-${VERSION_TEXT}-text-cygwin.x${BUILDER_ARCH}+ocaml-${SWITCH}"
git clean -fdx
make src OSTYPE=$OSTYPE UISTYLE=text
strip src/*.exe
mkdir -p "${TARGET_DIR}"
cp src/*.exe "${TARGET_DIR}"/

## * build unison UISTYLE=gtk2
VERSION_TEXT=$(git describe)
TARGET_DIR="${TARGET_BASEDIR}/unison-${VERSION_TEXT}-gtk2-cygwin.x${BUILDER_ARCH}+ocaml-${SWITCH}"
git clean -fdx ## needed? => test that generated executable is the same even if not cleaned after the UISTYLE=text build
make src OSTYPE=$OSTYPE UISTYLE=gtk2
strip src/*.exe
mkdir -p "${TARGET_DIR}"/bin
cp src/*.exe "${TARGET_DIR}"/bin
# * package artifacts with required GTK+ files; ref: <https://gtk-rs.org/docs-src/tutorial/cross> @@ <https://archive.is/wip/tIulF>
# [gtk-rs ~ bundling] ref: https://gtk-rs.org/docs-src/tutorial/cross
# ... + gdk-pixbuf; ref: https://github.com/gtk-rs/gtk/issues/422#issuecomment-686927113
# required DLLs
dll_refs() { eval "$(opam env)" ; eval "$(ocaml-env cygwin)" ; objdump -p "$@" | grep -Po "\S+[.]dll$" | xargs -I{} 2>/dev/null which '{}' | sort -u ; }
filtered_dll_refs() { list="$(dll_refs "$@" | grep -vF "$(cygpath ${WINDIR})" | perl -lape '$_ = qq/@{[sort @F]}/')" ; echo -n "$list" ; }
recursive_dll_refs() { list="$(dll_refs "$@")" ; n=0 ; while [ $n -lt $(echo "$list" | wc -l) ]; do n=$(echo "$list" | wc -l) ; list="$(dll_refs $list)" ; done ; echo -n "$list" ; }
recursive_filtered_dll_refs() { list="$(filtered_dll_refs "$@")" ; n=0 ; while [ $n -lt $(echo -n "$list" | wc -l) ]; do n=$(echo -n "$list" | wc -l) ; list="$(filtered_dll_refs $list)" ; done ; echo -n "$list" ; }
IFS=$'\n' DLL_list=( "$(recursive_filtered_dll_refs "${TARGET_DIR}"/bin/*.exe)" )
for dll in ${DLL_list[@]} ; do cp "${dll}" "${TARGET_DIR}"/bin ; done
mkdir -p "${TARGET_DIR}"/lib
# * copy gdk-pixbuf lib/dlls
cp -r /usr/${TARGET_ARCH_ID}-w64-mingw32/sys-root/mingw/lib/gdk-pixbuf-2.0 "${TARGET_DIR}"/lib/
# * update loader.cache to point to local relative installation
mv "${TARGET_DIR}"/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache "${TARGET_DIR}"/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache.original
cat "${TARGET_DIR}"/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache.original | sed -E 's#([^"]*)(lib/gdk-pixbuf-2.0/2.10.0/loaders/[^"]*[.]dll)#../\2#' > "${TARGET_DIR}"/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache

popd
```

glondu: The scripts don't work totally out of the box, but I managed to do something. I've pushed to your branch what I had in mind.

glondu (changes requested):
Please rebase and clean up history, and leave out the GHA workflow commit (and submit in another PR once this one is merged).

smorimoto:
I understand the concern that the already fragile yaml code for CI is becoming more complex and fragile. That's one of the reasons we took this approach to the dune build system's repository. https://github.com/ocaml/dune/blob/master/.github/workflows/ci.ml

src/lwt/win/lwt_unix_impl.ml
rivy: Can you give me feedback on what problems you encountered?

rivy:
I've done as requested and slimmed this down to just the "fix" commits.
It does successfully build on all platforms (see <>).
I just triggered a release build and I'll check on the resultant windows binaries for functionality.
I'll get back to you with some anecdotal results before a merge.

I've got the CICD commit on my fork (building the release now...), and I'll PR it here after this is merged.

glondu:
> I'll get back to you with some anecdotal results before a merge.

Great! Waiting for your feedback, then.

glondu:
> I understand the concern that the already fragile yaml code for CI is becoming more complex and fragile. That's one of the reasons we took this approach to the dune build system's repository. https://github.com/ocaml/dune/blob/master/.github/workflows/ci.ml

I don't understand what you mean: what is "this approach" and how is it better?

rivy:
@glondu ,

> > I'll get back to you with some anecdotal results before a merge.
> 
> Great! Waiting for your feedback, then.

Everything looks good on all of my local tests.

- windows local
- windows <-> linux
- linux remote

I think this is good to merge.
When you do merge it, I'll immediately post my automation PR rebased onto the new master.

rivy:
Rebased to current 'master'.

glondu (approved):

rivy:
@glondu , I've posted the PR for the GHA upgrades (#388).

PR #385, opened 2020-10-06
Remove explicit reference to Stdlib and Pervasives

tleedjarv:
Referencing Stdlib or Pervasives explicitly may cause an unintentional dependency on OCaml version. Name `Stdlib` was introduced in OCaml 4.07 and name `Pervasives` has been deprecated. Without any explicit references to these names, the code remains compatible with all OCaml versions.


hmeine:
Good one! ✅ Re-enables build on Ubuntu 18.04 LTS with the official ocaml 4.05.0 package.

Issue #42, opened 2016-10-15
gtk related stack overflow in array.ml, line 111, when syncing lots of files


ian-kelling:
I was doing a sync which had lots of files with permission changes.

```
[pred] preferpartial 'q/p/c/treetowl.default/webappsstore.sqlite-shm' = false
[pred] forcepartial 'q/p/c/treetowl.default/webappsstore.sqlite-wal' = false
[pred] preferpartial 'q/p/c/treetowl.default/webappsstore.sqlite-wal' = false
<>!<Growing heap to 654360k bytes
Growing heap to 655352k bytes
>!<Growing heap to 656344k bytes
Growing heap to 657336k bytes
Growing heap to 658328k bytes
>!<Growing heap to 659320k bytes
Growing heap to 660312k bytes
>!<Growing heap to 661304k bytes
Growing heap to 662296k bytes
>!<Growing heap to 663288k bytes
Growing heap to 664280k bytes
>!<Growing heap to 665272k bytes
Growing heap to 666264k bytes
>!<Growing heap to 667256k bytes
Growing heap to 668248k bytes
>Growing gray_vals to 2048k bytes
!<Growing heap to 669240k bytes
Growing heap to 670232k bytes
>!<Growing heap to 671224k bytes
>!<Growing heap to 672216k bytes
Growing heap to 673208k bytes
>!<Growing heap to 674200k bytes
Growing heap to 675192k bytes
Growing heap to 676184k bytes
>!Growing heap to 680192k bytes
<>!Uncaught exception Stack overflow
Raised by primitive operation at file "array.ml", line 111, characters 14-41
Called from file "/tmp/buildd/unison-2.48.3/uigtk2.ml", line 3401, characters 12-55
Called from file "/tmp/buildd/unison-2.48.3/uigtk2.ml", line 156, characters 18-22
Called from file "/tmp/buildd/unison-2.48.3/uigtk2.ml", line 3919, characters 4-37
Called from file "/tmp/buildd/unison-2.48.3/uigtk2.ml", line 4276, characters 4-16
```

Using unison-gtk 2.48.3-1 in debian.

ian-kelling:
Btw, I worked around this by using -b for batch and console version, since I knew there would be no conflicts for this run.

bcpierce00:
Not sure what we could do to fix this one, besides giving a more helpful error message and suggesting that people do exactly what you did.  :-)

kaefert:
[...]

PR #386, opened 2020-10-08
uigtk2: Use a tail-recursive function to prevent stack overflows

### LINKED ISSUE #207: stack overflow in GUI if number of changes is huge
From Ubuntu Linux 18.04 I get the following error while running unison-gtk 2.48.4-1ubuntu1, taken from the Ubuntu repository:

Fatal error

Uncaught exception Stack overflow
Raised by primitive operation at file "array.ml", line 139, characters 14-41
Called from file "/build/unison-Y9EcQW/unison-2.48.4/uigtk2.ml", line 3401, characters 12-55
Called from file "/build/unison-Y9EcQW/unison-2.48.4/uigtk2.ml", line 156, characters 18-22
Called from file "/build/unison-Y9EcQW/unison-2.48.4/uigtk2.ml", line 3919, characters 4-37
Called from file "/build/unison-Y9EcQW/unison-2.48.4/uigtk2.ml", line 4276, characters 4-16

tleedjarv:
`List.map` is not tail-recursive and using it will overflow stack in case of large number of reconciliation differences. Use tail-recursive `Safelist.map` instead to prevent stack overflows.

Closes #207 and #42

### DIFF
--- a/src/uigtk2.ml
+++ b/src/uigtk2.ml
@@ -1034,7 +1034,7 @@ module GtkReact = struct
 
   let current_tree_view_selection (t : #GTree.view) =
     let m =t#model in
-    List.map (fun p -> m#get_row_reference p) t#selection#get_selected_rows
+    Safelist.map (fun p -> m#get_row_reference p) t#selection#get_selected_rows
 
   let tree_view_selection_changed t =
     let (res, trigger) = React.make_event () in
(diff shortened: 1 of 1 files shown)


gdt:
Is safelist available all the way back to 4.06 or whatever we aspire to have as the oldest ocaml version? If so, then I think this is ready to merge.

However, a position that we should stop applying changes and only make the minimal fixes to allow a 2.51.3, and then work through the rest of the backlog before making a format change is also entirely reasonable.

tleedjarv:
Yes, Safelist implementation is included in the Unison codebase and the entire code uses (preferably) Safelist rather than List.

bcpierce00:
This looks like a clear (and benign) improvement.

PR #388, opened 2020-10-11
Add GHA build improvements and automatic packaging/publishing

rivy:
- add Windows support
- add much wider compiler variant support
- add GTK+2 build support

* moved from #380

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -1,5 +1,10 @@
 name: Main workflow
 
+env:
+  PROJECT_NAME: unison
+  PROJECT_DESC: "`unison` file synchronizer"
+  PROJECT_EXES: "unison unison-fsmonitor"
+
 on:
   - pull_request
   - push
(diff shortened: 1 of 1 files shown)


tleedjarv:
Please also add OCaml versions 4.11.1 and 4.05.0

I suggest changing the workflow name to something more descriptive. Perhaps just "Build"?

Looking at how complex the workflow definition has become, would it make sense to separate Windows build out into a separate workflow?

rivy:
> Please also add OCaml versions 4.11.1 and 4.05.0

If they build, it's very easily done.

> I suggest changing the workflow name to something more descriptive. Perhaps just "Build"?

I use "CICD"  for workflows with similar characteristics.

> Looking at how complex the workflow definition has become, would it make sense to separate Windows build out into a separate workflow?

I don't think so. You'd only save a dozen or so lines in the current workflow. Much of the added complexity is building the graphic (GTK2) UI binaries and the build deployment logic. If you do separate the Windows portion, you'll be duplicating that and essentially doubling the workflow code size without any real positives. And, given that Windows support has lagged for some time, I think that separating the Windows build story would tend to allow that same drift into second class status again.
Whereas, here, all platform builds share center stage.

gdt:
Would you be willing to prepare a new PR adding them, and we can see how that goes?

rivy:
@gdt, sure, I'll do it now.

tleedjarv:
> I use "CICD" for workflows with similar characteristics.

Works for me! Current name "Main workflow" is really meaningless and non-descriptive, so anything really is better.

> > Looking at how complex the workflow definition has become, would it make sense to separate Windows build out into a separate workflow?
> 
> I don't think so. You'd only save a dozen or so lines in the current workflow. Much of the added complexity is building the graphic (GTK2) UI binaries and the build deployment (packaging and publishing) logic. If you do separate the Windows portion, you'll be duplicating that and essentially doubling the workflow code size without any real positives. And, given that Windows support has lagged for some time, I think that separating the Windows build story would tend to allow that same drift into second class status again. Whereas, here, all platform builds share center stage.

Ok, then clearly no point. Thanks for clarifying.

rivy:
Changing the nomenclature to "CICD" is also easy. I can PR it if desired. @gdt, ?

tleedjarv:
> Changing the nomenclature to "CICD" is also easy. I can PR it if desired. @gdt, ?

That would be great! Once there is new PR, I'd use the comments there to gather agreement from Benjamin and others.
I can also create a PR, if needed, to take it off your shoulders.

gdt:
Sure, sounds good.  Even though this doesn't affect users, I don't want to merge anything until release, but don't let that stop you from preparing it.

rivy:
Rename done (PR #398).

PR #390, opened 2020-10-12
Report time taken propagating changes

thingsconnected:
Although logging is already quite verbose, the time taken to propagate changes is not logged. In our case it is typically in the subsecond range, which is why the duration is printed in ms.

In cases where unison is running as a daemon, monitoring filesystems and keeping them in sync, this is a valuable metric.

### DIFF
--- a/src/transport.ml
+++ b/src/transport.ml
@@ -175,9 +175,12 @@ let transportItem reconItem id showMergeFn =
 
 (* ---------------------------------------------------------------------- *)
 
+let lastLogStart = ref 0.
+
 let logStart () =
   Abort.reset ();
   let t = Unix.gettimeofday () in
+  lastLogStart := t;
   let tm = Util.localtime t in
   let m =
     Printf.sprintf
@@ -195,11 +198,12 @@ let logFinish () =
   let tm = Util.localtime t in
   let m =
     Printf.sprintf
-      "%s finished propagating changes at %02d:%02d:%02d.%02d on %02d %s %04d\n%s"
+      "%s finished propagating changes at %02d:%02d:%02d.%02d on %02d %s %04d, %.3f s\n%s"
       (String.capitalize_ascii Uutil.myNameAndVersion)
       tm.Unix.tm_hour tm.Unix.tm_min tm.Unix.tm_sec
       (min 99 (truncate (mod_float t 1. *. 100.)))
       tm.Unix.tm_mday (Util.monthname tm.Unix.tm_mon)
       (tm.Unix.tm_year+1900)
+      ( t -. !lastLogStart )
       (if Prefs.read Trace.terse || Prefs.read Globals.batch then "" else "\n\n") in
   Trace.logverbose m


tleedjarv:
There actually is a timers mechanism already in the code, so it _may_ be better to use that one. But I think this PR is still relevant because the timers mechanism seems to be more for developer/debugging purposes and in text UI it is not activated for the syncing step.

tleedjarv:
Would it make sense to do some pretty-printing of the value? For subsecond range, as you mention, there is no issue here. For my use cases, synchronization usually lasts at least a couple of minutes but could be as long as several hours. In that case the millisecond display makes no sense (even more so since it's printed without thousands separators).

It sounds a bit like over-doing it but I want to put it out here for discussion. Perhaps something simple like this:
``` ocaml
(let sec = (t -. !lastLogStart) in
if sec <= 1. then
  Printf.sprintf "%.1f ms" (sec *. 1000.)
else if sec <= 60. then
  Printf.sprintf "%.3f s" sec
else
  Printf.sprintf "%.0f min %.0f s" (sec /. 60) (mod_float sec 60))
(* could even continue to print hours *)
```
and then the main format parameter would become just `(%s)` to print the value of this expression (string). This expression could directly replace the current expression in the PR, or - slightly better - placed just before this line https://github.com/bcpierce00/unison/pull/390/files#diff-98a72bb1fefdcf003c30294d26411ee9a1ee5652b2ffa1d3808b80069d94b536R199
Thoughts?

gdt:
I lean to simple, but I see your point.  What's wrong with "%.3f s" with an arg in seconds, always?   People who care can do math,  and this is how time(1) works in unix.

tleedjarv:
Like this then:
``` ocaml
let sec = (t -. !lastLogStart) in
Printf.sprintf "%.0fm%.3fs" (sec /. 60) (mod_float sec 60))
```
(just the concept - actual change can be more like the original PR and just update the format and expression in place, and the format could be `%.0f:%.3f` or whatever)

This would work.

thingsconnected:
For checking longer run times, it is feasible to just manually inspect the timestamp before and after (difference is bigger and the number of runs is smaller).
My PR was really with high volume and machine parsable log files in mind, in order to integrate it in a broader set of system performance metrics, so %.3f in seconds would fit that perfectly. Any kind of dynamic format would almost kill that use case.

tleedjarv:
> For checking longer run times, it is feasible to just manually inspect ...
> My PR was really with high volume and machine parsable log files in mind

Fair point and it makes more sense. Go by that and make the format such that it lends itself for easy parsing.

I'm good with this PR.

gdt:
@thingsconnected I wasn't going to ask you to rebase these together, but I merged other things and came back to merge this after the CI checks finishing, but now it says there is a conflict.  Could you rebase to current master and then I'll hit merge (after CI run).

thingsconnected:
Done!

PR #391, opened 2020-10-13
Add additional OCaml versions to GHA CICD

rivy:
- additional CICD OCaml versions/variants to be build by CICD workflow (by request; see <https://github.com/bcpierce00/unison/pull/388#issuecomment-707397387>)

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -15,20 +15,26 @@ jobs:
       fail-fast: false
       matrix:
         job:
+        - { os: macos-latest   , ocaml-version: 4.11.1 }
         - { os: macos-latest   , ocaml-version: 4.10.1 }
         - { os: macos-latest   , ocaml-version: 4.09.1 }
         - { os: macos-latest   , ocaml-version: 4.08.1 }
         - { os: macos-latest   , ocaml-version: 4.07.1 }
+        - { os: macos-latest   , ocaml-version: 4.05.0 }
+        - { os: ubuntu-latest  , ocaml-version: 4.11.1 }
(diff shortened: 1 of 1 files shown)


.github/workflows/workflow.yml
```diff
@@ -15,20 +15,26 @@ jobs:
...
         - { os: macos-latest   , ocaml-version: 4.08.1 }
         - { os: macos-latest   , ocaml-version: 4.07.1 }
+        - { os: macos-latest   , ocaml-version: 4.05.0 }
+        - { os: ubuntu-latest  , ocaml-version: 4.11.1 }
         - { os: ubuntu-latest  , ocaml-version: 4.10.1 }
         - { os: ubuntu-latest  , ocaml-version: 4.10.0+musl+static+flambda }
         - { os: ubuntu-latest  , ocaml-version: 4.09.1 }
         - { os: ubuntu-latest  , ocaml-version: 4.08.1 }
         - { os: ubuntu-latest  , ocaml-version: 4.07.1 }
-        - { os: windows-latest , ocaml-version: 4.10.1+mingw64 }
```
gdt, line 27: Why is this deleted?

rivy: The Windows OCaml installation defaults to the MinGW 64-bit version, so, "4.10.1+mingw64" is the same as just "4.10.1".

rivy:
The failures look like a transient network error when downloading the ubuntu-18 musl-tools fix (see <https://github.com/ocaml/ocaml/issues/9131#issuecomment-599765888>) during the OCaml setup. If it recurs with any frequency, `ubuntu-20.04` can be used instead of `ubuntu-latest`.

gdt:
OK - we can think about transient errors and versions later.

PR #392, opened 2020-10-13
Update NEWS (content, move to where it is expected)

gdt:
This PR:
  * moves src/NEWS to top-level, which is where NEWS is generally expected to be
  * converts it to markdown (lightly)
  * adds a brief overview
  * adjusts wording about which NEWS item is about which release
  * adds item for the 2.51.3 entry

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,8 @@
+# unison NEWS
 
 Changes in Version 2.51.3
 
-   Changes since 2.51.2:
+## Changes since 2.51.2:
      * Some nontrivial changes to profile parsing (G.raud Meyer)
           + ’=’ has been considered whitespace until now: several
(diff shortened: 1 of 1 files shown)


glondu:
Note that src/NEWS is generated by `make export` (which also updates doc strings built in the executable). I find this Makefile target... embarrassing... but I didn't take the time to propose something better for the 2.51.3 release.

gdt:
Well in that case we shouldn't merge this.  I'll leave the PR open for post 2.51.3 because I think we should detangle the source code from the giant latex document.  (I am a latex fan, but I see "user manual" and "following source code conventions" as separate.)

gdt:
See #393

gdt:
This needs to be fixed, but having an open PR is not helping anyone.

PR #398, opened 2020-10-15
Refactor ~ rename GHA CICD workflow

rivy:


gdt:
Looks good to me.  Will hit merge after release.

tleedjarv:
I think README.md has the workflow name hardcoded in it (?). If so, you'd have to change that as well. Otherwise looks good to me.

rivy:
I found it... README change added/squashed into the single PR commit.

glondu (approved):

PR #400, opened 2020-10-16
Put -custom in CAMLLIBS instead of CAMLFLAGS

glondu:
This affects only bytecode compilation.

The -custom option should not be used combined with -c. Moreover, in
Debian, -custom is an alias to -output-complete-exe (since OCaml
4.11.1), which cannot be used combined with -c. Hence, without this
patch, unison fails to build on bytecode-only architectures.

### DIFF
--- a/src/Makefile.OCaml
+++ b/src/Makefile.OCaml
@@ -375,6 +375,7 @@ ifeq ($(NATIVE), true)
     CAMLFLAGS+=-p
     CLIBS+=-cclib -ldl
   endif
+  CAMLLDFLAGS=
 
   CAMLOBJS=$(subst .cmo,.cmx, $(subst .cma,.cmxa, $(OCAMLOBJS)))
   CAMLLIBS=$(subst .cma,.cmxa, $(OCAMLLIBS))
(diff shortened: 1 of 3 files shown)


glondu:
I think this should be part of the 2.51.3 release.

gdt:
Could other people that understand ocaml comment?   Coming from C where there is LDFLAGS and LIBS, this option really feels like LDFLAGS, so putting it in LIBS seems off.  However it may be that we need three variables to avoid this perception, and that may not make sense.

If this belongs in 2.51.3, I'm conceptually ok with that but would want to cycle an rc.

gdt:
Reading Makefile.OCaml, my understanding is:
 * This flag is not "a lib", but belongs on the link link for -blob.o only, and does not belong when compiling an individual file.
 * OCAMLFLAGS is like CFLAGS and used for individual compiles
 * arguably we should have OCAMLLDFLAGS also on the blob link command, and put -custom there instead of LIBS

However, I'm find with skipping that for now and maybe/probably adjusting later, on the theory that the smaller the change now for an rc iteration the better.

tleedjarv:
Intent of the change looks good to me, but agree with @gdt, it really should be LDFLAGS-type variable.

glondu:
I agree with what you said: this option belongs to an LDFLAGS-type variable, but I've not found one. I've just put it in CAMLLIBS because I was out of time. I will propose something better.

tleedjarv:
Looks good to me.
Should recipe for $(NAME)-blob.o target also be changed? I think it is only used by GUI on macOS, so that's not bytecode. But CAMLLDFLAGS variable could still be added.

glondu:
> Should recipe for $(NAME)-blob.o target also be changed?

I don't think so. It uses -output-obj, which seems incompatible with -custom/-output-complete-exe. But I don't have easy access to a macOS system to check...

PR #401, opened 2020-10-16
Fix typos

glondu:


glondu:
This is documentation changes that can be part of 2.51.3 if we decide to merge #400 (and make a third release candidate).

gdt:
This looks fine, so if we merge the other and do rc3 (which feels like how we are headed, which is fine) happy to merge this too.

When I was confused about NEWS I added some entries, in my not-going-to-be-merged PR https://github.com/bcpierce00/unison/pull/392   Since you are set up for handling latex regen, if you are inclined, sticking those in the manual and regenerating seems like it would be nice.   I will get to that eventually, just not before next rc as I want to be measured to make sure I don't do the regen wrong.

gdt (approved):
Approved from content viewpoint; still on release management hold.

PR #404, opened 2020-10-18
Replace deprecated String functions

tleedjarv:
### Description

Re-work of the patch replacing deprecated String functions that got reversed by #376 due to introducing incompatibility.

This PR is based on ideas from https://github.com/bcpierce00/unison/pull/376#pullrequestreview-501558712

I believe the usage of `uppercase` in transport.ml has been unintentional, so I've replaced those with `capitalize` (as has been in GUI and documentation, for example). I will revert if that is not correct.

My testing shows that compatibility with 2.51 is retained.

### Background

OCaml standard `String.{lowercase,uppercase,capitalize}` etc. functions operating on strings assume Latin1 (ISO 8859-1) encoding. These functions are now deprecated in favor of functions operating on only ASCII characters, not assuming any extended encoding.

Most invocations of such functions in Unison can be replaced by their ASCII counterparts with no side-effects.

As Unison supports either Unicode or Latin1 for actual file name synchronization then add a Latin1 lowercase function, to replace the OCaml standard library function.


glondu (approved):

gdt:
I have a few questions:

Will this work on all versions of ocaml currently supported?  It looks like it, from the CI output.

Does unison have any internationalization?  Is this only about user filenames?  I know latin1 accomodates most European languages, but it seems people are moving to unicode and UTF8.   I am in the land of ASCII and I don't know what people do in terms of file names with either latin1 or UTF8 characters.   Does this change have any impact on those practices?

I see capitalize_ascii, but other helpers using latin1.  Is that because what is being capitalized is a non-localized text string that is known to be ascii only?

tleedjarv:
Yes, it works with all currently supported OCaml versions (down to 4.03, I think).

No internationalization currently, it is only about user filenames. Historically, Unison has supported only Latin1 and Unicode with Unicode being default. This change has no functional impact (well, that's the intention at least).

*_ascii are OCaml stdlib functions (introduced in 4.03 to replace the deprecated functions), *_latin1 is introduced by this change specifically for Unison because using *_ascii would break case insensitive Latin1 filename support.

gdt (approved):
Thanks for the explanation.  Will merge after release.

PR #405, opened 2020-10-20
Fix const-related warnings

glondu:
This is not technically required for the 2.51.3 release, but fixes something that may be broken with future versions of OCaml (or C) compiler, so it should be merged eventually.

### DIFF
--- a/src/fsmonitor/linux/inotify_stubs.c
+++ b/src/fsmonitor/linux/inotify_stubs.c
@@ -58,7 +58,7 @@ static int inotify_return_table[] = {
 
 static void raise_inotify_error(char const *msg)
 {
-        static value *inotify_err = NULL;
+        static const value *inotify_err = NULL;
         value args[2];
 
         if (!inotify_err)
--- a/src/bytearray.ml
+++ b/src/bytearray.ml
@@ -39,10 +39,10 @@ external unsafe_blit_from_string : string -> int -> t -> int -> int -> unit
   = "ml_blit_string_to_bigarray" [@@noalloc]
 
 external unsafe_blit_from_bytes : bytes -> int -> t -> int -> int -> unit
-  = "ml_blit_string_to_bigarray" [@@noalloc]
+  = "ml_blit_bytes_to_bigarray" [@@noalloc]
 
 external unsafe_blit_to_bytes : t -> int -> bytes -> int -> int -> unit
-  = "ml_blit_bigarray_to_string" [@@noalloc]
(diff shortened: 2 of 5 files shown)


src/fsmonitor/linux/inotify_stubs.c
```diff
@@ -58,7 +58,7 @@ static int inotify_return_table[] = {
 
 static void raise_inotify_error(char const *msg)
 {
-        static value *inotify_err = NULL;
```
gdt, line 61: It's not obvious from the diff why declaring a variable not const is a programming error.

gdt:
Obviously for later, but I think a PR that changes language requirements needs to 1) say that in the PR and 2) provide rationale for what it is changed from and to, explain the impact, and why that's ok.   Not saying it isn't, just that it's hard to know.   I also don't follow 4.06.1 vs .0.   Plus I am unclear on where we document the language versions that are supported,  but that needs updating and probably the CI list.

tleedjarv:
@glondu I think you can avoid bumping OCaml version by just defining `Bytes_val` macro for OCaml < 4.06
I.e.
``` c
#ifndef Bytes_val
#define Bytes_val(x) ((unsigned char *) Bp_val(x))
#endif
```

src/fsmonitor/linux/inotify_stubs.c
glondu: In this case, the variable is later assigned a "const value *" (returning from caml_named_value). By not declaring const in this declaration, the const attribute is forgotten, which can cause bad surprises later.

glondu:
> 1) say that in the PR

There is (was) an explicit commit saying that in the PR.

> 2) provide rationale for what it is changed from and to, explain the impact, and why that's ok

The commit dropping support for OCaml 4.05.x was explicit about that, wasn't it?

> I also don't follow 4.06.1 vs .0.

4.06.1 was the last release in the 4.06.x series, and 4.06.0 was the first one in that series.

glondu:
> I think you can avoid bumping OCaml version by just defining Bytes_val macro for OCaml < 4.06

Thank you for this, I've updated the branch.

src/fsmonitor/linux/inotify_stubs.c
gdt: Thanks for the explanation.  I didn't follow that - I should get in the habit of looking at more context than github shows me.

gdt (approved):
Will merge post-release.

PR #410, opened 2020-10-22
Assume documentation strings are in UTF-8

glondu:
This removes calls to transcodeDoc on documentation strings. The
transcodeDoc function transcodes from Windows Codepage 1252 to
Unicode, and assumes that the strings in strings.ml are in the former
encoding. Actually, they are in the encoding of the locale used to
generate strings.ml, which should be UTF-8 on most systems these days
(and was during the preparation of 2.51.3). Therefore, the call to
transcodeDoc is no longer needed.

Without this, the documentation (windows spawned by the "Help" menu)
contains corrupted characters.

### DIFF
--- a/src/uigtk2.ml
+++ b/src/uigtk2.ml
@@ -2522,7 +2522,6 @@ let documentation sect =
   ignore (t#event#connect#delete ~callback:(fun _ -> dismiss (); true));
 
   let (name, docstr) = Safelist.assoc sect Strings.docs in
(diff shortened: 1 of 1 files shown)


gdt (approved):
Looks fine to me; on the list to merge soon.

(I think there are larger issues lurking in the docs process; the processing should not be affected by choice of OS or current locale, but this PR shouldn't be help up by that.)

PR #417, opened 2020-10-24
WIP: GHA: Enable manual build and user-selected make target

tleedjarv:
This PR is for discussion of the concept and for figuring out the details.

This will make it possible to manually trigger CICD build and to optionally enter a freely chosen make target to run after regular build and test targets.

This can be used to manually run ad hoc test targets or the docs target, for example.

Additionally, I was hoping to enable user entry of ad hoc OS and OCaml version (you can see this in the `inputs`) but I could not figure out how to change the job matrix to follow the user input (it is simple in a separate manual build-only workflow). @rivy do you have any suggestions? It may not be that relevant in the end but it seems unnecessary to run the manual make target for all OS-OCaml combinations.

I did not add the line
``` yml
shell: bash --login --norc --noprofile -e -o pipefail -c "dos2unix '{0}' 2>/dev/null ; source '{0}' ;"
```
I don't know if it is needed.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -6,8 +6,20 @@ env:
   PROJECT_EXES: "unison unison-fsmonitor"
 
 on:
-  - pull_request
-  - push
+  pull_request:
+  push:
+  workflow_dispatch:
+    inputs:
+      os:
+        description: 'OS (e.g. ubuntu-18.04, ubuntu-latest, windows-latest, macos-10.15)'
+        required: false
+      ocaml:
+        description: 'OCaml version (e.g. 4.03.0, 4.11.1)'
+        required: false
+      target:
+        description: 'Make target (e.g. -C src test)'
+        required: false
+        default: 'test'
 
 jobs:
   build:
@@ -163,6 +175,9 @@ jobs:
     - if: runner.os != 'Windows' ## [2020-09] Windows builds fail self-testing
       shell: bash --login --norc --noprofile -e -o pipefail -c "dos2unix '{0}' 2>/dev/null ; source '{0}' ;"
       run: opam exec -- make OSTYPE=$OSTYPE STATIC=${{ steps.vars.outputs.STATIC }} test
+
+    - if: github.event_name == 'workflow_dispatch'
+      run: opam exec -- make ${{ github.event.inputs.target }}
 
     - uses: actions/upload-artifact@v2
       with:


rivy:
> ```yaml
> shell: bash --login --norc --noprofile -e -o pipefail -c "dos2unix '{0}' 2>/dev/null ; source '{0}' ;"
> ```
> 
> I don't know if it is needed.

It's only strictly necessary for using `bash` scripting under Cygwin (ie, Windows) builds, after OCaml is installed. The default shell for Windows builds is now `PowerShell` and your included single line works under `PowerShell` or `bash`, so it's not needed here... But, not knowing/understanding that Windows is using a different shell here can come back to bite you if you change the script. And, additionally, the Windows default shell has been changed since GHA inception from `cmd` to `PowerShell`. So, for clarity, simplicity, and robustness, I recommend specifying the shell and therefore using this as the shell specification.

I haven't used manual triggering of GHA workflows, so I'm probably not the best person to ask. But, if I built this, even though it would involve some code/script duplication, I'd build it as a separate workflow. It's seems logically separate to me and threading the manual build logic within the current CICD code for this would seem to be distracting within an admittedly already complex process.

I'm happy to help build that alternate workflow if that's ultimately desired.

tleedjarv:
I also think separate workflow would be better. But I don't want to even think of duplicating all of this code/script. Perhaps a local GHA action could be used to share code/script but I don't want to introduce that complexity either without knowing all the options.

If the `os` and `ocaml` inputs are removed from this PR then I think it still provides minimal benefits as is. But to take it further, add the possibility to retrieve artifacts, for example, then a separate workflow starts looking almost unavoidable. All ideas and suggestions are welcome.

gdt:
I would really like to see the manual build and CI build be separate actions, or maybe I don't understand the GHA way.  I see your duplication point though.

tleedjarv:
I looked into using GHA "actions" to share common steps between several workflows (namely something called "composite run steps action"). It does show some promise but steps in actions are much more limited compared to steps in workflows. The most severe limitation currently is the inability to invoke other actions (e.g. to set up ocaml toolchain, etc.). I haven't looked into using external scripts and whether the workflow steps could be replaced by one or several scripts.

All limitations can probably be overcome in some way but I don't know GHA nearly enough to know what that way would be. Most likely that would involve workarounds, making the already complex workflow even more complex.

Opinions from those with more experience with GHA are welcome.
@smorimoto ?

gdt:
I guess having a separate action is a nicety, and if we can build the manual on the platforms we want (not that many!), and nothing else breaks, that seems ok.   I can be refactored later if someone figures out how to do that nicely.

smorimoto:
Composite action is probably useful for that purpose: https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/creating-a-composite-run-steps-action

tleedjarv:
> Composite action is probably useful for that purpose

No good. This is the action type I was referring to. It lacks support for invoking other actions and if-conditions.

tleedjarv:
Closing this as there currently seems no good way to complete the idea and there is no availability to keep maintaining the workflow.

PR #419, opened 2020-11-01
Fix incompatible type on 64-bit platforms

### LINKED ISSUE #166: failure syncing from windows: input_value_from_block: bad object
Hi,

I'm experimenting with syncing files between a windows host and linux host. Went already through OCaml/Unison building process and have two binaries with the same version (2.51.2) built with the same OCaml compiler (4.04.0). Unison on windows built with Mingw64 from MSYS2.

Unfortunately, I'm getting this error while trying to sync files:

unison -auto -batch -perms=0 -sshargs "-i secret.pem -o StrictHostKeyChecking=no" ./ ssh://test@test.host//home/test/
Unison 2.51.2 (ocaml 4.04.0): Contacting server...
Unison server failed: Fatal error: Fatal error during unmarshaling (input_value_from_block: bad object),
possibly because client and server have been compiled with differentversions of the OCaml compiler.
Fatal error: Lost connection with the server

the unison version on the remote host:
unison version 2.51.2 (ocaml 4.04.0)

Any ideas how can I fix this?

thanks

tleedjarv:
Fix a type in C stub code which is not the same for OCaml on some 32 and 64-bit platforms. The problem only manifests itself on Windows platforms with MinGW 64-bit compiler. This caused the marshaling code to produce invalid output.

Closes #166

### DIFF
--- a/src/bytearray_stubs.c
+++ b/src/bytearray_stubs.c
@@ -10,7 +10,7 @@
 CAMLprim value ml_marshal_to_bigarray(value v, value flags)
 {
   char *buf;
-  long len;
+  intnat len;
   output_value_to_malloc(v, flags, &buf, &len);
   return alloc_bigarray(BIGARRAY_UINT8 | BIGARRAY_C_LAYOUT | BIGARRAY_MANAGED,
                         1, buf, &len);


glondu (approved):
Well spotted!

src/bytearray_stubs.c
```diff
@@ -3,14 +3,15 @@
...
 
+#include "caml/mlvalues.h"
 #include "caml/intext.h"
 #include "caml/bigarray.h"
 #include "caml/memory.h"
 
 CAMLprim value ml_marshal_to_bigarray(value v, value flags)
 {
   char *buf;
-  long len;
```
gdt, line 13: I am not really following this.  Certainly it is wrong to use "long" as the size of long is not fixed.    On i386, the normal (unix) convention is ILP32, and on x86_64 LP64.   So with the use of long, I'd expect far more trouble than we are seeing.   Can you add to the commit message an explanation of how this is wrong, and why you believe that after this change we will have no wire protocol break compared to all systems except windows-x86_64-mingw?

intnat is basically a typedef for value, but I can't find that.

In config.h I see a lack of ILP64, like Alpha.  But that's an ocaml bug, not a unison bug, I think.

Or is this about the bits ending up in different places in &len and happy aliasing depending on LE vs BE?

Were there warnings before?

glondu: I don't follow you. Maybe things are clearer knowing that long is 32-bit even on Windows x86_64?

> Were there warnings before?

Yes: look at Win64 CI/CD build logs (for example https://github.com/bcpierce00/unison/runs/1311424481?check_suite_focus=true ). With @tleedjarv 's patch, the warning disappears.

gdt: So Windows x86_64 is IL32 P64?

What I don't understand is that long changes size from i386 to x86_64 on unix, so if this sizeof(long) issue caused trouble on Windows (which I believe from the reports of course), then I don't understand why it has caused no other trouble.

gdt: So is it that intnat is defined to be the natural word size, which is long=32 on i386, long=64 on x86-64, and ends up long long=64 on windows x86-64 mingw?   In other words, long and intnat are both variable, but one tracks what you'd expect and long is strange on Windows x86-64 mingw?

Where is intnat defined?  I grepped the ocaml .h files and didn't find it, but I admit to not reading them all.

glondu: Yes to all.

> Where is intnat defined?

In config.h.

gdt:
Thanks, found it and totally makes sense now (except that ocaml behaving diffferently based on CPU does not make sense to me).    I'm now good with the first commit, but @tleedjarv it would be nice to have this explained because there is a lot of complexity here.  Now you know what I didn't already understand - I get it that it's hard to guess about that when you do really understand.

gdt (approved):
I think this is mergeable now, but I'm also happy to wait for a bit more community testing if you would like that.

PR #420, opened 2020-11-01
Do not set USR1 signal handler on Windows

tleedjarv:
Not all signals are available on Windows. Failure to set a signal handler will raise an exception which is not caught. This signal handler is set at the toplevel code, so an unhandled exception will prevent program startup.

### DIFF
--- a/src/ubase/trace.ml
+++ b/src/ubase/trace.ml
@@ -128,10 +128,15 @@ let closelog _ =
     None -> ()
   | Some(ch,file) ->
       close_out ch;
-      logch := None;
-  ;;
+      logch := None
 
-Sys.signal Sys.sigusr1 (Signal_handle closelog);;
+let _ =
+  if Util.osType <> `Win32 || Util.isCygwin then
+    try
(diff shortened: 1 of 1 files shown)


src/ubase/trace.ml
```diff
@@ -128,10 +128,11 @@ let closelog _ =
     None -> ()
   | Some(ch,file) ->
       close_out ch;
-      logch := None;
-  ;;
+      logch := None
 
-Sys.signal Sys.sigusr1 (Signal_handle closelog);;
+let _ =
+  if Util.osType = `Win32 && not Util.isCygwin then
```
tleedjarv, line 134:
```suggestion
  if Util.osType <> `Win32 || Util.isCygwin then
```
This is the correct condition.

gdt (commented):
So if you adjust as you suggest, this looks like a good fix, and then the revert of this can be dropped from other PR.  Is that what you are thinking?

tleedjarv:
> So if you adjust as you suggest, this looks like a good fix, and then the revert of this can be dropped from other PR. Is that what you are thinking?

Yes, that's it.

PR #423, opened 2020-11-04
CICD: Build Unison.app for macOS

### LINKED ISSUE #422: Add macOS Unison.App to build system and CI build
Hi, apologies if I'm missing something obvious, but the 2.51.2 release for macOS was an application. 2.51.3 appears to be just the binaries. Is there still an intention to release a compiled application? Since I am tasked with deploying to managed Macs, we don't support Homebrew etc. as these tools are not very compatible and often interfere with enterprise management systems.

tleedjarv:
Unison.app builds were left out of the original CICD PRs (see #380 and #388) since they were not working. Unison.app is clearly something that has users (see #422).

Once the changes are approved, this PR will be split in two: one with build fixes and another with CICD workflow changes.

Closes #422

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -79,10 +79,11 @@ jobs:
         PKG_OS='linux' ; case '${{ matrix.job.os }}' in macos-*) PKG_OS='${{ matrix.job.os }}' ;; windows-*) PKG_OS='windows' ;; esac;
         PKG_STATIC='' ; case '${{ matrix.job.ocaml-version }}' in *+static*) PKG_STATIC='.static' ;; esac;
         PKG_ARCH='x86_64' ; case '${{ matrix.job.ocaml-version }}' in *+32bit* | *+mingw32*) PKG_ARCH='i386' ;; esac;
-        PKG_BASENAME="${PROJECT_NAME}-${REF_TAG:-$REF_SHAS}+ocaml-${OCAML_VARIANT/%+*/}+${PKG_ARCH}.${PKG_OS}${PKG_STATIC}"
+        PKG_VER="${REF_TAG:-$REF_SHAS}"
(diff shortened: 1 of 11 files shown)


tleedjarv:
**Update**
uimac builds are now working on all OCaml versions. The build fix could be separated from the CICD PR but there is a bit of a chicken and egg problem here. Separating into two PRs could be done once ready for merging.

tleedjarv:
**Update**
Added fixes to silence compiler warnings. Included in this PR only for testing purposes. Once approved, the GHA workflow changes and build/compiler warning fixes will be separated into two PRs.

tleedjarv:
**Update**
Proper publish packaging is now added in the workflow. The zip file that can be downloaded from workflow artifacts contains the .tar.gz that will be published for a release. Please test the .tar.gz as if it was downloaded from the releases page.

All the expected changes are now complete. I leave "WIP" in the PR title for time being because I can't test the produced `.app` myself and some packaging fixes might still be expected.

gdt:
From the test report, I think this is likely ready.    My only comment on the warnings fixes is that I don't follow the NSInteger/long and %ld changes.  There is obviously some complexity lurking about ocaml types, C types, and printf specifers.   Sometimes there is cast to long, sometimes to int, and I can't convince myself quickly that it's all right.   I don't doubt you; would just appreciate a few words of rationale when you rebase into a "fix code that causes warnings" PR and the add mac build PR.

It would also be nice to have more eyes on the warnings fixes from people that know ocaml better.

tleedjarv:
Actually, it should be three separate PRs: 1) build fix; 2) CICD workflow change (with build fix a pre-requisite); 3) compiler warning fixes.

gdt:
I await your 3 separate PRs :-)

(Not about this PR, but I would like to let the other merges I just did cool off 24h before merging another PR.)

tleedjarv:
Now split in three. #431 must be merged first, then #432

PR #425, opened 2020-11-06
uigtk2: Fix profile editing (lost user input)

### LINKED ISSUE #19: Gtk2 interface: Impossible to change integer/text values in profiles
It is impossible to change integer or text values (e.g. for properties
perms, key) in profiles using the GUI (unison-gtk). Instead, the values
become empty if one tries to change them.
The error seems to be around line 1820 in uigtk2.ml.

### LINKED ISSUE #188: Cannot set option value to 0 via GUI
**Steps to reproduce**

* Start the Unison GUI.
* Set up sync between to local directories.
* Add the `perms` option (or any other numeric setting of your choice)
* In the dialog box, enter 0 (which should be the suggested value).
* Select the profile.

**Behavior observed**

* Unison gives an error because of the option not having a value, and the directory cannot be synced.
* Upon examining the profile settings file manually, there is indeed a line `perms = ` (with no value instead of the expected zero value).

**Behavior expected**

* A zero value in the file, and Unison syncing the two directories.

**Additional information**

If I manually edit the profile settings file, supplying the missing zero value, and restart Unison, everything works as expected.

**System Environment**

Unison 2.48.3 on Ubuntu MATE 16.04 LTS. I have observed the same behavior in earlier Unison versions (around 2.4x) on Windows 7; did not test 2.48.3.

tleedjarv:
Profile preference edit component (text field) is currently destroyed before the value input by user is read out. This PR changes the execution order so that the value is always read before destroying the components.

Closes #19 
Closes #188

### DIFF
--- a/src/uigtk2.ml
+++ b/src/uigtk2.ml
@@ -1809,7 +1809,7 @@ let editPreference parent nm ty vl =
           (fun () -> [if trueB#active then "true" else "false"])
       | `INT | `STRING ->
            let valueEntry =
-             GEdit.entry ~text:(List.hd vl) ~width_chars: 40
+             GEdit.entry ~text:v ~width_chars: 40
                ~activates_default:true
                ~packing:(tbl#attach ~left:1 ~top:3 ~expand:`X) ()
            in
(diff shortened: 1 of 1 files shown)


gdt:
Looks ok to me; would appreciate review by someone who really knows ocaml.

PR #426, opened 2020-11-07
Fix more safe strings-related compiler warnings

tleedjarv:
More C code changes related to OCaml safe strings becoming default.


gdt:
Looks ok to me, but would appreciate review from someone who really understands the ocaml string/char world.

Issue #394, opened 2020-10-13
UI preference is a compile time setting


gdt:
One can build with UISTYLE=text, or UISTYLE=gtk2.  This sets the default way the unison binary behaves, and results in unwanted behavior from scripts wanting to use the text ui (as -auto) if the system-wide build of unison is built for GUI.

Building GUI or not should be separate from how the default invocation works.   Arguably there should be a separate executable "xunison" or some other name that runs GUI mode for those that want it, or some other way to have the unison command run the comand-line UI regardless of build setting.

(This bug probably doesn't make sense to those who build unison only for themselves, vs build it in a packaging system where it will be installed for multiple users of a timesharing system.)

rivy:
I'm not sure this is really a problem as most installations install the text-only UI as `unison` and the GUI UI under a different name such as `unison-gui` or `unison-gtk`. Additionally, the text UI is always available from either type of build by using the `-ui text` command line option. That is, using `unison -ui text ...` will always use the text UI no matter what GUI type is installed.

glondu:
> I'm not sure this is really a problem as most installations install the text-only UI as `unison` and the GUI UI under a different name such as `unison-gui` or `unison-gtk`.

Indeed, this is what is done in Debian, for example.

gdt:
If most distributions build the text only as unison and the GUI under a different name, then it seems like this notion should be hoisted into unison proper, both to make things easier, and to have uniformity of command names for users.   Right now the build system does one thing and some packaging systems are doing something else.

gdt:
And also the comment by @rivy indicates that Linux packagers think that typing unison (vs unison-gtk) should result in a text ui without having to ask for it.  That's really what I was getting at.

rivy:
The just merged GHA build/packaging system is building the text-only UI to `unison` and the text/graphic UI (when buildable) to `unison-gtk2`. Both are packaged into the "bin" directory with `unison-fsmonitor` (again, when buildable). If desired, a less specific name such as `unison-gui` could be used as an alternative.

gdt:
Thanks - I will have a look.  I would like to see whatever controls are used to do that merged to the main build and be default, rather than having special stuff in the GHA builds.   At least what you describe seems not to happen if you just run make.

gdt:
This appears related to the need for https://github.com/bcpierce00/unison/pull/427

gdt:
Thanks for the comments; they were useful to understand but I'm calling this a dup of #667.

PR #427, opened 2020-11-08
Fix self-tests on Windows

tleedjarv:
Fix self-tests on Windows and enable them in CICD workflow.

I've moved the test step before the gtk2 build as the gtk2 executable on Windows doesn't run on GHA environment with current workflow scripts (can't find required dll-s).

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -151,6 +151,9 @@ jobs:
           fi
         done
 
+    - shell: bash --login --norc --noprofile -e -o pipefail -c "dos2unix '{0}' 2>/dev/null ; source '{0}' ;"
+      run: opam exec -- make test
+
     - if: steps.vars.outputs.STATIC != 'true' ## unable to build static gtk for linux or windows/Cygwin MinGW platforms
       shell: bash --login --norc --noprofile -e -o pipefail -c "dos2unix '{0}' 2>/dev/null ; source '{0}' ;"
       run: |
@@ -159,10 +162,6 @@ jobs:
         # * copy only main/first project binary
         project_exe_stem=${PROJECT_EXES%% *}
         cp "src/${project_exe_stem}${{ steps.vars.outputs.EXE_suffix }}" "${{ steps.vars.outputs.PKG_DIR }}/bin/${project_exe_stem}-gtk2${{ steps.vars.outputs.EXE_suffix }}"
-
-    - if: runner.os != 'Windows' ## [2020-09] Windows builds fail self-testing
-      shell: bash --login --norc --noprofile -e -o pipefail -c "dos2unix '{0}' 2>/dev/null ; source '{0}' ;"
-      run: opam exec -- make OSTYPE=$OSTYPE STATIC=${{ steps.vars.outputs.STATIC }} test
 
     - uses: actions/upload-artifact@v2
       with:
--- a/src/uicommon.ml
+++ b/src/uicommon.ml
@@ -506,10 +506,11 @@ let promptForRoots getFirstRoot getSecondRoot =
 (* ---- *)
 
 let makeTempDir pattern =
-  let ic = Unix.open_process_in (Printf.sprintf "(mktemp --tmpdir -d %s.XXXXXX || mktemp -d -t %s) 2>/dev/null" pattern pattern) in
-  let path = input_line ic in
-  ignore (Unix.close_process_in ic);
-  path
+  let path = Filename.temp_file pattern "" in
+  let fspath = System.fspathFromString path in
+  System.unlink fspath; (* Remove file created by [temp_file]... *)
+  System.mkdir fspath 0o755; (* ... and create a dir instead. *)
+  path ^ Filename.dir_sep
 
 (* The first time we load preferences, we also read the command line
    arguments; if we re-load prefs (because the user selected a new profile)


gdt:
I'm a little uncomfortable with moving the tests earlier on non-Windows, because
  * it seems like it could reduce coverage
  * it seems like what should be "make check" should not rely on the main build, and instead depend on things, but somehow what the tests do on windows behaves differently with build and not done
  * I don't understand how we can build a GUI program with dlls that don't exist.

tleedjarv:
> * it seems like it could reduce coverage

That's not the case with current tests. These self-tests are not related to UI code.

Really, to get full test coverage of all UI options then `make test` should be run separately after each build (also, if in future there would be tests covering UI).

> * it seems like what should be "make check" should not rely on the main build, and instead depend on things, but somehow what the tests do on windows behaves differently with build and not done

Currently each build outputs an executable named `unison`. That is, gtk2 build will just overwrite the executable from text build (as currently they are built in that sequence--but there is nothing stopping building them in other sequence). `make test` executes an executable named `unison`. In other words, it tests whichever build is the most recent.

Mac .app is completely separate, of course, and should be tested separately (see first point). But then again, all current self-tests are for the core code and not related to UI.

> * I don't understand how we can build a GUI program with dlls that don't exist.

They exist. I guess it's just a matter of having the correct LD_LIBRARY_PATH (whatever is the equivalent in Windows).

gdt:
So this is related to #394, which is worked around in packaging and in the GHA builds, but not solved enough for tests to work.   I don't mind working around it; I just wanted to understand why as my read was that the need for the workaround was evidence of a bug someplace else.

Agreed with your comments, except about the UI build overwriting the text binary and that sequencing build/test/build/test is the preferred approach.  I think build should build both, and then test should test both, if that's how the overall build is configured.

I don't follow Windows linking well enough, but it seems like not being able to run the linked binary as a test is itself a bug, but one that is not currently causing any pain.

gdt (approved):
I'm now fine with applying this but as usual would appreciate someone else reading it and commenting.

tleedjarv:
> Agreed with your comments, except about the UI build overwriting the text binary and that sequencing build/test/build/test is the preferred approach. I think build should build both, and then test should test both, if that's how the overall build is configured.

Yes, that would either require a fix for #394 or a simpler alternative--add the `make test` recipe directly into GHA workflow. The recipe is just `./src/unison -selftest`, so adding it directly to workflow would allow calling any executable with any name and location with `-selftest`.

> I don't follow Windows linking well enough, but it seems like not being able to run the linked binary as a test is itself a bug, but one that is not currently causing any pain.

By the symptoms of it, I would guess it is a bug in the workflow itself (how the environment is set up). Perhaps it is as simple as adding `eval "$(ocaml-env cygwin)"`. But these are all just wild guesses, I don't follow the Windows build process either.

rivy:
> 
> 
> Fix self-tests on Windows and enable them in CICD workflow.
> 
> I've moved the test step before the gtk2 build as the gtk2 executable on Windows doesn't run on GHA environment with current workflow scripts (can't find required dll-s).

It should. After each build, the binary and needed DLLs are moved to the STAGING area (test-UI as `unison` and gtk-UI as `unison-gtk`). Just run the self tests on those binaries. You could alter the tests to check CI and run the appropriate staging binary (or both).

tleedjarv:
Thank you @rivy 

The gtk binary should definitely run in PKG_DIR. Running tests after packaging sounds like a good way forward (if just a bit unintuitive). Instead of running `make test` the workflow would have to directly execute `${PKG_DIR}/bin/unison-gtk${EXE_suffix} -selftest` (and same for other binaries). I don't have any problem with that.

> alter the repository tests to check CI and run the other tests on the appropriate staged binary (or both)

I lost you there...

tleedjarv:
I propose that testing-related changes in GHA workflow are a subject for a separate PR (volunteers?) and not make it part of this one.

rivy:
> I lost you there...

Sorry to be opaque. I meant that either that the tests could detect operation under CI and test the correct binary, or (likely a better option now that I'm thinking about it) tests could take an optional `make` option (eg, `BINARY=...`) that they use for testing. Either would take some modification of the tests though...

But, this seems like a good stop-gap until some of the necessary packaging is uplifted into the GTK build step and some sort of test modification is accomplished. I would agree that the work on those could be done at a later date/PR.

I'm happy to see someone care about the Windows tests! 🎉

tleedjarv:
On another note, I did verify that adding `eval "$(ocaml-env cygwin)"` seems to be what is needed to have the dll-s found even before packaging.

gdt:
I think we need to be very careful about what should be part of tests and what should be part of CI.

We should keep in mind that we are off the rails of Free Software by using proprietary CI.  While it's convenient now, I don't think it's a good thing, and we should be careful to have it entangled as little as possible.   Things should be set up so that anyone who wants can set up some other CI system.

Also, with respect to Free Software norms, people being able to build from source has to be the main approach.   It may be that many people use binaries, but that is only ok when it is merely a convenience.

It's an important principle that users can run builds, packaging, and tests.   And use those procedures for CI if they want.  So I don't think we should even consider having behavior differ when run under github CI.   The desire for differences I see as a clue that there is some other problem, and instead we should fix that problem.

So stepping back, there should be
  * documentation of build environment prerequisites
  * a build procedure, that anyone can run by checking out the repo and invoking some short command, via scripts/whatever in the repo
  * a test procedure, that perhaps requires build to have been done, that anyone can run, again scripted
  * an install procedure, that puts the binaries either in the running system, or in some DESTDIR (for packaging), again that anyone can run
  * optionally (as I believe this is really not the proper domain of a project, but obviously users expect this), a packaging procedure that takes the results of the install and bundles it up
  * some kind of CI, that is a wrapper that basically sets up an environment and runs the previous steps.  The CI wrapper should not contain anything that belongs in the previous steps, or rather, if it does, that content should be hoisted to the step itself.

That means running tests on packaged bits to me feels irregular.  Normally, tests are run on the built products before they are installed, and people only install when they pass.

So I think we should
  * fix the unison/unison-gtk build collision, not as a CI patch, but for real.  I don't think it is hard and all that is required is the will to do it (and picking a name).   That takes most of the trouble off the table
  * if after having built, on some platform, one can't run, because of some dll trouble, add a wrapper script that makes that come out right, and then use that wrapper in tests
  * probably a bunch of other small things

Finally, I don't mean to criticize anyone's efforts or how we got where we are.   This is all complicated and we have had many local optimizations.  We are in far far better shape than we ever have been with respect to packaging/etc.   I am merely trying to step back and look at the big picture.

tleedjarv:
Commenting here and not in the mailing list for completeness with previous comments.

The problem of not being able to run gtk binary directly on Windows has a solution, so there is no longer any need for testing to depend on packaging. All those problems now have a solution and the only real remaining one is #394.

Coming back to your earlier comment:

> except about the UI build overwriting the text binary and that sequencing build/test/build/test is the preferred approach. I think build should build both, and then test should test both, if that's how the overall build is configured.

As long as this is not possible yet, as a first step, the sequencing is what can be achieved already today with just a tiny change in Makefile (the test invocation must be done with `-ui text` as @rivy correctly pointed out) while still remaining independent of GHA CI.

rivy:
For me, the pluses of the GHA CICD are that all three major platforms are represented; all platforms can be scripted in a relatively platform-independent manner (eg, `bash` scripting is available on all the platforms); spin-up, build, and testing is generally much faster than other alternatives; and packaged user-accessible releases are now automatically generated for each release. Those properties, in combination, are hard to find or create and make for a much easier development/release process. This all helps keep the platforms at better parity and makes installing and using `unison` much easier for multiple platforms. For me, availability of "specialty" static Linux-like and Windows builds *was* a significant problem. I'd hate for the project to backslide into scatter-shot builds at multiple website locations.

And one positive of the current setup is that the CICD process serves to document the exact steps that can be used to build the MacOS, Linux, and Windows packages. I am leveraging that to develop more modern build documentation. I should have fairly detailed documentation of the Windows build process done soon (#429). I'll try to do some further Ubuntu experimentation and document the Linux build to the same degree. (But, similar instructions for MacOS aren't possible for me given a lack of MacOS hardware and experience.)

Ultimately, I do agree with @gdt that being tied to a specific CICD isn't a good idea. Working toward a more flexible build process which can be used locally and on multiple CICD systems seems wise. I'm happy to continue lending a Windows-centric hand where helpful.

gdt:
You mention "3 major platforms", but I find that notion problematic, because any other operating system  is left out.  Even worse, people in the Free Software community can not do work to add them, because GHA is proprietary software.  That said, I agree that it's useful.

It's also great to encode how to do builds in scripts.  That's  really my point: the main thing should be easily scripted builds that are possible to perform with free tools, with as little a non-free config layer on top as possible.

rivy:
> You mention "3 major platforms", but I find that notion problematic, because any other operating system is left out. Even worse, people in the Free Software community can not do work to add them, because GHA is proprietary software. That said, I agree that it's useful.
> 
> It's also great to encode how to do builds in scripts. That's really my point: the main thing should be easily scripted builds that are possible to perform with free tools, with as little a non-free config layer on top as possible.

I fundamentally support your ideas here, but it’s a question of resources. Maintaining a CICD platform, especially of multiple OS platforms, is a *very* resource intense proposition (of both hardware and human). I don’t see how it can be done independent of donated proprietary processes for projects of any size less than “enormous” that have some sort of donation network. And I think it’s best practice to maintain a CICD process which helps guide and measure development otherwise the project tends to founder off course.

I, again, do agree that depending on a single resource is problematic, and I’d love to help build on a CICD process that is more open. But I don’t know of one that has matching properties...

Making local builds easy is also a great development target which I think has major long term benefits for a project. It’s just hard to do it in a cross-platform compatible way. The DevOps Tooling is just still too rough, IMO. (I’ve been on the lookout for a good cross platform, touring complete, build toolset for a long time and haven’t seen one that ticks all the boxes yet... maybe Deno, Go, or Rust will come through with something soon?)

The “3 major platforms” comment is not meant to be exclusionary, just a practical statement. Those three are essentially 90+% (99% would not surprise me) of the usage of this and most software projects. I’m all for supporting more widely whenever possible, but again it’s resource intensive and tough. If the tooling surrounding the project advances to support cross-platform building and testing, I’m happy to spend some time helping get such a process working. I think multi-platform development is a true long term win for any software project, and I’d enjoy expanding my skills in that area.

$0.02

All the work you folks are doing here is much appreciated! This is likely in the top five pieces of software that I use every day, on and between all the platforms and machines that I develop on. Sincerely... Thanks!

PR #428, opened 2020-11-10
Remove unused value Watchercommon.really_write

pgossman:
Found from a warning when building with dune


tleedjarv:
The same function exists in fswatch.ml
These functions were made redundant by the safe-string changes. I don't see a problem removing them.

pgossman:
Updated to remove really_write from fswatch.ml as well

PR #431, opened 2020-11-13
uimac: Fix Xcode build on OCaml versions >= 4.08

tleedjarv:
Build fix from PR #423


gdt (commented):
OK with me.  Would appreciate additional reviewers.

PR #432, opened 2020-11-13
CICD: Build and publish Unison.app for macOS

tleedjarv:
GHA workflow change from PR #423 
PR #431 must be merged before this.


gdt (commented):
Loks ok to me.   Will hope GHA re-runs after #431 is merged, but it's pretty clear what's going on.

As usual, additional reviewers would be nice, and I'll merge after timeout if not.

PR #433, opened 2020-11-13
uimac: Fix compiler warnings

tleedjarv:
Compiler warnings fix from #423

### DIFF
--- a/src/uimac/Bridge.m
+++ b/src/uimac/Bridge.m
@@ -111,7 +111,7 @@ typedef unsigned int NSUInteger;
     NSString *msg = [NSString stringWithFormat:@"Uncaught exception: %@", [exception reason]];
     msg = [[msg componentsSeparatedByString:@"\n"] componentsJoinedByString:@" "];
     NSLog(@"%@", msg);
-    NSRunAlertPanel(@"Fatal error", msg, @"Exit", nil, nil);
+    NSRunAlertPanel(@"Fatal error", @"%@", @"Exit", nil, nil, msg);
         exit(1);
         return FALSE;
 }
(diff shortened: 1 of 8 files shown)


gdt (approved):
I think this is fine, but I would really appreciate someone who uses the mac UI to run the build produced by CI from this (recently re-pushed) branch.    I'll merge anyway if that doesn't happen in a few days.

bcpierce00:
If someone can remind me where to grab it, I can try it…

gdt:
I just asked for the jobs to be rerun, so this will be tested against master of this minute (with the two merges I just did), vs master of yesterday.

In the checks tab, there is an 'artifacts' pulldown.   THe names look alike, but if you mouse/hold you can figure out which one is which.  It will take a bit, maybe 30m, for the new build run.

gdt:
Click checks, then CICD to get https://github.com/bcpierce00/unison/actions/runs/362997071

bcpierce00:
The plain app from 26342157.zip opened a terminal window and then did the expected text-UI thing.  The gtk one failed like this:



Was that expected?  Should there also be a native macOS GUI version?

   - B

tleedjarv:
> The gtk one failed like this:

The image is not visible.

> Should there also be a native macOS GUI version?

Yes there should be, and this PR only impacts the native macOS GUI. It seems that this PR builds againts previous GHA workflow before the uimac build changes were merged (even after the rerun?). @gdt can you see if it's possible to rerun again - though if you've done it once, it doesn't seem to have picked up the new workflow.

As a proxy, you can download a build with ".app" in the suffix here https://github.com/bcpierce00/unison/actions/runs/352740603
It has exactly the same commits. You may need to manually remove the quarantine attr (as mentioned [here](https://github.com/bcpierce00/unison/issues/422#issuecomment-723738735), for example).

bcpierce00:
Seems to work!

gdt:
With a positive test report, I'm going to hit merge.

About the CI/etc.my guess is that this is running CI on the last commit in the PR, and if it isn't rebased after other merges, it won't be using that other changed CI code.   We'll see how it is once merged.

Issue #318, opened 2020-02-17
Store logs in $HOME/.unison, rather than polluting $HOME


LocutusOfBorg:
Hello, Ubuntu applied a similar patch some years ago, to avoid HOMEDIR being polluted with many conf files

```
Description: Change logfile location
 ... by prepending .unison to name of logfile in ubase/trace.ml
Author: Tan Rui Boon <ruiboon@gmail.com>
Bug-Debian: http://bugs.debian.org/484997
Bug-Ubuntu: https://bugs.launchpad.net/bugs/234365
Forwarded: no
Last-Update: 2011-09-10

--- a/ubase/trace.ml
+++ b/ubase/trace.ml
@@ -113,10 +113,10 @@ let logging =

 let logfile =
   Prefs.createFspath "logfile"
-    (Util.fileInHomeDir "unison.log")
+    (Util.fileInHomeDir ".unison/unison.log")
     "!logfile name"
     "By default, logging messages will be appended to the file
-     \\verb|unison.log| in your HOME directory.  Set this preference if
+     \\verb|unison.log| in your .unison directory.  Set this preference if
      you prefer another file."

 let logch = ref None

```

is it possible to upstream this change? thanks!

gdt:
A PR would help; seems like a reasonable change.

glondu:
Note that the UNISON variable should be honored, maybe use Util.fileInUnisonDir instead?

LocutusOfBorg:
Sorry but I don't speak ocaml since lots of time, can anybody please do the pull request?

glondu:
> can anybody please do the pull request?

I will.

PR #436, opened 2020-11-24
Make logfile default location .unison rather than $HOME

pgossman:
Addresses #318

Accessing the unison directory was not initially possible from ubase/trace.ml—where
the logfile preference is defined—because unisonDir was defined in Os. Currently,
to access the unison directory from Util, Util defines a function reference where top-level
modules supply Os.fileInUnisonDir to it (via Util.supplyFileInUnisonDirFn).

Instead of moving the logfile preference from trace.ml to some module above ubase,
this PR moves the definition of Os.unisonDir and Os.fileInUnisonDir to Util. This has the
added benefit of being able to delete three existing duplicated calls to
Util.supplyFileInUnisonDirFn. The unisonDir and fileInUnisonDir values now reference
those in Util.

### DIFF
--- a/src/fswatchold.ml
+++ b/src/fswatchold.ml
@@ -25,7 +25,7 @@ let debug = Util.debug "fswatch"
 
 let watchinterval = 5
 
-let watcherTemp archHash n = Os.fileInUnisonDir (n ^ archHash)
+let watcherTemp archHash n = Util.fileInUnisonDir (n ^ archHash)
 
 let watchercmd archHash root =
   let fsmonfile =
--- a/src/main.ml
+++ b/src/main.ml
@@ -191,12 +191,6 @@ let init () = begin
     exit 0
   with Not_found -> () end;
 
-  (* Install an appropriate function for finding preference files.  (We put
-     this here just because the Prefs module lives below the Os module in the
-     dependency hierarchy, so Prefs can't call Os directly.) *)
-  Util.supplyFileInUnisonDirFn
-    (fun n -> Os.fileInUnisonDir(n));
-
   (* Start a server if requested *)
   if Util.StringMap.mem serverPrefName argv then begin
     catch_all (fun () ->
--- a/src/os.ml
+++ b/src/os.ml
@@ -305,19 +305,10 @@ let fullfingerprintEqual (fp, rfp) (fp', rfp') =
 
 (* Gives the fspath of the archive directory on the machine, depending on    *)
 (* which OS we use                                                           *)
-let unisonDir =
-  try
-    System.fspathFromString (System.getenv "UNISON")
-  with Not_found ->
-    let genericName =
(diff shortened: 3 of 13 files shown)


gdt (approved):
Thank you for keeping the code cleanup and the functional change in separate commits.

I would like to have review from someone who knows ocaml better to ensure that the code cleanup results in the same behavior under all circumstances.    (I don't mean to imply that I think it doesn't.)

I think having the log file in .unison is a good change; I have long noticed it appearing in $HOME where it obviously doesn't belong.

I'll wait for others to comment, and will merge this anyway after 7 days even if not.   Thanks for contributing to unison.

tleedjarv (approved):
Looks ok, with a few comments.

I don't particularly like the code duplication with `isMacOSXPred` but I see this is difficult to avoid here.

Why have you chosen to not replace all occurrences of `Os.unisonDir` and `Os.fileInUnisonDir` with the new `Util.` functions? Either way, I suggest you move code comments from os.ml to util.ml.

pgossman:
> Why have you chosen to not replace all occurrences of Os.unisonDir and Os.fileInUnisonDir with the new Util. functions? Either way, I suggest you move code comments from os.ml to util.ml.

I haven't worked on open-source before and I wasn't sure what is appropriate for the scope of change in a PR, so I stayed minimal. I agree that it is better to remove the values in Os and replace occurrences with the new ones in Util. 

The commit that was just pushed removes the values from Os and moves the comments as you suggested.

Thank you for reviewing.

tleedjarv:
> I haven't worked on open-source before and I wasn't sure what is appropriate for the scope of change in a PR, so I stayed minimal.

Reasonable approach. And that's why the review is open. Everyone can help and voice their opinion if this is too much or too little. I am not a guardian of this project either.

The change itself looks good to me.

The merge commit in the PR may not be a good idea, though. I don't know how it will look like once the PR itself is merged but I would prefer your commits rebased. @gdt or anyone else, any clue or preference?

pgossman:
> The merge commit in the PR may not be a good idea, though.

Sorry about that. I just rebased and reapplied the commits.

tleedjarv (approved):
All looks good to me.

gdt:
Answering question from earlier, I have a strong preference for rebasing and a fully-cleaned up set of commits, to the point where I'd be very uncomfortable approving a merge that isn't clean.  It's not that hard to do, and it makes looking at the history far easier.  I see the rebase has happened - thanks! - but since I was asked wanted to answer.

gdt (approved):
This looks ok to me too.

Scope of a branch merge is always tricky, but I see all the cleanup/moving of the helper functions as clearly belonging together.   If anything, it's the move of the logfile that should be separate, but for this project, a separate commit is good enough, in the grand balance between making progress easier and making understanding what happened later easier.

PR #441, opened 2020-11-29
[RFC] CICD: Initialize opam with requested OCaml version

tleedjarv:
This PR is intended for discussion.

There is exactly one benefit from this change, and that is conservation of resources by not having to compile an entire ocaml compiler twice and by halving the size of opam cache that needs to be compressed/decompressed, stored and uploaded/downloaded.

I would argue that this change belongs upstream in the action that sets up OCaml in GHA but I don't know how that action works (and I don't know how the Windows setup works). @rivy you can possibly help here.

Incidentally, this change did uncover an issue with OCaml < 4.06 (this is what you see as the failed run). It is fixed since 4.06 by https://github.com/ocaml/ocaml/commit/adcb4497968dbdc1b0db97998488ad20c02fe225#diff-d8b9301c9b040433303c891069704b2eff356f3b6d1756ff2d9b761192f05a16

### Background

OCaml compiler in GHA workflow is set up with opam, but this commonly includes an unnecessary step of building an entire ocaml compiler twice. First, opam init will use the system ocaml compiler, or if that is lacking, build the latest ocaml release. Then, opam switch will build the requested version of ocaml compiler.

This PR creates an opam configuration file indicating default-compiler with requested OCaml version. Setting default opam configuration instructs opam to use the correct version of ocaml compiler already during opam init, thereby skipping the building step during opam switch.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -92,6 +92,8 @@ jobs:
         mkdir -p '${{ steps.vars.outputs.STAGING_DIR }}'
         mkdir -p '${{ steps.vars.outputs.PKG_DIR }}'
         mkdir -p '${{ steps.vars.outputs.PKG_DIR }}'/bin
+        # initial opam config
+        echo "default-compiler: \"ocaml-system\" { = \"${{ steps.vars.outputs.OCAML_VARIANT }}\" } | \"ocaml-base-compiler\" { = \"${{ steps.vars.outputs.OCAML_VARIANT }}\" }" > ~/.opamrc
 
     - name: Enable/config MSVC environment (if/when needed)
       uses: ilammy/msvc-dev-cmd@v1.3.0
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -106,7 +106,7 @@ jobs:
       id: cache-opam
       with:
         path: ~/.opam
-        key: v1-${{ matrix.job.os }}-opam-${{ matrix.job.ocaml-version }}
+        key: v2-${{ matrix.job.os }}-opam-${{ matrix.job.ocaml-version }}
 
     - name: Use OCaml ${{ matrix.job.ocaml-version }}
       uses: CICD-tools/ghactions-ocaml.toolchain@dev


tleedjarv:
Have to also flush GHA workflow cache to see the effects of this change.

gdt:
This seems reasonable in concept to me.  Would like @rivy to review and approve before I merge.

rivy:
Theory seems good, but I can't see any change in setup speed or cache size.
Likely the cache needs to be newly primed after the key change.
Have you tested it with a commit past the cache flush commit?

And the problem with OCaml 4.05?
From the GHA logs, it looks like OCaml is installed but lablgtk is failing installation. Is that what you mean?
And if there is a bug compiling with OCaml v4.05, should that build configuration be removed? Or should all v4.05 build configurations changed to v4.06 or v4.07? I'm not sure how important v4.05 builds are...

tleedjarv:
> Theory seems good, but I can't see any change in setup speed or cache size.
> Likely the cache needs to be newly primed after the key change.
> Have you tested it with a commit past the cache flush commit?

Not in this repository yet, but I definitely should. It also seems that GHA has been under increasing load, and with caching the benefits of this PR seem decreasing. Especially since it doesn't help on Windows where builds are currently the slowest. Getting caching working on Windows would be good.

> And the problem with OCaml 4.05?
> From the GHA logs, it looks like OCaml is installed but lablgtk is failing installation. Is that what you mean?
> And if there is a bug compiling with OCaml v4.05, should that build configuration be removed? Or should all v4.05 build configurations changed to v4.06 or v4.07? I'm not sure how important v4.05 builds are...

Yes, that's what I meant. It seems that this bug only appears on macOS, so the easy solution--if considering merging this PR--is to just remove building with 4.05 for macOS.

I think overall, it seems that with unpredictable GHA availability and load patterns, this PR may not bring much value after all, even if in theory it should. It does not seem worth the effort maintaining in this repository alone, if it doesn't get upstreamed in some form.

tleedjarv:
Closing this due to lack of availability and priority.

Issue #367, opened 2020-08-23
Add pre/suffix option for temporary files so that other processes/systems will ignore them


ChristianRiesen:
Unison works great for my uses but I keep running into an annoying situation in my setup. The temporary files unison creates sometimes get picked up by an external process over which I have no control over. However that process ignores files if they start with .~ and a couple other rules.
For that it would be great to have an option both for a suffix and a prefix that would be only used with the temporary files.
I hope that would not be too complex a change but would make unison more flexibel.

gdt:
It might also be nice to just change the default to something that is ignored by most things that try to ignore some files.  Perhaps ending in ~ is enough?

ChristianRiesen:
From https://help.dropbox.com/installs-integrations/sync-uploads/files-not-syncing

> Temporary files
> When some applications (such as Microsoft Word, Excel, or PowerPoint) open a file, they will often save a temporary file in the same directory and name it in one of the following ways:
> 
> Name begins with ~$ (a tilde and dollar sign) or .~ (a period and tilde)
> Name begins with a tilde and ends in .tmp, such as ~myfile.tmp
> Dropbox doesn’t sync these temporary files on any operating system.

As you see there are different options like `~file.tmp` or `.~file` or `~$file ` for this specific use case. So having the option of prefix/suffix would be ideal, though as you can see only prefix would already work as well, just so Dropbox isn't greedy and tries to grab the temporary unison files that can be in the folder.

gdt:
I see.  This is not following the usual gitignore pattern (which follows longstanding traditions), so it does seem that having his be configurable would be useful for various people to accomodate various other software.  (I don't think aligning unison to some particular proprietary software  is a good approach, personally.)

ChristianRiesen:
Ran into this issue now about a dozen times since posting this. I resolved to have a script that I can trigger by hand, searching for unison files and removing them. Less than ideal though.
Anything I can do to get this done? :)

gdt:
Yes, you can make a code change and test it, and then submit a pull request with that code change and the corresponding docs change, all while being careful to do it in a way that will seem acceptable to most.    Probably not the answer you wanted, but that's how it is.

ChristianRiesen:
I can code, but not this language unfortunately :) Might give it a go anyways and see how far I get.

tleedjarv:
If you don't need to make it configurable then have a look here https://github.com/bcpierce00/unison/blob/master/src/os.ml#L42-L43

[...]

Issue #372, opened 2020-09-12
Add preference file that is always read


gdt:
The documentation almost implies that default.prf is always read, but really it is read only if there are no arguments.  When invoking unison on a target over ssh, it is given two roots, and normally this works.   But if one wants to set clientHostName on that machine, e.g. because the hostname changed, there is no way to do that.

To fix, add a new filename, perhaps options.prf, which is read first, for every invocation.

A workaround is to set UNISONLOCALHOSTNAME in ~/.ssh/environment after enabling PermitUserEnvironment.  But this only works for options with corresponding environment variables.

PR #447, opened 2020-12-04
Enable customization of temporary file prefixes and suffixes

pgossman:
Addresses #367.

This PR adds two flags, -tempfileprefix and -tempfilesuffix with defaults of ".unison" and ".unison.tmp".

I tested this change locally by interrupting a sync to leave an old temp file on disk, and unison found it successfully on the next run and was able to finish the sync. However, if a sync is interrupted, and then unison is run again with different args to tempfileprefix/suffix, then it does not pick up the old archive. It instead views the old archive as a new file to sync. I am unsure if this edge case should be handled, and if so, how to go about implementing a fix.

Would reading environment variables be preferred over new flags? I could see the case for it considering an envvar is used for the location of the .unison dir. I am unsure when one is preferred over the other.


bcpierce00:
Doing it with environment variables has two advantages: (1) the settings can be different on different hosts, and (2) doesn't change the wire protocol.

I still wonder whether it might be best to just change the default names to something different, globally, and leave it at that.

gdt:
I generally dislike environment variables rather than explicit configuration, when explicit configuration is possible.

I don't see it as a problem that running unison again with different settings after an interrupted run doesn't do what you might want.

Agreed about changing the defaults if there is a single plan, but so far I haven't seen a global analysis of all the situations and issues.    (I also haven't seen the counter-notion, of "here's how to configure Dropbox to ignore these files"; it's not clear where the line of responsibility should be.)    All that said, if there is a convention on each of Windows, Mac, POSIX that is already recognized and aligning to it isn't problematic, that sounds like a good first step.

gdt:
> Doing it with environment variables has two advantages: (1) the settings can be different on different hosts, and (2) doesn't change the wire protocol.

I do not follow how environment variable vs command-line (or in a profile, or a unison-wide config file) bears on either making thing different on different hosts, or the wire protocol.   I don't see any reason for temp file names to have to be consistent across hosts being synced.

Overall I feel like any unison-wide config that should happen for all profiles belongs in UNISONHOME/something, which is the subject of #372.  Otherwise we are just moving things that should be into a config file into environment variables one at at time.  However, this is awkward and error prone, as the default ssh config does not set environment variables.

gdt:
And, thanks for your contribution.   This may well go through a bit of rototill, but that's part of how progress happens.

tleedjarv:
I don't know if this is really an issue that the current proposal does not allow different prefixes and suffixes on different hosts, but this is something I wanted to point out as well. I think this would only ever matter if one of the hosts is Windows and the other not, if even then.

However, having user preferences in the first place seems only really useful to cover for file system/OS special needs, which in turn almost implies that the preferences are different on client and server.

For this scenario, and if talking about built-in prefixes and suffixes, the simplest thing I can think of would be
``` ocaml
let tempFilePrefix = if Util.osType = `Win32 then "~$unison." else ".~unison."
```
(taking the prefix examples from #367)

bcpierce00:
This seems OK to me.  (Not knowing too much about these conventions on different OSs…)

pgossman:
> I also haven't seen the counter-notion, of "here's how to configure Dropbox to ignore these files"; it's not clear where the line of responsibility should be.

It seems the responsibility should be on Dropbox to allow filters on which files to exclude. Other backup services (Tarsnap, pCloud, Google Drive) support this. The only backup services I could find that don't support exclude patterns are Box and Dropbox, which will both ignore files starting with ".~" on Unix and "~$" on Windows.

As of now, the only use case for this PR is to avoid pickups by Dropbox. Perhaps it would be better to revisit this if users have other cases for customizing temp files, rather than introduce new config options that are not yet needed. I have updated this PR to a change in the default prefix per the suggestion from @tleedjarv.

tleedjarv (approved):
I can't say anything about the prefixes but otherwise this looks good to me.

ChristianRiesen:
Just came across this again and wondered if this could be merged.
Ping @pgossman and @bcpierce00 perhaps to take another look?

bcpierce00:
Here are the tradeoffs I see:

- Changing the default names and doing nothing else has the (significant) advantage of being simple.  But (1) if this change is motivated by Windows concerns, do we need to change anything on non-Windows systems, and (2) even on Windows we sure that this change won't break things for anybody?
- Looking at an environment variable has three advantages over adding preferences: (1) it allows different choices on different hosts, and (2) it avoids changing the wire format (but maybe that's less of an issue with other changes in progress), and (3) it avoids adding more preferences to the (already large) list that people see when they ask what prefs are available.  The downside is that these customizations are harder to find.

ChristianRiesen:
The default change would work for me perfectly. 

A potential issue is of course of someone configured externally some filter/ignore patttern and expects `.unison.` to be there instead of the two different ones now, depending on OS. If you are ok with the default change @bcpierce00 then that would solve my issue for certain.

If you envision changes down the road you could still add a parameter once those changes are done.

gdt:
I am fuzzy now but I think this stalled partly because I'm not inclined to accommodate proprietary software that arguably should have a configurable ignore list on that side,  and because it seems like there are perhaps OS-specific norms that we should follow.  And further, perhaps, that the normal OS-specific norms aren't followed by dropbox.

rereading, it seems like this is basically a workaround for dropbox.

acolomb:
I do see value in having the same (default) prefix on different OSes.  Using different synchronization solutions in conjunction with Unison, and changing their ignore patterns based on which host we're on would make it harder.

Regarding the configuration method, I'd slightly prefer the option over environment vars.  I wouldn't use it though, and in this case I think predictability is more important than accommodating other systems' shortcomings. (And tbh, Dropbox is really turning worse every month in so many aspects.)

ChristianRiesen:
@gdt My initial use case comes up from Dropbox, yes. However now I'm also having another issue with this as there is a tool (I have no control over) that I can't exclude unison files from either. The `.~` prefix would already solve that issue as well.

IMHO the endresult is that there is no actual standard in the "what file prefix/suffix to ignore" patterns and only having a single one (even if it matches a lot) will not match all use cases. My personal case would be solved with `.~unison` instead of `.unison` but a prefix option would make it a much more flexible approach, especially since there are no true standards in that realm.

ChristianRiesen:
> I am fuzzy now but I think this stalled partly because I'm not inclined to accommodate proprietary software that arguably should have a configurable ignore list on that side, and because it seems like there are perhaps OS-specific norms that we should follow. And further, perhaps, that the normal OS-specific norms aren't followed by dropbox.
> 
> rereading, it seems like this is basically a workaround for dropbox.

Just stumbled across this again. There is a tangible usecase to make it work with Dropbox, sure. It also supplies much more control over the temp files, which can be then configured to be ignored. An example I just came across again recently is an indexing engine that has a list of prefixes hardcoded. Another is a backup tool, again no way to tell it to just ignore these files as it takes it's own "logic" of what is and what isn't a temp file. Allowing this to be configureable would make unison a better tool, that helps in all those instances where other tools are failing.

chrisblossom:
Any status update on this? It would help a small subset of users—would a configuration PR be accepted?

gdt:
I just re-read the discussion.   I remain concerned about a few things, not really changed from earlier:
  - stacking sync solutions is messy.  I mostly try not to do it, but I do run unison between syncthing and nextcloud, to deal with platforms that seem not to allow syncthing (iOS).
  - This is basically an accomodation of proprietary software (dropbox) that should be configurable about excluding, as virtually all other similar sync software is.   There's no link to a dropbox bug report in a public tracker.
  - The current choice is probably not great.  It would help to have an analysis of conventions/etc. for various OS and sync schemes, to understand what kind of commonality there is.   Configurability in general hurts, and of course it helps sometimes.
  - It seems likely that what's ignored normally (either by default or easily configurable) may vary by platform (where platform is normal POSIX-like system as the main case, macOS, and Windows, I think).    However, people also use remote file systems and use unison on them, and that leads to further complexity about unison on POSIX via CIFS to windows and a mac accesing that same FS and also running dropbox.  This gets crazy pretty fast, and I don't know where we should be drawing the line.  One place is "if you're using a second sync protocol on the same bits, then unison, the files, and the 2nd sync must all be on the same computer".   (Personally I don't run unison via remote filesystems -- I use it *instead* of remote filesystems.)

gdt:
Long message sent to `unison-hackers@`.  Please feel free to follow up there.

Issue #439, opened 2020-11-28
CI-built macos Unison.app fails to start on 10.13


gdt:
Downloading ``Unison-4713e608.ocaml-4.09.1.macos-10.15.app.tar.gz``, unpacking, and running ``Unison.app/Contents/MacOS/Unison`` results in
```
dyld: lazy symbol binding failed: Symbol not found: ____chkstk_darwin
  Referenced from: /private/tmp/U/Unison.app/Contents/MacOS/Unison
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: ____chkstk_darwin
  Referenced from: /private/tmp/U/Unison.app/Contents/MacOS/Unison
  Expected in: /usr/lib/libSystem.B.dylib

Abort trap: 6
```

The ``unison`` binary from the non-app macOS build does run.

tleedjarv:
Unison.app itself is built with `-mmacosx-version-min=10.6` OCaml compiler is not. `____chkstk_darwin` is being pulled in by ocaml libs.

I think the issue can be solved by, when building OCaml by opam, passing `-mmacosx-version-min=10.6` to C compiler and `-ccopt -mmacosx-version-min=10.6` to ocaml compiler. Perhaps @rivy you know how to do this?

rivy:
Sorry @tleedjarv, I don't have an answer for you. MacOS is a bit out of my zone of experience...

If you can tell me what incantation to use with `opam` and OCaml, I might be able to integrate it into 'CICD-tools/ghactions-ocaml.toolchain' (or preferably, get it included in the original action, [avsm/setup-ocaml](https://github.com/avsm/setup-ocaml)).

Unfortunately, I don't have any MacOS hosts to test and iterate upon.

mroi:
I can prevent the inclusion of the `____chkstk_darwin` symbol by building the OCaml compiler with this environment variable set:
```
MACOSX_DEPLOYMENT_TARGET=10.6
```
I have set this in my shell before running `./configure` in `ocaml-4.09.1`. Maybe this can be included in the CICD builds?

gdt:
Thanks @mroi for the confirmation of that.

I wonder slightly why you pick 10.6, or maybe that's just the oldest supported by the SDK and it works ok, so why not?  Or maybe that's what the other build has...

tleedjarv:
I don't have access to macOS either. I was experimenting a little with GHA macOS environments.

[...]

PR #448, opened 2020-12-04
prevent collapsed toolbar on Big Sur

mroi:
MacOS Big Sur features a new toolbar design, where the window title gets inlined as the leftmost toolbar item. This causes window layout issues with the Unison macOS app, because the window sizes and toolbar button arrangements only work well for the pre-Big-Sur toolbar style.

This change will configure toolbars to always use the pre-Big-Sur style. I have tested on Big Sur and it works well there. I have **not** tested on previous macOS versions, since I no longer have access to any. The CI run seems to be unaffected by this change when building for macOS 10.15.

### DIFF
--- a/src/uimac/English.lproj/MainMenu.xib
+++ b/src/uimac/English.lproj/MainMenu.xib
@@ -13,7 +13,7 @@
                 <outlet property="delegate" destination="209" id="571"/>
             </connections>
         </customObject>
-        <window title="Unison" allowsToolTipsWhenApplicationIsInactive="NO" autorecalculatesKeyViewLoop="NO" releasedWhenClosed="NO" visibleAtLaunch="NO" frameAutosaveName="mainWindow" animationBehavior="default" id="21" userLabel="Window">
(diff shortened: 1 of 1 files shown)


gdt:
Thanks.  I really want to see a report from somebody on earlier versions.

There is also #439 which may be making the CI testing not have the coverage we wish it did; I'm nervous about changes until we fix that or somebody builds on 10.13.  (My view is that pre-10.13 is ancient, and certainly pre-10.11 is irrelevant, but 10.13 is pretty common due to Apple's policy of de-supporting perfectly functioning hardware that has enough RAM to run modern bloatware.)

mroi:
I have one option of testing my PR on a 10.15 machine. I will report what happens.
I’ll also have a look at #439, because my custom-built ocaml does not pull these `____chkstk_darwin` symbols in.

gdt:
Please do report what happens on 10.15.  I am nervous about changes only tested on the new version.  Really, I'd like to see testing on 10.13 too, but given how much testing happens, that isn't a reasonable demand.

mroi:
I have tested the change on both Big Sur (macOS 11.0) and Catalina (macOS 10.15) and it works fine on both systems. The older systems seem to ignore the new toolbar style property.

PR #449, opened 2020-12-06
uimac: Disable Xcode cross-compiling for other archs

tleedjarv:
GHA has upgraded the default Xcode version and while GHA build environments are x86_64, the new Xcode attemps to cross-compile for the new M1 (arm64).

Disable cross-compilation as it will fail due to OCaml code not being cross-compiled. This will remove CICD errors you've been seeing lately. I don't know if this is the correct approach but it seems to get the job done.


mroi:
I’d say this is exactly the right approach.

PR #451, opened 2020-12-10
uitext: Periodically write progress to log

### LINKED ISSUE #28: Write some progress into log indicating possible ETA
Especially useful when unison is running with "-repeat watch" setting in the background. While in the foreground progress it displays what file/directory is currently being processed. On large roots or high system load it can take 10-20 minutes to accomplish.

Having some periodic information in the log is preferable to ensure it's running properly and when to expect new changes applied.

tleedjarv:
In terse (incl. silent) and batch modes, write sync progress information to log. Progress is logged every 60 seconds to limit log file growth.

Progress display on the screen remains as is.

Closes #28

### DIFF
--- a/src/uitext.ml
+++ b/src/uitext.ml
@@ -742,7 +742,7 @@ let doTransport reconItemList =
          Uutil.Filesize.zero items)
   in
   let t0 = Unix.gettimeofday () in
-  let showProgress i bytes dbg =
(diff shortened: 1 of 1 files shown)


gdt:
Looks reasonable to me.  Awaiting review by an ocaml speaker and somebody else saying they ran it, or timeout.

PR #452, opened 2020-12-10
stasher: Fix creation of stashed directories

### LINKED ISSUE #30: -backupcur causes failure when symlink is replaced with a directory
Here is an example script which demonstrates the problem.
-batch and -auto are just for convenience of the script.
# !/bin/bash

set -x
d=$(mktemp -d)
cd $d
mkdir -p root1/dir/bin
touch root1/dir/bin/bin
ln -s /usr/bin/bin root1
x="/usr/bin/unison-2.48 -root root1 -root root2"
HOME=$PWD $x -backupcurr "Regex ._" -batch
tree -a # just to show what files exist
rm root1/bin
mv root1/dir/bin root1
HOME=$PWD $x -backupcurr "Regex ._" -batch -auto
echo $? # prints 2
tree -a # files synced despite error, but backup dir is unchanged
touch root1/otherfile # test syncing some file now
HOME=$PWD $x -backupcurr "Regex .*" -batch -auto
echo $? # prints 3
tree -a # now files did not sync, backup dir is unchanged
cd; rm -rf $d # cleanup temp dir

tleedjarv:
When creating a stashed directory in the backup directory then we must check if such directory already exists or not. Previous code did this by `Os.exists` which just checks for the absense of object. If the object was present then the code just assumed it is a directory. This will cause problems when the type of object changes (from a symlink to a directory, for example).

The fix is to check if the name exists and is actually a directory. If not then rotate backup versions until the new directory can be created.

Closes #30

### DIFF
--- a/src/stasher.ml
+++ b/src/stasher.ml
@@ -300,13 +300,13 @@ let showContent typ fspath path =
 let backupPath fspath path =
   let sFspath = stashDirectory fspath in
 
-  let rec f i =
+  let rec f path i =
     let tempPath = makeBackupName path i in
     verbose (fun () -> Util.msg "backupPath f %s %d\n" (Path.toString path) i);
     if Os.exists sFspath tempPath then
       if i < Prefs.read maxbackups then begin
         verbose (fun () -> Util.msg "need to rename backup file\n");
-        Os.rename "backupPath" sFspath tempPath sFspath (f (i + 1))
+        Os.rename "backupPath" sFspath tempPath sFspath (f path (i + 1))
       end
       else if i >= Prefs.read maxbackups then
         Os.delete sFspath tempPath;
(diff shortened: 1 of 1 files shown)


tleedjarv:
Code in the fix calls `mkdirectories backdir` again. Now that I think about it, this is actually not necessary. `Os.createDir` should work just the same. I used `mkdirectories` due to backupprefix, as mentioned in the comment. But that should not be needed - if `f` returns a new path, it's implicit that it should have created the tree for that path. This is not working today with backuplocation 'central'.

I think this is due to the concept of including directories in backupprefix, while backuplocation = 'central', being flawed (or then I don't understand it). There is a special case in the code (function `makeBackupName`) that skips the prefix for the first version. This is what breaks and prevents including directories in backupprefix while backuplocation = 'central'. The backupprefix tests work because they use backuplocation = 'local' and they do not test the versioning.

tleedjarv:
It became clear to me that this issue is only relevant for backuplocation = 'central'. The fix is updated accordingly.

PR #453, opened 2020-12-11
CICD: Use Xcode 11.7 for OCaml 4.05 on macOS

tleedjarv:
Bulding lablgtk with Xcode >= 12 and OCaml < 4.06 will fail. It appears to be an OCaml bug, which is fixed since OCaml 4.06.


gdt:
Is it safe that this invocation of select cannot affect anything else?  Is each build done in a fresh image?   If that's true, and I'm guessing it is, then this seems fine and I'll merge it.

tleedjarv:
Yes, every build is done in a fresh image. The other versions are not impacted.

PR #460, opened 2021-01-01
fix CI/CD builds for older macOS platforms

### LINKED ISSUE #439: CI-built macos Unison.app fails to start on 10.13

mroi:
Globally setting the deployment target to 10.6 (oldest version supported by Unison) should produce binaries suitable for older macOS platforms. Both the `opam` build of the OCaml compiler and the unison build respect this environment variable and it does not harm other platforms like Linux.

Should fix #439.


gdt:
Is there no reasonable way to set this only on mac?   I agree it shouldn't hurt, but it's a bit icky to leak mac things into other build environments.

mroi:
You are right. I’ve changed the patch, this should work as well.

mroi:
You are right. I have used a less invasive approach that also seems to work.

gdt:
Thanks for addressing my concerns so fast!

mroi:
This fix was ineffective, because the cache in the GitHub workflow confused my testing. The (hopefully) real fix is #461.

PR #465, opened 2021-01-14
Properly skip empty output from ssh

### LINKED ISSUE #328: Add workaround for buggy ssh password prompt behavior in macOS 10.15.4
Hi,
I am syncing my MacBook and iMac data folders with Unison 2.51.2 for several years now and it worked great.
But after updating macOS Catalina from version 10.15.3 to 10.15.4 released on 23rd March 2020 Unison is not able to connect to the second computer. It is just showing "Connecting..." and nothing more happens.
I already checked the SSH connection between both computer via Terminal and it is working in both directions. Also there was no change in the .prf-file.
Does some has any fix or idea how to solve this?

tleedjarv:
Fixes #328
The first commit fixes the origin of unrecognized prompt from ssh. The second commit fixes the GUI hanging due to the unrecognized prompt.

### DIFF
--- a/src/terminal.ml
+++ b/src/terminal.ml
@@ -230,7 +230,7 @@ let rec termInput fdTerm fdInput =
       Lwt.return None
     else
       let query = Bytes.sub_string buf 0 len in
-      if query = "\r\n" then
+      if query = "\r\n" || query = "\n" || query = "\r" then
         readPrompt ()
       else
         Lwt.return (Some query)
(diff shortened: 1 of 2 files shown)


tleedjarv:
It turns out one of the underlying issues was not specific to macOS.
OpenSSH added a single `\r` in front of the password prompt in version 8.0 (https://github.com/openssh/openssh-portable/blob/4ae7f80dfd02f2bde912a67c9f338f61e90fa79f/readpass.c#L135-L141). Confirmed on Linux that this affects other platforms.

PR #468, opened 2021-01-17
Disable ANSI color output by default on Windows

tleedjarv:
It turns out that ANSI color support in Windows is... a complex matter.
Disable default color output on Windows.


gdt:
I intend to merge this next time I notice it and checks have completed.   Speak up if you object.

PR #473, opened 2021-01-30
Get correct ctime in Windows

tleedjarv:
Standard Win32 API functions do not provide the status change time (ctime) of files and directories. A common substitute is to provide creation time, which is actually btime (and not available in the traditional POSIX stat interface) and is not usable as ctime.

On some filesystems, it is possible to get the real ctime using Windows syscalls via the Native API. Using Native API is a bit tricky, though. Native API header files (for non-driver development?) are not available, so relevant type definitions and function prototypes must be included in your own code. Further, the functions are not (easily?) available for build-time linking and must be dynamically linked at runtime.
This approach also ensures compatibility with other compilers, like Mingw.

This patch borrows code from libuv to get the definitions and prototypes and do the runtime linking.

If the Native API function `NtQueryInformationFile` is available then this patch will use it, making correct ctime available in OCaml code, else it will fall back to Win32 API. The fallback is to using creation time, which is not a suitable alternative to ctime in any way. This is only done for compatibiltiy with the previous code.

For file times, the previous code using Win32 API discarded fractions of a second (not sure if this has been intentional). This patch does not change that and uses the same time conversion formulas as is.

Getting correct ctime is not particularly interesting by itself. I am soon adding Windows code to ACL synchronization PR and then the ctime will provide actual value. Nevertheless, this code is not directly related to ACL synchronization and could be of general value.

**About licensing**

This PR includes code borrowed from libuv (MIT license). The borrowed code is not in any way specific to libuv or modified for libuv use, and the same code is also present in for example Wine (LGPLv2.1+), ReactOS (GPLv2), Cygwin (LGPLv3+), and probably elsewhere.

### DIFF
--- a/src/fs.ml
+++ b/src/fs.ml
@@ -69,5 +69,6 @@ let fingerprint f = System.fingerprint (Fspath.toString f)
 
 let canSetTime f = System.canSetTime (Fspath.toString f)
 let hasInodeNumbers () = System.hasInodeNumbers ()
+let hasCorrectCTime = System.hasCorrectCTime
 
 let setUnicodeEncoding = System.setUnicodeEncoding
--- a/src/system/system_generic.ml
+++ b/src/system/system_generic.ml
@@ -93,6 +93,13 @@ let canSetTime f =
    have access to the lower 32 bits on 32bit systems... *)
 let hasInodeNumbers () = isNotWindows
 
+(* Cygwin can apparently provide correct ctime.
+ *
+ * With current OCaml Unix library, ctime is not correct on Win32.
+ * This can change in future, in which case [hasCorrectCTime] should
+ * be made dependent on OCaml version. *)
+let hasCorrectCTime = isNotWindows
+
 (****)
 
 type terminalStateFunctions =
--- a/src/system/system_intf.ml
+++ b/src/system/system_intf.ml
@@ -47,6 +47,13 @@ val fingerprint : fspath -> Digest.t
 val canSetTime : fspath -> bool
 val hasInodeNumbers : unit -> bool
 
+(* [hasCorrectCTime] is true when [stat] and [lstat] return the status change
+ * time. This is commonly broken on Windows, where creation time (completely
+ * unrelated to ctime; it is the birthtime) is returned instead. However, it
+ * is possible to get the correct status change time on Windows, which is why
+ * [hasCorrectCTime] can have a different value on different systems. *)
+val hasCorrectCTime : bool
+
 end
 
 module type Full = sig
--- a/src/system/system_win.ml
+++ b/src/system/system_win.ml
@@ -313,6 +313,10 @@ let canSetTime f = true
    number of "a". *)
 let hasInodeNumbers () = true
 
+external hasCorrectCTime_impl : unit -> bool = "win_has_correct_ctime"
+
+let hasCorrectCTime = hasCorrectCTime_impl ()
+
 (****)
 
 external getConsoleMode : unit -> int = "win_get_console_mode"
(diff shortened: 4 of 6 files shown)


gdt:
I would definitely like to hear some other thoughts on this.

It strikes me that the the current code is doing something basically wrong with ctime, and should arguably be returning some NULL sort of value (encoded how?) and that should be handled as "no ctime is available".

(I am not a fan of using interfaces without headers and libraries to link with -- but I guess that's just how the MS world is.)

tleedjarv:
> It strikes me that the the current code is doing something basically wrong with ctime

I guess this also is just how the MS world is. It's not that Unison code is doing something unique here. Only software that has gone the extra length (like using the interface without headers and libraries to link with, in this case) can do the right thing. And even then, ctime (proper) will not be available on FAT, for example.

gdt:
I'm not averse; just would like to understand.   The call for others to review has timed out, so I mostly read the diff.

 * How is ctime used in unison?
 * Will merging this cause replicas that used to sync as not needing changes to appear to have changes?
 * What problem in terms of user experience/spec compliance is this fixing?

tleedjarv:
I think we can park this one for a while... It would be good if someone who knows more about Windows, Win32 API and all their bugs and caveats reviewed all Windows-specific patches.

* ctime is currently not used if I remember correctly. I'v added a commit in the ACL branch to use ctime for quickly detecting status-only changes. Without extended props, such as ACL and xattrs, it may not actually speed up updates scanning (but platforms and systems are so different, so hard to tell).
* Merging this by itself should not make any new changes (real or not) appear as ctime itself is not synced nor used for updates detection right now.
* Several comments on the last bullet:
   * first, current `stat` implementation on Windows reports as ctime the creation time. This is so far off incorrect that even reporting mtime would be a huge improvement. This patch would make `stat` on Windows more POSIX compliant.
   * second, although more proof is needed, it _may_ speed up updates detection (given the patch from ACL branch is merged, that is) which on Windows is probably the slowest of all platforms.
   * yet another thing, which I have no proof for, have not tested and have no use for myself, is that using proper ctime may actually correct an update detection issue on Windows where some files are updated without mtime being updated (reading in the current code, I think Excel and MS Project are the usual suspects here -- I think there was also an open GH issue about this).

So, all in all, there is no direct need to merge it by itself. It makes only sense if there is also some other code that will make use of having correct ctime available.

gdt:
Based on your explanation, this PR is correcting a bug that the ctime field of the (unison-provided?) stat function is just wrong on windows, and having it more accurate seems like a good thing for the future, even if nothing uses it now.

So that makes me inclined to merge it, as an actual fix for a latent bug, with no immediate impact on behavior.   We've given more than ample opportunity for those theoretical windows-clueful unison reviewers to comment; it's been 5 months since my call for review.

Does merging now therefore sound reasonable?

tleedjarv:
> the (unison-provided?) stat function

Yes. OCaml's Unix library historically has not had the strongest Windows support. It started getting better only a few years ago when some people really started putting in focused effort. Unison has local implementations of many Win32 API stubs for this reason. I think most of those local implementations can be removed when using OCaml >= 4.06 (so we're almost there!).
(This is my understanding; I've really only known OCaml for less than a year, when I first started hacking on Unison.)

I've filed a bug for this same issue on OCaml's `Unix.stat` but I don't expect it to be resolved any time soon (although there was some early tenative feedback that it's a good idea).

I think merging is safe but obviously my testing is very limited.

gdt:
If Windows people don't review and test we just have to do the best we can.   So I'm hitting merge, and we'll fix it if issues -- we are not near a release this minute.  Thanks for all the explanations.

PR #478, opened 2021-02-28
Update update.ml

zzkt:
add detail to "Archive file seems damaged" error message


tleedjarv (approved):
Looks good, other than the line becoming too long (which to me is not a major issue).

zzkt:
Thanks. I couldn't see an obvious way to keep it roughly the same length....

PR #480, opened 2021-02-28
OCaml 4.12 compatibility: use new hash function

tleedjarv:
**Important!** This is an incompatible change. Another PR with temporary compatibility shim is in #481. Once the current PR is ready to be merged then the compatibility shim must be reverted.

OCaml 3.x hash function has been removed since OCaml 4.12. The only reason to have kept using the old hash function was Unison backwards compatibility.


tleedjarv:
Note to reviewers: an alternative way is simpler.
(this is the entire change)
``` diff
diff --git a/src/uutil.ml b/src/uutil.ml
index fb2bb75..0039eae 100644
--- a/src/uutil.ml
+++ b/src/uutil.ml
@@ -34,9 +34,7 @@ let myNameAndVersion = myName ^ " " ^ myVersion
 
 let hash2 x y = (17 * x + 257 * y) land 0x3FFFFFFF
 
-external hash_param : int -> int -> 'a -> int = "caml_hash_univ_param" [@@noalloc]
-
-let hash x = hash_param 10 100 x
+let hash = Hashtbl.hash
 
 (*****************************************************************************)
 (*                             File sizes                                    *)
```
I originally did a slight variation of this and dismissed it. But now I'm thinking this is the preferred way.

tleedjarv:
Rebased, now that #481 is merged. Added a commit to revert the temporary fix and another commit with a permanent fix.

I've now settled for the simplest fix. For the record, for future reviewers, I initially did a partial revert of 15c73c5dec3b5432e355e4692436b88337bd33a8 but dropped that idea because it touches the code everywhere the hash function is used and that is just unnecessary. Changing `Uutil.hash` is only one change and ensures that the same hash function is used everywhere.

gdt:
I see this as tied up with the format fixes, because using ocaml-defined hashes feels unstable and something I think we should be trying to get away from.   Ideally it should be reasonable to implement unison in another language from the protocol spec.   Not going to happen I know, but I see it as a bug if that's not doable given effort.    But all of this will need to be discussed another day anyway.  In any case, thanks for adapting to 4.12 without a protocol break for the upcoming release - that's hugely helpful.

tleedjarv:
> I see this as tied up with the format fixes, because using ocaml-defined hashes feels unstable and something I think we should be trying to get away from.

Yes, agree. OCaml implementation is based on MurmurHash3 but I'm not entirely sure what does "based on" mean exactly. Perhaps it only means that the result is forced into range [0, 2^30-1] (result & 0x3FFFFFFFU). But it's still true that it remains completely tied to the OCaml version and long term that's not an ideal solution. Format fixes should replace this hash function as well.

tleedjarv:
As much as I don't like the solution I called temporary, given that we now have very good chances of being backwards compatible for a long time still, that temporary solution is also going to be with us for a long time still. As we don't want to introduce another shaky dependency, I'm closing this PR. (Also, there is a non-breaking way now with the proposed "features" mechanism.)

PR #482, opened 2021-02-28
GHA: Add OCaml 4.12.0, replace 4.11.1 -> 4.11.2

tleedjarv:
Build on 4.12.0 will succeed once either #480 or #481 is merged.


gdt:
Yes, I know that now 4.12 will fail.

Issue #483, opened 2021-03-02
No -servercmd equivalent for unison-fsmonitor executable?


GrqPnegre:
I've build unison (native=true) on a server where I don't have permission configure the path for SSH, so I'm needing to specify `servercmd=/path/to/unison` to locate the unison executable.  

When I run with `-repeat watch` I get the following error:
```bash
0 $ unison default a.tmp ssh://myserver//home/user/a.tmp -repeat watch -times -logfile /tmp/unison.log
Unison 2.51.3 (ocaml 4.08.1): Contacting server...
---8<---
Fatal error: No file monitoring helper program found
```

If I had to guess, it looks like the fsmonitor isn't found on the SSH path (which will be for a non-interactive shell).  I've tried numerous things to work around this problem with no luck.  Help!

Is there a missing command like `-fsmonitorcmd`?

-G

tleedjarv:
I created PR #484 to add more debugging information. Try a build from that PR (you can download the build artifacts from the 'Checks' section). Just run as you would normally and add `-debug fswatch`.

GrqPnegre:
Thx.

Now I'm questioning whether I'm wrong about how the monitor is invoked.  I'm running the above command on macOS where now I realized the build didn't produce `unison-fsmonitor`.  I built `unison` via a custom macport portfile to get the right versions between my boxes and didn't take it from the port repo.

My assumption was that the monitor wasn't found on the remote, but it's probably the local machine.  Actually, does `-repeat watch` execute the monitor on the remote side as well?

tleedjarv:
Yes, the monitor is executed on both hosts. I think on macOS the monitor is a Python script, not a native binary. So you have to make sure you have `fsmonitor.py` and whatever libraries it requires and that it actually works with your Python version. The added debugging information in the PR I mentioned is only for the binary fsmonitor.

gdt:
This seems like a question, not a bug report, and after discussion I see no remaining issue.

PR #484, opened 2021-03-02
Add more debugging during the search for fsmonitor helper program

tleedjarv:


gdt:
Is the notion that this will not change behavior if debugging is not enabled, and thus this should be merged?  Or that this is debugging branch to help someone and would be too noisy for normal use?

tleedjarv:
Yes, it will only add output if debugging is enabled. There are no other functional changes, debugging enabled or not. It is not too noisy for normal use. With debugging, it will list all the paths and file names where the fsmonitor helper is being searched for.

Its usefulness is very limited, basically the class of issues similar to #483. I don't think these are frequent issues, so I don't mind just closing this without merging.

gdt:
I suspect someone else will have this problem anyway.

PR #485, opened 2021-03-05
Add a proper opam file and add support for alternative libc

kit-ty-kate:
Adding a proper opam file allows to use `opam pin` on this repository and well as fix https://github.com/ocaml/opam/issues/4584
`glibc` 2.4 was released in 2006, the `inotify_compat.h` is not present in the repository and alpine/musl has this include file.

### DIFF
--- a/unison.opam
+++ b/unison.opam
@@ -1,0 +1,23 @@
+opam-version: "2.0"
+maintainer: "juergen@hoetzel.info"
+authors: [
+  "Jürgen Hötzel <juergen@hoetzel.info>"
+  "Louis Gesbert <louis.gesbert@ocamlpro.com>"
+]
+homepage: "https://www.cis.upenn.edu/~bcpierce/unison/"
+bug-reports: "https://github.com/bcpierce00/unison/issues"
+dev-repo: "git://github.com/bcpierce00/unison.git"
+build: ["dune" "build" "-p" name "-j" jobs]
+depends: [
+  "ocaml" {>= "4.03" & < "4.12"}
+  "dune" {>= "1.3"}
+  "lablgtk" {>= "2.18.6"}
+]
+available: os-family != "alpine" # uses glibc-only macros
+synopsis: "File-synchronization tool for Unix and Windows"
+description: """
+Unison is a file-synchronization tool for Unix and Windows. It allows
+two replicas of a collection of files and directories to be stored on
+different hosts (or different disks on the same host), modified
+separately, and then brought up to date by propagating the changes in
+each replica to the other."""
--- a/unison.opam
+++ b/unison.opam
@@ -13,7 +13,6 @@ depends: [
   "dune" {>= "1.3"}
   "lablgtk" {>= "2.18.6"}
 ]
-available: os-family != "alpine" # uses glibc-only macros
 synopsis: "File-synchronization tool for Unix and Windows"
 description: """
 Unison is a file-synchronization tool for Unix and Windows. It allows
--- a/src/Makefile.OCaml
+++ b/src/Makefile.OCaml
@@ -286,12 +286,8 @@ endif
 ########################################################################
 ### Filesystem monitoring
 
-OCAMLSWITCH_IS_MUSL := $(shell opam switch show 2> /dev/null | grep -iF '+musl')
-
 ifeq ($(OSARCH),Linux)
-ifeq ($(OCAMLSWITCH_IS_MUSL),)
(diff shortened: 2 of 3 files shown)


unison.opam
```diff
@@ -0,0 +1,23 @@
+opam-version: "2.0"
+maintainer: "juergen@hoetzel.info"
+authors: [
+  "Jürgen Hötzel <juergen@hoetzel.info>"
+  "Louis Gesbert <louis.gesbert@ocamlpro.com>"
+]
+homepage: "https://www.cis.upenn.edu/~bcpierce/unison/"
+bug-reports: "mailto:unison-users@yahoogroups.com "
```
gdt, line 8: yahoogroups has shut down.  see https://github.com/bcpierce00/unison/wiki/Mailing-Lists

gdt (commented):
Thanks.   The mailinglist pointer is out of date.   Other than that I don't see anything problematic, and I've asked a a few others to look at this.

tleedjarv:
Does fsmonitor build with musl? In that case you should revert 4c0b7fb672e6c8e6e62f33d8a5b76fbd847ae97f

unison.opam
kit-ty-kate: I pointed to this repository instead. This is usually what people use for this field anyway.
The opam file I used as a base was the latest available version in opam-repository and it contained this.

kit-ty-kate:
> Does fsmonitor build with musl? In that case you should revert [4c0b7fb](https://github.com/bcpierce00/unison/commit/4c0b7fb672e6c8e6e62f33d8a5b76fbd847ae97f)

From what I'm seeing from my tests, yes. For instance, using the following Dockerfile:
```
FROM ocaml/opam:alpine-ocaml-4.11
RUN sudo ln -f /usr/bin/opam-2.1 /usr/bin/opam
RUN opam pin add -yn unison.dev git://github.com/kit-ty-kate/unison.git#opam
RUN opam install --deps-only unison
RUN opam source unison.dev
WORKDIR unison.dev
RUN opam exec -- dune build -p unison --verbose
RUN opam exec -- make
```
both dune and make seem to correctly build the `unison-fsmonitor` executable.

gdt:
I'm in general a fan of keeping separate things separate (e.g. adding opam and fixing things for alpine), but I don't want to torque people around spitting things when they are simple enough not to be confusing to someone later bisecting, theoretically.

So, my sense is that this PR is now mergeable, and it's passed the build checks, and thus there's no reason I shouldn't hit merge.   Is that everyone else's view?

(I don't object to a new "build fsmonitor on alpine" PR with one commit at all.)

kit-ty-kate:
My 3 separate commits should be enough for this purpose as long as the PR is "merged" using the "Rebase and merge" button

kit-ty-kate:
If you prefer to have a merge commit, I can split it in 2 separate PRs

gdt:
I prefer merge commits as I feel it makes the history easier to read, with PR refs.    How about you just fold the fix to the opam file into the opam commit and force-push, and I'll then merge it, not worrying that the two commits  aren't strictly related.

kit-ty-kate:
> I prefer merge commits as I feel it makes the history easier to read, with PR refs. How about you just fold the fix to the opam file into the opam commit and force-push, and I'll then merge it, not worrying that the two commits aren't strictly related.

like this?

tleedjarv:
> So, my sense is that this PR is now mergeable, and it's passed the build checks, and thus there's no reason I shouldn't hit merge. Is that everyone else's view?

I think for this PR to be ready (at least the "add support for alternative libc" part), it must explicitly revert 4c0b7fb672e6c8e6e62f33d8a5b76fbd847ae97f.

gdt:
Yes, thanks - looks good to me.

I don't mind enabling it without fsmonitor as an intermediate step.    So I'm fine with this as is.

However, our CI is broken I think - will bring that up on the list.

kit-ty-kate:
> I think for this PR to be ready (at least the "add support for alternative libc" part), it must explicitly revert [4c0b7fb](https://github.com/bcpierce00/unison/commit/4c0b7fb672e6c8e6e62f33d8a5b76fbd847ae97f).

Oh i see, the +musl variants are deprecated anyway by https://github.com/ocaml/opam-repository/pull/17541

kit-ty-kate:
CI passed correctly on `4.10.0+musl+static+flambda`

tleedjarv:
I can't comment on opam but otherwise looks good!

gdt:
Thanks for your patience with our reviews/etc.

PR #492, opened 2021-03-20
fsmonitor: Add Solaris/illumos support

tleedjarv:
Add fsmonitor support for File Event Notification (FEN) interface provided by Solaris and illumos (OpenIndiana, SmartOS, OmniOS, ...).

Due to the way FEN is implemented, it's memory usage might get quite big when watching a huge number of files. I went through a couple of iterations to reduce memory usage at expense of slightly increasing code complexity (but well worth it with 2x reduced memory usage compared to a straightforward implementation). There are some further reduction opportunities, should they ever be needed. Performance has not been a priority and is not cause for concern, with potential exception of when very large directories are monitored.

Also includes a slight Makefile modification to reduce the number of ocamlc arguments that have to be included on all platforms.

Depends on #488 and #489 (those commits are included in this PR and will be removed before merging). Dependency on #488 is created because I decided to define all events as non-immediate. Truth to be told, I don't understand the purpose of distinguishing immediate and non-immediate events, so this decision may very well be incorrect. Still, I have tested the functionality both ways.

### DIFF
--- a/src/.depend
+++ b/src/.depend
@@ -359,6 +359,16 @@ fsmonitor/linux/watcher.cmx : \
     fsmonitor/linux/lwt_inotify.cmx \
     lwt/lwt.cmx \
     fsmonitor/linux/inotify.cmx
+fsmonitor/solaris/watcher.cmo : \
+    fsmonitor/watchercommon.cmi \
+    ubase/safelist.cmi \
+    lwt/lwt_unix.cmi \
+    lwt/lwt.cmi
+fsmonitor/solaris/watcher.cmx : \
+    fsmonitor/watchercommon.cmx \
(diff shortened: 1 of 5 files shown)


gdt:
I merged the other two, so this needs a rebase.

I have a preference to put the makefile adjustments for DEP_FLAGS in a separate PR, or at least a separate commit.  I don't expect trouble but I think it's good to have separate things separate in the case of future investigation.

Other than that, I don't expect trouble on non-solaris, and this has had way more than long enough for others, so I'm inclined to merge quickly after rebase.

tleedjarv:
Rebased and extracted makefile changes to another PR. This PR has no impact on other platforms.

PR #493, opened 2021-03-20
GHA: Package fsmonitor on Windows

tleedjarv:
`unison-fsmonitor` is currently not packaged on Windows (see #380) because it causes a hang. Exact comment from workflow file:
```
windows gtk2 build hangs at "looking for changes" when `unison-fsmonitor.exe` is present
```

Can anyone actually reproduce this hang? I propose to package `unison-fsmonitor` on Windows unless the hang is confirmed.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -153,10 +153,9 @@ jobs:
       run: |
         opam exec -- make src OSTYPE=$OSTYPE UISTYLE=text STATIC=${{ steps.vars.outputs.STATIC }}
         # stage
-        # * notes: darwin/macos doesn't build `unison-fsmonitor` ; windows gtk2 build hangs at "looking for changes" when `unison-fsmonitor.exe` is present
+        # * notes: darwin/macos doesn't build `unison-fsmonitor`
         for file in ${PROJECT_EXES} ; do
           if [ -f "src/${file}${{ steps.vars.outputs.EXE_suffix }}" ]; then
-            case '${{ matrix.job.os }}' in windows-*) if [ "${file}" = "unison-fsmonitor" ]; then continue; fi ;; esac
             cp "src/${file}${{ steps.vars.outputs.EXE_suffix }}" '${{ steps.vars.outputs.PKG_DIR }}/bin'
             echo "'src/${file}${{ steps.vars.outputs.EXE_suffix }}' copied to '${{ steps.vars.outputs.PKG_DIR }}/bin'"
           fi


jedfrechette:
I'm not sure if it is the same symptom referred to here, but I regularly see version 2.51.3 installed via Chocolatey hang while looking for changes with `unison-fsmonitor.exe` using 100% of one core.

tleedjarv:
It definitely looks to be related. @jedfrechette do you think you can get a stack trace from the `unison-fsmonitor.exe` process while it's stuck using 100% of one core?

jedfrechette:
Sure thing. Next time it happens I'll see what I can get. I seem to recall Process Explorer can get stack trace's but I'm not very familiar with Windows dev tooling so any suggestions on the best way to get what you need would be appreciated.

tleedjarv:
> I seem to recall Process Explorer can get stack trace's but I'm not very familiar with Windows dev tooling so any suggestions on the best way to get what you need would be appreciated.

I don't know about Windows dev tooling either, so I'm not sure what is possible there at all. Anything is better than nothing, let's start with what you can get out of Process Explorer.

jedfrechette:
Here is a stack trace of fsmonitor while it was hung. 

```
ntoskrnl.exe!KeSynchronizeExecution+0x5b66
ntoskrnl.exe!KeWaitForMutexObject+0x1460
ntoskrnl.exe!KeWaitForMutexObject+0x98f
ntoskrnl.exe!KeWaitForMutexObject+0x233
ntoskrnl.exe!ExWaitForRundownProtectionRelease+0x7dd
ntoskrnl.exe!KeWaitForMutexObject+0x3a29
ntoskrnl.exe!KeSynchronizeExecution+0x3140
unison-fsmonitor.exe+0x59eb4
0x0000000000000000
```

The system this was captured on didn't have debugging symbols installed so I'll also see if I can get another one with them.

winny-:
Not sure if this helps, but this does not appear to happen on my `unison version 2.48.4` installation.  I am using the build available from the [2.48.4 GitHub release](https://github.com/bcpierce00/unison/releases/tag/2.48.4).  [Direct link](https://github.com/bcpierce00/unison/releases/download/2.48.4/Unison-Windows-2.48.4.zip).

Using `Microsoft Windows [Version 10.0.19041.928]` (Windows 10 Professional).

gdt:
It looks like the builds from this PR completed.

So I'm guessing we should merge this now, and deal with the problem if it surfaces in GHA.   Do y'all agree?

tleedjarv:
This one is tricky. There are reports of hangs on some builds/machines and there are reports of everything working fine on other machines.
 `watch = true` is a default preference even when the user has no intention of using the fsmonitor process. This makes it a bit dangerous to merge this PR as all users on Windows will then get the fsmonitor and that _may_ cause a hang. That will happen even if they have no interest in using it.

Let's hold off merging this one for a while longer. Perhaps some progress will be made in tracking down the cause of the hang.

Looking at the stack trace provided by @jedfrechette, it is clearly some kind of race condition (waiting for Mutex in Windows kernel). There is no explicit threading or synchronization going on in the fsmonitor code. It may very well be that Win32 API is being used slightly incorrectly, which leads to the race condition. Could also be an OCaml bug.

gdt:
I intend to merge this after merging #528, as then people who don't try to use fsmonitor won't use it.

PR #501, opened 2021-04-15
remove unused THREADS variable from Makefile

olafhering:
Fixes commit b043e4c394834fff541bffa2f2e6b4a6980fcf86

Signed-off-by: Olaf Hering <olaf@aepfle.de>

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -7,16 +7,7 @@
 
 # Set NATIVE=false if you are not using the native code compiler (ocamlopt)
 # This is not advised, though: Unison runs much slower when byte-compiled.
-#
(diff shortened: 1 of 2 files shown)


tleedjarv (approved):
Looks good to me but I don't understand the commit message (perhaps drop the "fixes commit ..." part?).

olafhering:
The referenced commit introduces the issue, this commit fixes the issue.

tleedjarv:
> The referenced commit introduces the issue, this commit fixes the issue.

That's very misleading. This is just not the case here (it's not a "fix" and it definitely does not fix an initial commit that imported the entire codebase). Please remove that statement from the commit message.

PR #515, opened 2021-05-14
Some tidying in TODO.txt

tleedjarv:
The first commit removes items that by my judgement are obsolete. Add a comment if you think something should not be removed or if you think even more items should be removed.

The second commit moves some items around to group similar ones together. There are no changes, no additions.

### DIFF
--- a/src/TODO.txt
+++ b/src/TODO.txt
@@ -16,14 +16,6 @@ promise that anybody is going to implement it!)
      - backup
      - fs watcher command
 
-* Merge issues:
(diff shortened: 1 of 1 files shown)


gdt:
Looks ok to me, and given that I am not seeing review comments from others on substantive things, I am inclined not to have a long timeout on things like this.    If anyone is interested in reviewing and would benefit from a week or so vs 24h, please feel free to speak up - this is meant simply to avoid waiting for a theoretical reviewer which is actually zero people, rather than to exclude anyone who wishes to start.

PR #527, opened 2021-06-05
Have consistent copyright in all files

tleedjarv:


tleedjarv:
Closing after further discussion. Going forward, this is not the way.

PR #528, opened 2021-06-07
Default "watch" preference to false

tleedjarv:
As proposed in the mailing list.


gdt:
Will merge after 18Z on 10 June, barring rationale why not.

tleedjarv:
Once this is merged, #493 becomes safe to merge.

PR #530, opened 2021-06-10
fsmonitor: Specify IN_EXCL_UNLINK for inotify

tleedjarv:
From `inotify(7)`:
```
Specifying IN_EXCL_UNLINK changes the default behavior, so that events
are not generated for children after they have been unlinked from the
watched directory.
```


gdt:
Is inotify specified anywhere?   Is there a reason to expect all systems which have an implementation to behave the same way?

Issue #538, opened 2021-06-15
unison.2.51.4~rc3 fails to build using opam


kit-ty-kate:
Detected in https://github.com/ocaml/opam-repository/pull/18861
```
#=== ERROR while compiling unison.2.51.4~rc3 ==================================#
# context              2.0.8 | linux/x86_64 | ocaml-base-compiler.4.05.0 | pinned(https://github.com/bcpierce00/unison/archive/refs/tags/v2.51.4_rc3.tar.gz)
# path                 ~/.opam/4.05/.opam-switch/build/unison.2.51.4~rc3
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p unison -j 31
# exit-code            1
# env-file             ~/.opam/log/unison-21485-0a5907.env
# output-file          ~/.opam/log/unison-21485-0a5907.out
### output ###
#     ocamlopt src/linkgtk2.exe (exit 2)
# (cd _build/default && /home/opam/.opam/4.05/bin/ocamlopt.opt -w -40 -w -3-6-9-27-32-52 -g -o src/linkgtk2.exe /home/opam/.opam/4.05/lib/ocaml/unix.cmxa -I /home/opam/.opam/4.05/lib/ocaml /home/opam/.opam/4.05/lib/ocaml/threads/threads.cmxa -I /home/opam/.opam/4.05/lib/ocaml /home/opam/.opam/4.05/lib/ocaml/str.cmxa -I /home/opam/.opam/4.05/lib/ocaml src/lwt/lwt_lib.cmxa /home/opam/.opam/4.05/lib/ocaml/bigarray.cmxa -I /home/opam/.opam/4.05/lib/ocaml src/unison_lib.cmxa -I src /home/opam/.opam/4.05/lib/lablgtk2/lablgtk.cmxa /home/opam/.opam/4.05/lib/lablgtk2/gtkThread.cmx -I /home/opam/.opam/4.05/lib/lablgtk2 src/.linkgtk2.eobjs/native/uigtk2.cmx src/.linkgtk2.eobjs/native/linkgtk2.cmx)
# /usr/bin/ld: src/unison_lib.a(uutil.o): in function `camlUutil__hash_1212':
# /home/opam/.opam/4.05/.opam-switch/build/unison.2.51.4~rc3/_build/default/src/uutil.ml:39: undefined reference to `unsn_hash_univ_param'
# /usr/bin/ld: src/unison_lib.a(uutil.o): in function `camlUutil__43':
[...]

PR #540, opened 2021-06-15
GHA: Add a dune build via opam

tleedjarv:
To avoid future embarrasments like #538.

Some notes and caveats:
1. This GHA job is created specifically to test dune builds. Building with dune is actually not required by `opam` -- it is ok to instruct opam to build with `make`, and this may be just what we have to do if nobody steps up to maintain dunefiles.
2. Unison's dunefiles are a bit outdated and currently unmaintained. They need an active maintainer who knows dune.
3. Self-tests are currently not working with dune builds. (Current dunefile explicitly excludes self-tests. I tried changing the dunefile to include the `Test` module in the 'library' or in an 'executable' but that didn't cut it. No idea how to get this working with dune.)

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -237,3 +237,27 @@ jobs:
         files: ${{ steps.macapp.outputs.APP_NAME }}
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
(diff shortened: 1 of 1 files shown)


gdt:
Given that this passes, it seems better to have it than not have it, so I'm going to merge and if anyone has improvements they are of course welcome.

PR #557, opened 2021-06-25
uigtk2: Replace GtkCList with GtkTreeView and GtkListStore

tleedjarv:
Based on a report in the mailing list I decided to take a look at the GTK GUI performance.

I don't think it is common to run the GUI with tens of thousands or hundreds of thousands of sync differences but there are some relatively low hanging fruits here to improve the performance even in these rare cases.

There was one small fix in #555 which only improved the speed of population of the display list. This PR drops the previously used (and deprecated) GtkCList completely and replaces it with GtkTreeView with GtkListStore backing model.

The idea is that functionality does not change. This change should only bring with it considerably improved performance. Unfortunately, the population of the list is slightly slower. In return, once the list is populated, all further actions are much faster. Next step, to extract even more performance, would be to create a custom tree view model to use the internal state array directly as GtkTreeView backing model. I don't think this is worth the effort, though. Already now, some actions are down from taking several minutes to taking less than half a minute.

**Remaining issues**

I have no experience with GTK at all. This was mostly a mechanical change where not much GTK knowledge was needed. There are a few quirks remaining which require a better understanding of GTK. Help is welcome with these issues.

~~First, there seems to be a 'focus row' separate from the selected row. This means that when programmatically changing the selected rows (move to next row, for example), the 'focus row' does not move. If the user now uses keyboard to move up or down the list, the selection will jump to where the focus row was, not where the visible selection was.~~
Also, multiple select by just dragging with mouse over the rows does not work any more. I don't know if this is possible with GtkTreeView.

Second, the status column displays either an icon (success, failed) or text (percentage progress) but not both at the same time. With GtkCList the content was nicely centered. With GtkTreeView the icon and text are placed next to each other. Even though only one of them is displayed at any time, it will not be centered. The content is still correct but just looks visually off.

### DIFF
--- a/src/uigtk2.ml
+++ b/src/uigtk2.ml
@@ -128,13 +128,6 @@ type stateItem = { mutable ri : reconItem;
 let theState = ref [||]
 let unsynchronizedPaths = ref None
 
-module IntSet = Set.Make (struct type t = int let compare = compare end)
-
-let current = ref IntSet.empty
-
-let currentRow () =
-  if IntSet.cardinal !current = 1 then Some (IntSet.choose !current) else None
-
 (* ---- *)
 
 let theToplevelWindow = ref None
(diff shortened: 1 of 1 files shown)


gdt:
I am not averse to improvements to the GTK support, but
  * Given that this is improving speed of something that is not broken, I would really like to get a confirming report that the new scheme works, and code review by somebody who knows gtk
  * I think we really need to consider moving to gtk3 somehow, rather than any greater-than-very-small work on gtk2.  gtk2 is just not an ok even medium-term strategy at this point.

acolomb:
I don't know ocaml at all, but might find some time to look at this from a Gtk perspective, as a happy daily unison-gtk user.

Switching to GtkTreeView as a first step is not bad at all, as it reduces the changes necessary when going to Gtk3. Coupling those two will make it even harder and less likely that the Gtk3 port should ever materialize.

gdt:
@acolomb Thanks.  I didn't mean that I didn't want to merge this; it seems small and sensible.  Knowing that it's an incremental step to gtk3 is good.    I am certainly a fan of multiple smaller things vs one huge change.   Even if you don't have time to do it all, if you are able to sketch out a path to switching to GTK3, please add that to https://github.com/bcpierce00/unison/wiki/GTK3-migration-plan

acolomb:
> @acolomb Thanks. I didn't mean that I didn't want to merge this; it seems small and sensible. Knowing that it's an incremental step to gtk3 is good. I am certainly a fan of multiple smaller things vs one huge change. Even if you don't have time to do it all, if you are able to sketch out a path to switching to GTK3, please add that to https://github.com/bcpierce00/unison/wiki/GTK3-migration-plan

Sorry for not responding here in time.  Was simply too busy to even take a look at the changes.  Now I briefly skimmed the diff, but frankly the OCaml GTK bindings look very different from any other GTK usage I've previously looked at.  So I'm not sure I can be of much help with judging Unison's GTK code, sorry.

PR #559, opened 2021-06-25
Fix error handling for log file related errors

tleedjarv:
There was a report in the mailing list
```
THere is one change that didn't go down well: The default file location
has changed from ~ to ~/.unison.

Now, I had already
        logfile = .unison/unison.log
in my config files.

That crashed unison with the error message that there is no such
directory
        ~/.unison/.unison/unison.log

On the TUI this is easy to see, **but** the GUI does not realize the
breakage and hangs forever not reacting at all.
```

This PR has two fixes. First, if the dir tree to log file does not exist then it is (attempted to be) created. Second, if opening or writing to log file fails, don't freak out about it in TUI and GUI error handlers.

### DIFF
--- a/src/ubase/trace.ml
+++ b/src/ubase/trace.ml
@@ -244,6 +244,12 @@ let log s = displayMessage (Log, s)
 
 let log_color s = displayMessage (LogColor, s)
 
+let logonly s =
+  let temp = !sendLogMsgsToStderr in
+  sendLogMsgsToStderr := false;
+  displayMessage (Log, s);
+  sendLogMsgsToStderr := temp
+
 let logverbose s =
   let temp = !sendLogMsgsToStderr in
   sendLogMsgsToStderr := !sendLogMsgsToStderr && not (Prefs.read terse);
--- a/src/ubase/trace.mli
+++ b/src/ubase/trace.mli
@@ -93,6 +93,10 @@ val log : string -> unit
    ANSI color escapes intact *)
 val log_color : string -> unit
 
+(* Like 'log', but only send message to log file and not to stderr, even when
+   [sendLogMsgsToStderr] is set to true *)
+val logonly : string -> unit
+
 (* Like 'log', but only send message to log file if -terse preference is set *)
 val logverbose : string -> unit
 
--- a/src/uicommon.ml
+++ b/src/uicommon.ml
@@ -670,6 +670,10 @@ let initPrefs ~profileName ~displayWaitMessage ~getFirstRoot ~getSecondRoot
   (* Print the preference settings *)
   debug (fun() -> Prefs.dumpPrefsToStderr() );
 
+  Trace.logonly (Printf.sprintf "\n%s log started at %s\n\n"
+    (String.capitalize_ascii Uutil.myNameAndVersion)
+    (Util.time2string (Unix.gettimeofday ())));
+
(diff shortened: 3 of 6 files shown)


gdt:
(with hat of authority off)  I think I am -1 on creating the directory.   I more or less feel that it's 50/50 whether a missing directory is because the user meant to make the directory and didn't or the path is wrong, and I lean to less magic and clear errors.

tleedjarv:
I'm 50/50 on whether it's a good idea to create the directory or not. Do you know if there is some common practice?

Not creating the directory will actually require bigger changes. It may happen in practice that the first time the log file is opened is quite late in the sync process. Even though the error reporting itself is fixed by this PR, reporting such an error late in the sync process is rather ridiculous. The log file would have to be opened/created very early at startup. I may look into this at some point.

bcpierce00:
Why would it be hard to open/create the file early in the process?

tleedjarv:
Not hard, just additional work. I think it's two changes. First to log some startup notice immediately after prefs init to trigger log file opening. And then to get some meaningful error reporting if unable to create the log file. The latter is not strictly needed because the 'raw' error returned by syscall will be properly reported in UI anyway (I think).

gdt:
I don't think there is a requirement to error early.  To me, if the log is missing, that's a not a big deal, and what I'd like to avoid is to have log silently not get created repeatedly with the user having no idea.   So if there is an error printed the way other errors  are printed once per invocation, when the first attempt is made to write to the log, that's fine.

Also, I wonder if we can just do a log of "Unison 2.51.4 (ocaml 4.11) starting" at the beginning to provoke it early.

I am not really aware of creating directories for logs.   To me that is pretty irregular and I expect to have to create the dir and have it have the right mode ahead of time.    But that may be a windows culture vs unix culture thing.

tleedjarv:
I don't have a personal stake in this so I'm trying to think what would be the best for users and it's hard to tell. To make the decision easy, I will consider that those impacted are only those who set the `logfile` preference. This means it is users who are already tweaking with the setup beyond the syncing conf and therefore it's ok for them to create the dirs on their own. So we are getting towards a clean consensus.

If that's ok with you as well @bcpierce00 then I'll go ahead and change the PR.

> So if there is an error printed the way other errors are printed once per invocation, when the first attempt is made to write to the log, that's fine.

Except, that's not the case right now. Logfile error is a fatal error. I think it would be easy to change it to a transient error but I don't know how that would behave. Probably print an error at every logging attempt. I'll just start with the approach of trying to log the startup notice if nothing better comes up meanwhile.

bcpierce00:
It feels like the simplest design is
* attempt to write the log file near the beginning of the run
* fatal error if this fails for any reason (including directory not existing)

tleedjarv:
Patch updated according to discussion.

gdt:
It looks to me like there is a commit and a revert, and a third commit that does what we want.  If that's true, could you rebase to just the intended commit?  And if I'm confused, explain?

(Agreed that we have consensus on the 1) don't create dirs and 2) error out really early if log can't be opened approach

tleedjarv:
Rebased, squashed the first two commits, fixed the fix a little (don't catch quite _everything_, just `Util.Fatal`) and added the fix for uimac, too.

Should be good to go now.

PR #563, opened 2021-07-04
Listen on multiple sockets when needed

### LINKED ISSUE #562: "unison -socket N" listens only on IPv6
With v2.51.4_rc2, unison -socket creates a TCP socket of type IPv6.   While some operating systems support mapped addresses and hence can accept IPv4 connections, not all do.  On NetBSD 9, trying to connect to 127.0.0.1:N, unsurprisingly, gets a connection refused.

The normal practice is to open two sockets, one on each address family, and put them both in the select loop.  I am not clear if the server assumes there is only one connection; hence the effort-high tag.

tleedjarv:
See #562 for background.

This is the first attempt at it, and may be sufficient. Listen and accept on multiple sockets if multiple addresses are returned by `getaddrinfo`.
There are some indentation changes, `diff -w` may help when reviewing.

Fixes #562

### DIFF
--- a/src/remote.ml
+++ b/src/remote.ml
@@ -1022,7 +1022,7 @@ let printAddr host addr =
   | Unix.ADDR_INET (s, p) ->
       Format.sprintf "%s[%s]:%d" host (Unix.string_of_inet_addr s) p
 
-let buildSocket host port kind =
+let buildSocket host port kind ai =
   let attemptCreation ai =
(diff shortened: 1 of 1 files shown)


gdt:
As usual, I would really love to see a code review or test report (from other than the submitter, who I am confident has tested).  But will time out on that and merge anyway after a while if not.

Issue #472, opened 2021-01-28
When moving a file, sync on two local replicas does a copy from replica1 rather than from the file in replica2


davidde:
I'm mostly drawn to unison because of its wonderful feature to keep 2 directories in sync without excessively copying and deleting. In other words, because it can be smart about file moves.

After some testing, it appears this smart behavior is only available when I specify at least 1 remote replica, even for syncing 2 local directories. For example:
```
unison -auto /home/david/Documents ssh://localhost//backup/Documents
```
Obviously, this requires an SSH daemon to be running, which is not really desirable if you don't need it. It is also kind of obscure to track down why the "smart moving" feature doesn't work by default, and not very intuitive for new users.

I don't know much about OCaml, or how this is implemented internally, so there may be a valid reason for this behavior. But if there is no specific blocker, wouldn't it make more sense to drop the SSH requirement?

gdt:
I think this is not about ssh but about some notion of local and remote.   But if you can find out details from reading the code, please update the issue.

davidde:
I may not have been clear enough about what I meant.

Continuous sync *does* work on 2 local replicas. It's just not very efficient because it will delete moved files first and then copy them again. I mention SSH because it works as it should when using SSH. That is, it moves the files in the destination directly, without any deleting/copying. This makes a huge difference when backing up large files/directories, as well as reducing wear on the disks.

I'm not sure the current title reflects what I'm trying to express.

davidde:
Also I'm not sure it should be tagged `effort-high` without verifying it's not a simple oversight in the code; after all, the functionality is already there.

What I'm basically asking is for the command:
```
unison -auto /home/david/Documents /backup/Documents
```
(which works, but just is not "smart" about file moves), to work identical as the command:
```
unison -auto /home/david/Documents ssh://localhost//backup/Documents
```
Which **is** smart about file moves. The big benefit then is that users are no longer required to run sshd for no reason at all.

tleedjarv:
@davidde Can you write up a repeatable test case that demonstrates the issue (is it just that a simple `mv` will result in delete + copy in the other replica?). I will have a look at it.

gdt:
[...]

PR #571, opened 2021-08-02
Resume partial transfers for local syncs

### LINKED ISSUE #570: Local sync does not resume partial transfer
Hey,

I seem to have tripped over some odd opinions that Unison has about behaviour of remote vs local syncing!

To elaborate, I want to sync files to and from a local FUSE mount; the code behind this mount then transparently encrypts them and sends them to a cloud store. However Unison appears to unexpectedly have separate code paths for remote and ostensibly "local" syncs:

* Resuming partial transfers works flawlessly remotely, but mostly does not work locally:
    * When the base directory does not yet exist at the target, it restarts the entire directory.
    * When the base directory already exists at the target, it recognises completed files and folders one level down, but restarts any of these that were partially transferred.
* Hitting `ctrl-c` (SIGINT, I think?) on the terminal cancels the whole process when remote syncing, but when local syncing just stops the current file and moves on to the next, leaving partial files.

I think I see in the tracker others stumbling into similar differences (e.g. #472), though not these two specifically.

If fixing them would be very difficult, I am satisfied it's possible to work around by making Unison SSH to my local machine (thus forcing "remote" behaviour). However the "local" path appears to contradict what's in the manual (which is maybe written from a "remote-only" perspective?), so perhaps it could be handy to have a prominent warning somewhere about it.

tleedjarv:
Closes #570

I've opened the PR as draft not because the code isn't finished but because I think more discussion about the (potential) cons of this change may be needed.

While the resume logic exists for remote transfers, it has not been used for local transfers so far. Maybe there is (or was historically) a good reason for doing so. Resuming partials does require digesting the entire contents of partial transfers and perhaps this was seen as unnecessarily expensive locally where a re-copy would be just as fast or even faster. However, I don't believe that is an accurate reason. There are many local sync scenarios where resuming could make sense: different disks, network shares, local "cloud" mounts as mentioned in #570, to name a few.

Usually a local transfer would happen like this (please correct me if I got it wrong):
- **read** the entire source file to fingerprint it,
- copy the file (once more **read** the entire file and **write** it),
- **read** the entire target file to fingerprint it and compare fingerprints (so-called paranoid check).

With this patch a local transfer will happen like this:
- **read** the entire source file to fingerprint it,
- **read** the entire partial target file to fingerprint it, compare fingerprints and discover they are the same - no copy needed,
- **read** the entire target file to fingerprint it and compare fingerprints (so-called paranoid check).

In sum, the total number of reads remains the same and the write is skipped completely. Already a win! But, one read is now done in the target replica instead of the source replica. This could be a win or a lose. A read instead of a write (assuming same file size) in most cases (thinking spinning disks, over network) would probably be faster or pretty much the same. All in all, seems like a net benefit or neutral.

Are there some obvious drawbacks that I'm missing; why was this functionality not enabled for local syncs?

### DIFF
--- a/src/copy.ml
+++ b/src/copy.ml
@@ -317,6 +317,26 @@ let localFile
           fspathFrom pathFrom fspathTo pathTo `RESS ressLength ido;
       setFileinfo fspathTo pathTo realPathTo update desc)
 
+let localFileWithResume
+      fspathFrom pathFrom fspathTo pathTo realPathTo update desc fp ress ido =
+  let (_, isTransferred) = fileIsTransferred fspathTo pathTo desc fp ress in
+  if isTransferred then begin
+    (* File is already fully transferred (from some interrupted
+       previous transfer).  So just make sure permissions are right. *)
+    Trace.log
+      (Printf.sprintf "%s/%s has already been transferred\n"
+        (Fspath.toDebugString fspathTo) (Path.toString realPathTo));
+    setFileinfo fspathTo pathTo realPathTo update desc;
+    match ido with
+    | None -> ()
+    | Some id ->
+        let len = Uutil.Filesize.add (Props.length desc) (Osx.ressLength ress) in
+        Uutil.showProgress id len "alr"
+  end else
+    localFile
+      fspathFrom pathFrom fspathTo pathTo realPathTo
+      update desc (Osx.ressLength ress) ido
+
 (****)
 
 let tryCopyMovedFile fspathTo pathTo realPathTo update desc fp ress id =
@@ -959,9 +979,9 @@ let file rootFrom pathFrom rootTo fspathTo pathTo realPathTo
   let timer = Trace.startTimer "Transmitting file" in
   begin match rootFrom, rootTo with
     (Common.Local, fspathFrom), (Common.Local, realFspathTo) ->
-      localFile
+      localFileWithResume
         fspathFrom pathFrom fspathTo pathTo realPathTo
-        update desc (Osx.ressLength ress) (Some id);
+        update desc fp ress (Some id);
         paranoidCheck fspathTo pathTo realPathTo desc fp ress
   | _ ->
       transferFile


gdt:
I think the idea that "local" implies two fast disks (at least spinning fast) is incorrect.  Many things can mount in via NFS, FUSE or some other mechanism.   So I don't see justifcation for different processing.  A good question to ask, but I tend towards just believing that this used to be a good idea, and after asking and not hearing, not being overly deferential to the past.

It's not clear to me that this PR makes the local sync case behave the same as the remote case, and if there is any remaining behavior difference, what the rationale is for that.

bcpierce00:
I'm afraid I don't remember the reasons why local sync was treated
differently -- I incline toward Greg's theory that it was mostly because
the local case was assumed to be fast, so not needed the extra mechanism.
In any case, moving to uniform treatment seems like a good idea (with the
caveat that this may involve touching some fairly convoluted code, so
plenty of testing is in order).

>

tleedjarv:
> It's not clear to me that this PR makes the local sync case behave the same as the remote case, and if there is any remaining behavior difference, what the rationale is for that.

No, it's still not exactly the same as the remote case. Remote case goes through more steps, very roughly so (not intended to be accurate):
- resume check (same as for local with this patch)
- replica-local copy shortcut (one would _think_ it makes no sense with local sync, but it's not so straightforward; this was discussed in #472)
- use external `copyprog` if requested
- apparently there's one more resume check, which is not included in this PR. It checks if a file was partially transferred and appends the remaining contents. This one I'll definitely look into adding for the local case as well.
- contents are transferred with rsync algorithm
- finally the paranoid check (same as with local sync)

It is not completely clear to me if it is trivial to make the local sync behave the same as the remote case and whether it is something to desire (the rsync thing does sound nice in theory).

gdt:
This may be unreasonable, but I wonder about making all syncs just use the remote path, just skipping ssh/socket and having a 2nd unison locally.   But making local more like remote seems good.

tleedjarv:
I went ahead and dug a little more in the code. With some hacking in the Remote module I got the remote copy code working with local sync and it seems to Just Work (is it all optimal though).

The reason I had to hack in the Remote module is that the interface Copy module is using requires a connection. The connection is actually not needed, as all other remote function calls have a proxy that just calls the function in the same process locally. Adding this proxy worked (seemingly) perfectly (and why shouldn't it have -- all other modules are using this proxied Remote interface which doesn't require a connection).

What I think I'll do next, I add the partial file resume (append) functionality for local syncs in this PR as that seems useful.
But... This functionality, too, is currently dependent on the Remote module interface without the proxy code, so maybe not worth hacking at the moment.

As a next, bigger step, I will look into properly changing the remote function call proxies in the Copy module (as mentioned, I actually did this already; the emphasis here is on **_properly_**). This will allow the entire remote code path to be used locally. It should require very little code change but a whole lot of testing. And then some more testing. Are we ready for that?

gdt:
My overall reaction is that fixing this right, even if a bigger step, is preferred, and that it's less total work to test a major rework once than to test a bunch of small steps.     (And, it's just local sync, and people can always function via ssh to self if things get really troubled.  That's assuming something sneaks by after we have the usual 30 test volunteers :-( )

tleedjarv:
That bigger change is now #572.
I will keep this PR still open for a while but if it seems that the other one is The Right Thing to do then I will close this one.

PR #572, opened 2021-08-02
Use the same code path for local and remote syncs

### LINKED ISSUE #570: Local sync does not resume partial transfer

tleedjarv:
This came out of the discussion at #571.

Closes #570 

It is my third iteration. The first iteration was very similar to this, just more hackish and with more changes in `Remote`. For second iteration, I wanted to make the `Remote` module simpler (could drop the `registerRootCmdWithConnection` function completely) but unfortunately that's not possible. It would mean the signatures of remote functions in `Copy` would have to change and that would break compatibility with 2.51. This third iteration is fully compatible with 2.51.

The code path taken by local and remote syncs is now the same. Maybe there are some optimizations that make sense locally (for example, bigger buffers? then again, all buffers are currently quite small so this question does not have to come in here). This PR does not look at that side of things at all.

I have done some very brief testing locally and remotely with ssh and socket server and also with 2.51.

(As draft so it doesn't get accidentally merged. The change itself is finished as of now.)

### DIFF
--- a/src/copy.ml
+++ b/src/copy.ml
@@ -144,7 +144,7 @@ let rec fingerprintPrefix fspath path offset len accu =
   end
 
 let fingerprintPrefixRemotely =
-  Remote.registerServerCmd
+  Remote.registerServerCmd'
     "fingerprintSubfile"
     (fun _ (fspath, path, len) ->
        Lwt.return (fingerprintPrefix fspath path 0L len []))
@@ -461,7 +461,7 @@ let compress conn
        Util.convertUnixErrorsToTransient "transferring file contents"
          (fun () -> raise e))
 
-let compressRemotely = Remote.registerServerCmd "compress" compress
+let compressRemotely = Remote.registerServerCmd' "compress" compress
 
 let close_all infd outfd =
   Util.convertUnixErrorsToTransient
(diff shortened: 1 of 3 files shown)


tleedjarv:
Rebased after more than 2 years. Having thought about it some more, I'm not convinced this is the right way to go. At least not without having some exceptions for the local sync anyway. Otherwise it can introduce more IO than what's done currently for local syncs (and remember: "local" can mean that sync happens on a network mount, which could even be a WAN, not just LAN). But the code is here for future reference.

PR #577, opened 2021-08-20
Accelerated local copy (reflink (CoW block cloning), alternatively in-kernel copy) when available

tleedjarv:
This is an interesting experiment which, if deemed safe and working properly, can potentially offer huge gains in some situations.

This patch makes use of various OS and filesystem-specific syscalls to accelerate local file copies, making so-called reflink (COW) copies, if possible.

The following syscalls are tried:
 * `clonefile(2)` on macOS (version >= 10.12); supported on APFS
 * `ioctl FICLONE` on Linux (kernel >= 4.5); several filesystems are supported, at least BTRFS and XFS
   (also `BTRFS_IOC_CLONE` since Linux 2.6.29(?) should work)
 * `copy_file_range(2)` on Linux (kernel >= 4.5) and FreeBSD >= 13
 * `sendfile(2)` on Linux (kernel >= 2.2) and `sendfile(3ext)` on Solaris
 * `ioctl FSCTL_DUPLICATE_EXTENTS_TO_FILE` on Windows (since Windows Server 2016 and at least Windows 11 (likely Windows 10)); supported on ReFS (including Dev Drive)

Fallback to read-write loop (the current copy function) is used if none of the above are available or supported by the filesystem.

The first three bullets provide special optimizations, while `sendfile()` is more of a question mark. It's supposedly more efficient due to not having to move data to userspace but I don't know if it results in any practical gains here.

Compilation on OS X < 10.12 is still possible but binaries produced by GHA CI will not be able to run on those versions. I don't know if it's going to be a problem for users.

### DIFF
--- a/src/copy_stubs.c
+++ b/src/copy_stubs.c
@@ -1,0 +1,445 @@
+/* Unison file synchronizer: src/copy_stubs.c */
+/* Copyright 2021-2025, Tõivo Leedjärv
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <caml/memory.h>
+#include <caml/threads.h>
+#include <caml/unixsupport.h>
+#include <caml/version.h>
+
+#if OCAML_VERSION_MAJOR < 5
+#define caml_unix_error unix_error
+#define caml_uerror uerror
+#define caml_win32_maperr win32_maperr
+#endif
+
+
+#include <errno.h>
+
+
+/* ----------------------------------------------- */
+/* Clone a file given source and destination paths */
+/* It must fully complete or fully fail.
+
+   The function must not raise any exceptions.
+
+   Return true for success and false for failure
+   or if the operation is not supported.           */
+
+#if defined(__APPLE__)
+
+
+#include <AvailabilityMacros.h>
+
+#if defined(MAC_OS_X_VERSION_10_12)
+#include <string.h>
+#include <sys/attr.h>
+#include <sys/clonefile.h>
+
+CAMLprim value unison_clone_path(value src, value dst)
+{
+  CAMLparam2(src, dst);
+  char *srcn, *dstn;
+  int status;
+
+  srcn = strdup(String_val(src));
+  dstn = strdup(String_val(dst));
+  caml_release_runtime_system();
+  status = clonefile(srcn, dstn, CLONE_NOFOLLOW | CLONE_NOOWNERCOPY);
+  free(srcn);
+  free(dstn);
+  caml_acquire_runtime_system();
+
+  /* Don't raise an exception, just return false in case of errors */
+  CAMLreturn(Val_bool(status == 0));
+}
+#else /* MAC_OS_X_VERSION_10_12 */
+CAMLprim value unison_clone_path(value src, value dst)
+{
+  CAMLparam2(src, dst);
+  CAMLreturn(Val_false);
+}
+#endif /* MAC_OS_X_VERSION_10_12 */
+
+
+#else /* defined(__APPLE__) */
+
+
+CAMLprim value unison_clone_path(value src, value dst)
+{
+  CAMLparam2(src, dst);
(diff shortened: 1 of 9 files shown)


tleedjarv:
Rebased, with minor fixes and improvements.

gdt:
FWIW, I think macOS < 10.13 is totally irrelevant now.

tleedjarv:
Rebased, added Windows support, and other minor fixes and improvements.

The feature is ready. More testing is always welcome, of course.

The only question now is, if this feature is to be merged, should it be gated behind a user preference? Given your OS and filesystem support it, why would anyone actively disable reflinking?

src/make_tools.ml
```diff
@@ -109,6 +109,8 @@ let () = if build_macGUI then outp "macuimaybe: macui"
 (* Generate backtrace information for exceptions *)
 let () = "CAMLFLAGS" <-+= "-g $(INCLFLAGS)"
 
+let () = "CAMLCFLAGS" <-+= "-ccopt -D_FILE_OFFSET_BITS=64"
```
gdt, line 112: Is `_FILE_OFFSET_BITS` defined by POSIX, or is this a glibc thing?   I've never been clear on "large file support" because NetBSD simply changed `off_t` etc. to be 64-bit types long ago, regardless of the CPU type, so there's never been any need for or discussion of "large file support"; large files have simply worked with the normal syscalls.  (I have been boggled at how other systems have a parallel set of calls, which seems awkward, but that's off topic.)

I couldn't find this in POSIX, which makes me think it should be defined only where needed.   And, I don't understand why it's new with this change, as unison should already handle files over 2 GB.

tleedjarv: This discussion happened over at #876 (although the change always originated from this PR).

> And, I don't understand why it's new with this change, as unison should already handle files over 2 GB.

That is correct, OCaml Unix library has had this defined since 2002. However, this is the first time we have C code touching `off_t` directly. It is for the benefit of this C code.

tleedjarv:
> With sendfile, this is no longer reflink/COW. That's ok, but it's misnamed. It's really "use OS support for copying files" in the headline, with "OS support may involve references and copy-on-write, which should be more efficient. It may also involve data copying by the kernel instead of userland, which also should be more efficient than userland copying". That's all great, but no longer centered on reflink.

All very true. I even thought about mentioning it but changed my mind. Now I can confirm that we have the same understanding.

src/make_tools.ml
gdt: That would be a good comment to explain it, then, that it aligns with what ocaml is already doing

gdt:
> The PR title is not as accurate but that's easy to change. Let me know if you have specific wording ideas you'd like to change in the commit message specifically.

Replacing the PR title with the single commit headline seems fine.

src/copy_stubs.c
```diff
@@ -0,0 +1,445 @@
...
+
+   Return true for success and false for failure
+   or if the operation is not supported.           */
+
+#if defined(__APPLE__)
+
+
+#include <AvailabilityMacros.h>
+
+#if defined(MAC_OS_X_VERSION_10_12)
```
gdt, line 47: During review by unison-hackers@, I would like someone familiar with macOS to check this guard expression.  I think the point is to find out if the target SDK is >= 10.12, and if so use clonefile, and if not, don't.  That means that targeting 10.11 would fail to use clonefile on modern systems, but that's ok because it does not make sense to target 10.11 in 2025, unless you are building your own unison for your ancient (2008?) machine.

It seems possible to switch at runtime, but that feels like pointless work, given we're only talking about machines that are ~17 years old.   (Macs from 2010 can run 10.13.)

tleedjarv: I don't know the correct terminology but this is not necessarily the target SDK. This is checking the compilation SDK (which I understand can still allow running on older systems (the target), but maybe I'm completely mistaken). This is a compile-time guard because we don't have <sys/clonefile.h> in < 10.12 and, ignoring that, the linker would probably not be happy. My understanding is that this guard needs to be in place no matter what.

But compiling in >= 10.12 with SDK >= 10.12, it could be possible to add a runtime switch when `MAC_OS_X_VERSION_MIN_REQUIRED` is defined to be < 10.12. That would theoretically instruct the linker to mark `clonefile` as a weak import. I can't verify it because the only macOS I have access to is the GHA runners.
If `MAC_OS_X_VERSION_MIN_REQUIRED` is not explicitly defined then it is implicitly set by `AvailabilityMacros.h` based on the target SDK, which in the GHA CI was recently set to 10.13 by 0ec31df396fd1e669165c4ffd0439c46f07dee24, alternatively to whatever the default minimum is in `AvailabilityMacros.h`.

In summary, for GHA-built binaries this no longer matters anyway (as the minimum is set to 10.13).
For those building themselves, I don't mind adding a runtime switch if someone could test if it actually worked (not sure if clonefile can be weakly linked or not) and if the check is as simple as `if (clonefile)`

tleedjarv:
Did some minor refactoring, no functional changes.

PR #582, opened 2021-09-04
Fix macOS builds (inotify doesn't exist on macOS)

### LINKED ISSUE #581: opam build of unison on macOS fails because sys/inotify.h not found
This is on Mac OS X, Version 10.15.7, Build 19H1323.

Here's the build error:
```
$ opam install unison
The following actions will be performed:
  ∗ install unison 2.51.4

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫
⬇ retrieved unison.2.51.4  (cached)
[ERROR] The compilation of unison.2.51.4 failed at "dune build -p unison -j 7".

#=== ERROR while compiling unison.2.51.4 ======================================#
# context     2.1.0 | macos/x86_64 | ocaml-base-compiler.4.12.0 | https://opam.ocaml.org#5c5a526e
# path        ~/.opam/forunison/.opam-switch/build/unison.2.51.4
# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p unison -j 7
# exit-code   1
# env-file    ~/.opam/log/unison-8482-d96a86.env
# output-file ~/.opam/log/unison-8482-d96a86.out
### output ###
#           cc src/fsmonitor/linux/inotify_stubs.o (exit 1)
# (cd _build/default/src/fsmonitor/linux && /usr/bin/cc -O2 -fno-strict-aliasing -fwrapv -D_FILE_OFFSET_BITS=64 -D_REENTRANT -I/usr/local/opt/expat/include -O2 -fno-strict-aliasing -fwrapv -g -I /Users/tkb/.opam/forunison/lib/ocaml -I ../../lwt -o inotify_stubs.o -c inotify_stubs.c)
# inotify_stubs.c:22:10: fatal error: 'sys/inotify.h' file not found
# #include <sys/inotify.h>
#          ^~~~~~~~~~~~~~~
# 1 error generated.
```

kit-ty-kate:
Fixes #581 
I have no idea what's the macOS equivalent of `inotify`, or even how deeply unison needs `unison-fswatcher` but at least this makes the rest compile on macOS

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -244,6 +244,7 @@ jobs:
       matrix:
         job:
         - { os: ubuntu-latest  , ocaml-compiler: 4.12.x }
+        - { os: macos-latest  , ocaml-compiler: 4.12.x }
 
     runs-on: ${{ matrix.job.os }}
 
--- a/dune-project
+++ b/dune-project
@@ -1,2 +1,2 @@
-(lang dune 1.3)
+(lang dune 2.3)
 (name unison)
--- a/src/dune
+++ b/src/dune
@@ -7,7 +7,9 @@
         -w -3-6-9-10-26-27-32-34-35-38-39-50-52
         -warn-error -3-6-9-10-26-27-32-34-35-39-50-52
         -no-strict-sequence)
- (c_names bytearray_stubs osxsupport pty hash_compat)
+ (foreign_stubs
+  (language c)
+  (names bytearray_stubs osxsupport pty hash_compat))
  (c_library_flags -lutil)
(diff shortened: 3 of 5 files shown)


tleedjarv (approved):
Looks good.

gdt:
I don't fully grasp this but given that both of you think it's right and things are broken I'm going to hit merge and we'll see how things go.

My impression is that inotify is a Linux interface, possibly implemented elsewhere, and there is some sort of kqueue thing which is similar but different on BSDs and I have no idea how mac lands, being sort of Mach, sort of BSD, sort of Think Different.

PR #600, opened 2021-11-15
Add static linux version for linux 4.12

mergwyn:


mergwyn:
I raised this PR to get a Linux version which matched macOS

gdt:
I don't understand how making it static makes it match macOS.

In general, the point of CI is CI, not providing binaries for users, and I think we already have too many build flavors (on not enough operating systems).   That doesn't mean that this flavor isn't useful.

So:

  * can you explain why adding a static/musl/? version of a os/cpu/ocaml version that is already present is useful?
  * can you explain why the normal version doesn't run, and perhaps edit the CI Binary Instructions page to explain what the normal GNU/Linux binaries depend on in the environment?
  * can you explain "match macOS"?
 
I'll bring up the larger issues on unison-hackers, which I hope you're on.

.github/workflows/CICD.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         - { os: macos-10.15   , ocaml-version: 4.03.0 }
         - { os: macos-10.15   , ocaml-version: 4.02.3 }
         - { os: macos-10.15   , ocaml-version: 4.01.0 }
-        - { os: ubuntu-18.04  , ocaml-version: 4.12.0 }
+        - { os: ubuntu-18.04  , ocaml-version: 4.12.0+musl+static+flambda }
```
tleedjarv, line 30: Don't remove the regular 4.12 build. Should the static build be added to CI, it should not replace the regular 4.12 build.

gdt: Thanks; I didn't spot that.

I expect that after the discussion on hackers we'll make some other changes and not merge this, although this PR is useful for starting that larger discussion.

gdt:
No answers to questions after 30 days, so this has timed out.
Please feel welcome to discuss the larger "what CI targets should we have" on unison-hackers@, as it's certainly a difficult question.

PR #609, opened 2021-12-18
CI: Drop ocaml < 4.08 except for Ubuntu

gdt:
There are too many builds in our CI.  Hving old ocaml on ubuntu is
sufficient to test building with old ocaml in almost all cases, and
there's almost no additional CI value in doing those builds on other
systems.

It's an arbitrary guess to set 4.08 as the cutoff.  My view is that
anyone with a system with ocaml older than that is woefully out of
date, as 4.08 was released 2.5 years ago.


gdt:
Floated on -hackers with no objections.

Issue #590, opened 2021-10-04
Allow to silently ignore symlinks


azrdev:
I'm syncing between a Linux filesystem and a Windows one (smb with `mount.cifs`), thus symlinks are not supported.
They appear in the synchronization list and summary, though, as failed transfers, which makes for some noise I have to filter manually when reviewing the changes.

Please add a flag `-links ignore` to just skip symlink syncing when it's known in advance they are not supported. `-links false` skips them, but yields an error "is a symbolic link".

chrisl8:
I wonder if PR #535 might ultimately address this? If symlinks **can** work on Windows, then can we stop getting the errors?

I'm not actually sure if the "is a symbolic link" is an error or just that the code is set to always whine about the things you asked it to ignore?

Either way, at least that PR is a useful reference for someone attempting to address this issue in the code, such as myself.

gdt:
@chrisl8 Can you test the situation that you think @azrdev is encountering, on 2.51.5?    I don't use windows at all, and I'm really unclear on the sequencing of #535 and this report.

azrdev:
Actually I'm on linux, sorry that I didn't mention it explicitly.

chrisl8:
Sorry, upon reflection I realized I was confusing two experiences together.

The PR did help me understand the code, but this issue seems to be merely a message about every skipped simlink, not errors.

I wasn't initially sure if those messages were errors or not. I can see now that they are not.

This issue does affect me though. When I skip simlinks on a large sync I get hundreds of the messages, making it hard to see what is happening.

I intend to dig into the code and attempt an update to fix this soon, if nobody else gets to it first.

tleedjarv:
This is not as easy to solve as it may look like initially. A change in preferences (either by going for the solution suggested by the original reporter (to add a `-links ignore` flag) or by adding a separate new preference) is a version-incompatible change.

Just recently a PR was merged, which makes changes like this finally possible while keeping backwards compatibility. With the "features" mechanism introduced by that PR, a solution with a separate new preference is by far the easiest.

I can't judge if this is a good thing to have (it seems that there are several users who would like to see it) and I can't judge if it's really worth adding yet another preference (or a new option to existing preference), but I'm willing to quickly produce a prototype to demonstrate the solution.

tleedjarv:
[...]

PR #623, opened 2022-01-17
Allow to silently ignore symlinks

tleedjarv:
~~This is a quick'n'dirty prototype to demonstrate a possible solution to #590. This patch is completely backwards-compatible due to leveraging the recently-merged "features" mechanism.~~

~~I don't claim this to be correct or appropriate solution, it just shows one possible way forward while keeping backwards compatibility intact.~~

New solution, no longer quick'n'dirty. Still completely backwards compatible.

Closes: #590

### DIFF
--- a/src/fileinfo.ml
+++ b/src/fileinfo.ml
@@ -24,12 +24,14 @@ let allowSymlinks =
     "allow the synchronization of symbolic links (true/false/default)"
     ("When set to {\\tt true}, this flag causes Unison to synchronize \
       symbolic links.  When the flag is set to {\\tt false}, symbolic \
-      links will result in an error during update detection.  \
+      links will be ignored during update detection.  \
       Ordinarily, when the flag is set to {\\tt default}, symbolic \
       links are synchronized except when one of the hosts is running \
       Windows.  On a Windows client, Unison makes an attempt to detect \
       if symbolic links are supported and allowed by user privileges.  \
-      You may have to get elevated privileges to create symbolic links.")
+      You may have to get elevated privileges to create symbolic links.  \
+      When the flag is set to {\\t default} and symbolic links can't be \
+      synchronized then an error is produced during update detection.")
 
 let symlinksAllowed =
   Prefs.createBool "links-aux" true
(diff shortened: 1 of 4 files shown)


tleedjarv:
Rebased and changed the solution. Thanks to umarshal, now the existing `links` preference is extended with a new value `ignore`.

I am relatively confident that this will not impact existing users. With some positive testing reports from such users (also interop with 2.51) I would feel comfortable merging this into the upcoming release.
I also think that it does what is requested by users in #590 but I don't claim to have all corner cases covered.

tleedjarv:
Rebased and changed the solution again. The previous solution could not cover all the corner cases because it worked independently on each host/replica.

The current solution collects all the information from both hosts/replicas and works in the reconciliation stage. This of course comes with a slight downside of changing the reconciliation code. On the upside, the changes are very contained and should be very easily verified as not changing existing semantics.

The code itself is kept as small and local as possible, even though it comes at the cost of some (theoretical) fragility. A type-based solution would clearly be superior but would come with a much larger changeset, not to mention a compatibility burden. I did not think the need for/benefits of this feature warranted this approach.

This functionality is compatible with previous versions and will also work with older version servers.

With this patch, the following is true:
- if `links = true`, nothing changes; scanning errors are reported at scanning and propagation errors are reported at propagation.
- if `links = false` then symlinks are silently ignored, unless they conflict with the other replica (reported as error).
- if `links = default` then nothing changes. If default means `false` then all symlinks are reported as errors during scanning.

tleedjarv:
Not proposing to do this, just mentioning:
Other design questions aside, this code is completely generic and can be used to ignore paths based on other types (or even all unknown types). I've seen a request to silently ignore Unix sockets, for example.

gdt:
This seems ok but I would like to see the people @chrisl8 @azrdev  who have this problem in #590 test the most recent change.
(Personally, I don't sync anything in symlink-deficient filesystems, so I'm not set up to test.)

tleedjarv:
Rebased. Removed the draft status due to some positive testing feedback.

This patch can be tested also by those who don't use this feature because the changes are in the part of reconciliation that deals with scanning errors (no impact for normal syncs).

gdt:
It seems there's more testing once merged, so let's do it that way :-)

azrdev:
AIUI this has been merged into master, but not released yet. Where would I find binaries to test?
Have been using Fedora and <https://download.opensuse.org/repositories/Archiving:/unison/> so far, but that stopped working anyways since I upgraded to fedora 36.

tleedjarv:
You can see all release candidate builds at https://github.com/bcpierce00/unison/actions/runs/3356812838
It would be appreciated if you could test during this week. The release is currently scheduled for next week.

azrdev:
After upgrading to Fedora 37, OBS now gave me a rpm of unison 2.53.0 which contains that MR. I just tested it on the linux box with a (kernel) CIFS mount not supporting links:
Apparently `-links false` makes it skip links, instead of throwing "unsupported" errors. Thus it works as intended, hooray!

Thank you for the work and the patience!

juk0de:
Does the new `links` option also work with profiles? I've added `links=false` to my unison profile but now I get the following error for every symlink (in the unison GUI):
```
Error: path /bla/foo is a symbolic link
```
That's different from the old behaviour, where it tried to synchronize the symlinks and failed. But it's still not what I'd call "silently ignored"... Is there a way to suppress those error messages?

PR #625, opened 2022-01-21
Further fixes to uimac ssh interaction

tleedjarv:
Further fixes to the issue first reported as #328.

The first two commits are not specific to uimac.

Commit 763c5d85298e7e9a03a920439cfa513647adb3dc redirected ssh's stderr to the pty in order to scrape any error output and present it to the GUI users (note, this only affects the GUI). This works well in general. However, uimac had a bug in its processing of ssh output (unrelated to stderr redirection). First fix for uimac was produced in PR #465 but it was not sufficient nor fully correct. This patch tries to make ssh interaction in uimac more resilient; this applies with or without stderr redirection.

~~**Testing needed.**~~


tleedjarv:
Testing successful: https://github.com/bcpierce00/unison/issues/328#issuecomment-1019183290

PR #626, opened 2022-01-30
Umarshal - version-independent serialization format

tleedjarv:
### Description

The long-awaited improvement is finally here!

Umarshal is the new serialization format that will be used for client-server communication and archive files to make them independent of OCaml versions and Unison versions. This code will keep compatibility with 2.51 releases and future releases.

_No longer must you match Unison versions and OCaml versions (the latter applies only if both client and server have this PR)._

Closes #375
Closes #377
Closes #407

### Credits

The author of this work is @glondu.

I am just taking his work, merging with my own previous work (#507 and #509) and adding some tweaks to make sure that it remains compatible with 2.51 releases and a wide range of OCaml versions.

### Interoperability

Full compatibility with 2.51.x versions is intended (both as a server and as a client). _To interop with any 2.51.x version, you have to still match the OCaml versions._

### Upgrading

It is possible to upgrade from 2.51.x versions while keeping the archive intact. This means that no re-scan of the replicas is needed and the replicas don't have to be in sync before the upgrade. The user does not have to do anything for the upgrade -- just drop in the new version and run it. It will automatically pick up the existing archive and convert it to the new format.

_To upgrade from any 2.51.x version, you have to still match the OCaml versions._ Once you have upgraded all the clients and servers that have to work together, then you no longer have to match OCaml versions.

**Edit:** The upgrade works also from 2.48.x and maybe even older versions, if the OCaml versions match (or are similar, at least).

**Note:** once the upgrade has succeeded, you should not go back and run the old version. To go back to old version, you have to delete the archive files and start afresh.

### Testing

**Call for testing!** Everyone who would like to contribute is welcome to do testing (there is no script; test any way you like). Even though this code should not cause any data loss or corruption, it is best to test on special replicas or have working backups at hand.

Since this code is expected to be version-independent and backwards-compatible with 2.51 releases, various combinations of client/server versions and OCaml versions must be tested. I don't expect all combinations below to be tested but the tables are provided for a quick overview of which combinations have been tested.

**A. Umarshal code (this PR) with different OCaml versions**

**rows** = client; **columns** = server
L = Linux; W = Windows; i = illumos; L2L = Linux-to-Linux; etc.

<table>
<thead>
  <th></th>
  <th>4.14</th>
  <th>4.13</th>
  <th>4.12</th>
  <th>4.11</th>
  <th>4.10</th>
  <th>4.09</th>
  <th>4.08</th>
  <th>4.07</th>
  <th>4.06</th>
  <th>4.05</th>
[...]

PR #628, opened 2022-02-04
Restore compatibility with 2.48

tleedjarv:
### Description

Many users are still using Unison 2.48 and are unable to upgrade. This patch provides a best effort compatibility with 2.48. Even though there is no indication that interop with 2.48 clients and servers would not work, **there are no guarantees of success**. If it doesn't work, however, it is very unlikely to result in data loss or corruption.

With this patch, a single Unison version will be able to interoperate with 2.48, 2.51 and the future releases. It is also able to upgrade from 2.48 without needing to rescan the replicas.

Support for 2.48 is provided according the the table below. **OCaml versions must still match.**
<table>
<tbody>
<tr><th colspan="2" rowspan="2"></th><th colspan="2">Server</th></tr>
<tr><th>2.48</th><th>With this PR</th></tr>
<tr><th rowspan="2">Client</th><th>2.48</th><td>OK, native</td><td>Special setup is needed, see below</td></tr>
<tr><th>With this PR</th><td>OK, automatically</td><td>OK, automatically</td></tr>
</tbody>
</table>

### Server for 2.48 clients

Server support for 2.48 clients is not fully automatic. For the server to support 2.48 clients, the executable must be named with `unison-2.48` prefix. Only the prefix matters. The full executable name could be `unison-2.48.exe` or `unison-2.48-but-actually-2.52` or whatever you like. By using the name `unison-2.48` there is the benefit of being able to use the `-addversionno` preference on the client (relevant only for SSH connections). With other names, use the `-servercmd` preference on client (relevant only for SSH connections).

A socket server started this way is able to serve 2.48 clients and any clients with this PR or with #626 but not 2.51 clients.

### Notes on OCaml compiler version

It is possible to see the OCaml version used to compile Unison by executing `unison -version`. 2.48 did not display the OCaml version this way and it can be a bit difficult to find out the version. The easiest way is to see what is the `ocaml` package version in the same package repository where you got `unison` package from (hint, for Linux distributions, the typical version is 4.05). If that's not possible then the other way is trying your way through various versions (take this patch compiled with 4.05 and try, then compiled with another version and try and so on).

### Testing

Volunteer testing is expected of users who would like to use this patch.

### DIFF
--- a/src/globals.ml
+++ b/src/globals.ml
@@ -261,7 +261,7 @@ let ignorenotPred =
      synchronized will not work.  Instead, you should use the {\\tt path}
      preference to choose particular paths to synchronize.")
 
-let atomic = Pred.create "atomic" ~advanced:true
+let atomic = Pred.create "atomic" ~local:true ~advanced:true
   ("This preference specifies paths for directories whose "
    ^ "contents will be considered as a group rather than individually when "
    ^ "they are both modified.  "
--- a/src/remote.ml
+++ b/src/remote.ml
@@ -1134,6 +1134,9 @@ let connectionHeader = "Unison RPC\n"
 let compatConnectionHeader = "Unison 2.51 with OCaml >= 4.01.2\n"
 (* Every supported version released prior to the RPC version negotiation
    mechanism uses this connection header string. *)
+let compat248ConnectionHeader = "Unison 2.48\n"
+(* Additionally, even 2.48 can be supported, even though that support is
+   not official. *)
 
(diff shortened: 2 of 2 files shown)


gdt:
Given the number of people who seem stuck on 2.48, surely there will be testers.

I just sent a message to unison-hackers@ asking for testing.  I'fter umarshalis merged and probably released in an alpha. and we can rebase this and I'll followup to -users@

PR #629, opened 2022-02-06
GHA: Run self-tests over a socket

tleedjarv:
Currently, self-tests are run on local roots only. Run self-tests also over a socket to exercise the RPC code.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -166,6 +166,15 @@ jobs:
 
     - run: opam exec -- make test
 
+    - name: Run self-tests over RPC
+      shell: bash
+      run: |
+        # Separate backup dir must be set for server instance so that the central
+        # backup location of both instances doesn't overlap
+        UNISONBACKUPDIR=./src/testbak2 ./src/unison -socket 55443 &
+        sleep 1 # Wait for the server to be fully started
+        ./src/unison -ui text -selftest testr1 socket://127.0.0.1:55443/testr2 -killserver
+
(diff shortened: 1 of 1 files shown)


gdt:
Conceptually this is great!

That seems to open up a socket, from which random connections could cause writing in that root, and with bugs, arbitrary things.  But this is GHA, so it's a controlled environment and because they are github-hosted runners, VMs are created every time, so this is safe.  Do you agree with that logic?

Is it easy to limit thel listening socket to 127.0.0.1?

gdt:
Also, I wonder if there is a way to move this into 'make test', so that it happens for regular users building, packaging self-test, etc.  There it is necessary to not listen on the net, and even a localhost socket is a concern (vs a unix-domain socket that is accessible only to the same uid).  Probably too much to ask, but I thought I'd ask anyway :-)

tleedjarv:
Yes, the whole thing builds on the fact that with GHA VMs are created every time and destroyed at the end. This is also the reason why I did not include it in "make test"; it becomes too messy and potentially dangerous.

> Is it easy to limit thel listening socket to 127.0.0.1?

No, not with current code.

gdt:
OK, will merge then.  It would be nice to extend this; I'll open a wish-list issue.

tleedjarv:
> > Is it easy to limit thel listening socket to 127.0.0.1?
> 
> No, not with current code.

I have to correct myself. It is possible with current code, and I wasn't even aware of it. I'll soon prepare another PR.

PR #632, opened 2022-02-09
Capture stderr of the ssh process in GUI, take two

tleedjarv:
This is a new attempt at improving the capture and displaying of the stderr output from the ssh process. The previous attempt and a hack were a little too naive and primitive to do any good.

This whole thing is a bit messy but I still maintain that it is better for the user this way, rather than dumping all errors into the local stderr which the GUI users normally wouldn't see. The patch is also slightly more complicated and cumbersome due to limitations on Windows.

More details are in the commit message; in a nutshell this patch is about:

 - Detecting when the ssh connection has been established and processing stderr differently. Before the connection is established, all errors are assumed to be from the ssh process and must be displayed to the user in GUI. After the connection has been established, all output from stderr is assumed to be from the remote unison process and this output is dumped on the local stderr. This also fixes a bug where the GUI could hang.
 - Decoupling Lwt threads for reading from the controlling terminal of ssh process and for prompting the user in GUI. This allows for collecting the ouput from ssh with mutiple `read`s and building a buffer before displaying it to user. This fixes an issue where a prompt was displayed to the user before all output was collected from ssh. This was not only confusing to the user, it was also dangerous as the user had an opportunity to respond to the prompt without knowing what ssh was prompting for.

### DIFF
--- a/src/remote.ml
+++ b/src/remote.ml
@@ -1360,7 +1360,7 @@ let negociateFlowControl conn =
 
 (****)
 
-let initConnection onClose in_ch out_ch =
+let initConnection ?(connReady=fun () -> ()) onClose in_ch out_ch =
   let conn = setupIO false in_ch out_ch in
   let with_timeout t =
     Lwt.choose [t;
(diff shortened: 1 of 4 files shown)


gdt:
This is next in line for merging, after CI completes.

PR #633, opened 2022-02-10
CI: pin windows versions

gdt:
Use 2022 for 4.12, adn 2019 for everthing else, to get coverage, while
making less likely the binaries are unuseful.


tleedjarv:
All Windows builds succeeded but saving opam cache failed on windows-2022. This may be a temporary glitch, could you re-run the jobs to verify?

tleedjarv (approved):
Looks good. Minor nit: the commit message could be updated.

Issue #630, opened 2022-02-06
Stop using TCP sockets for testing, because they are unsafe, and instead use UNIX-domain sockets (on POSIX systems)


gdt:
Currently make test only does tests with two local roots.  CI tests now (#629) also test over a socket, to exercise the RPC protocol.  That is only safe because of the controlled environment of the github runner, and because the runner VMs are discarded after testing.

This issue is about adding features as necessary and modifying make test so that:
  * There is a way to use socket mode with unix-domain sockets
  * Tests create unix-domain sockets, mode 0600, so that other uids do not gain access to socket-mode unison spun up for testing
  * Tests actually exercise the RPC protocol

(This issue is not about adding tests for cross-unison-version RPC migration.)

gdt:
It's unclear if Unix-domain sockets work on Windows, so I'm saying this issue is about
 * use unix-domain sockets for 'make test' on POSIX (everything except Windows) 
 * use unix-domain sockets in CI
 * stop using TCP sockets in tests entirely, because they are unsafe, except perhaps with some "make socket-tests-yes-i-know-that-is-dangerous" option
 * 
It is not necessary to remove TCP sockets in testing on WIndows to close this ticket, but if that's how it comes out there should be a new issue for Windows opened as this is closed.

gdt:
Status now is:
  1) windows with old ocaml still uses  sockets
  2) CI tests run old versions for interop, which still use sockets

and our approach is
  1) people probably shouldn't use old ocaml on windows, or really anywhere
  2) CI tests use throwaway VMs and there's no way to time travel to fix old versions

so therefore this is as done as it is going to get, and there's nothing feasible remaining.

PR #639, opened 2022-02-17
Add support for Unix domain sockets

### LINKED ISSUE #163: Support Unix domain sockets for -socket option
Unix domain sockets would allow safe use of the `-socket` server option in many circumstances. For example, by using `spiped` to forward encrypted network connections to Unison without making the server available to every process running on the same host.

I suggest the `-socket` interpret an argument that begins with `/` as the absolute path name of a Unix domain socket to create and listen on. Meanwhile, in the specification of a root, a URL such as `socket:///path/name` would specify the Unix domain socket to which to connect.

tleedjarv:
What it says on the tin.

There are some collateral, somewhat loosely related, changes included.

- For TCP sockets, improve capabilities of manually specifying which addresses to listen on. Rename the command-line argument from "host", which is somewhat inaccurate and even misleading, to "listen", keeping the old name as alias. Allow the argument to be repeated so that multiple listening endpoints can be specified. Socket tests can now just add `-listen 127.0.0.1` (or `-listen ::1`, or both or just `-listen localhost`) to the server invocation. A fix was needed for the alias to properly load when client sends preferences to server.

- Change the code for parsing IPv6 literals in roots. While not strictly necessary for any of this, when working with the related code it just bugged me so much that I had to fix it.

- Add support for Unix domain sockets. I think mostly it should be pretty self-explanatory. Potential bikeshedding moment here is the syntax when specifying roots over the Unix socket. I currently settled for `socket://{/path/to/socket}/root/path`. The socket path is in curly braces. I don't know what does the latest RFC about URIs say but creating URLs over domain sockets does not seem to have a good standard way. I did not add any documentation about this support yet. Once any discussions have finished and the final design is confirmed, I will update the manual accordingly. I overloaded the `-socket` command-line argument to start the server with a Unix domain socket. When `-socket` is specified with an integer then it is treated as a port number and a TCP socket is created; otherwise it is treated as a path to a Unix socket. There are absolutely no limitations as to what the socket path may be, it is blindly passed on to the kernel. See example in the GHA patch below.

I did not add any test yet, not in Makefile, not in GHA jobs. I will add to GHA once any discussions have concluded. For those interested, I got green runs with the following patch.

```diff
diff --git a/.github/workflows/CICD.yml b/.github/workflows/CICD.yml
index 56c62298..1c0ef424 100644
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -175,6 +175,21 @@ jobs:
         sleep 1 # Wait for the server to be fully started
         ./src/unison -ui text -selftest testr1 socket://127.0.0.1:55443/testr2 -killserver
 
+    - name: Run self-tests over local socket
+      # Recent Windows versions do support Unix domain sockets
+      # but at least OCaml 4.14 is required to use that support
+      if: runner.os != 'Windows'
+      shell: bash
+      run: |
+        mkdir localsocket
+        chmod 700 localsocket
+        # Separate backup dir must be set for server instance so that the central
+        # backup location of both instances doesn't overlap
+        UNISONBACKUPDIR=./src/testbak4 ./src/unison -socket ./localsocket/test.sock &
+        sleep 1 # Wait for the server to be fully started
+        test -S ./localsocket/test.sock
+        ./src/unison -ui text -selftest testr3 socket://{./localsocket/test.sock}/testr4 -killserver
+
     - if: steps.vars.outputs.STATIC != 'true' ## unable to build static gtk for linux or windows/Cygwin MinGW platforms
       shell: bash
       run: |
```

Closes #163
It gets #630 to 90% completion but doesn't do any make test modifications. Volunteers needed to provide a make recipe for a portable (even some Windows versions are supposed to have Unix sockets nowadays!) test over Unix sockets.

### DIFF
--- a/doc/unison-manual.tex
+++ b/doc/unison-manual.tex
@@ -845,7 +845,10 @@ The full grammar is:
   \NT{user} ::= [-\_a-zA-Z0-9]+
 
   \NT{host} ::= [-\_a-zA-Z0-9.]+
+        |  \textbackslash[ [a-f0-9:.]+ \NT{zone}? \textbackslash]    IPv6 literals (no future format).
         |  \{ [\^{}\}]+ \}                   For Unix domain sockets only.
+
+  \NT{zone} ::= \%[-\_a-zA-Z0-9~\%.]+
 
   \NT{port} ::= [0-9]+
 \end{alltt}
--- a/doc/unison-manual.tex
+++ b/doc/unison-manual.tex
@@ -842,9 +842,10 @@ The full grammar is:
             |  ssh
             |  rsh
 
-  \NT{user} ::= [-_a-zA-Z0-9]+
+  \NT{user} ::= [-\_a-zA-Z0-9]+
 
-  \NT{host} ::= [-_a-zA-Z0-9.]+
+  \NT{host} ::= [-\_a-zA-Z0-9.]+
+        |  \{ [\^{}\}]+ \}                   For Unix domain sockets only.
 
   \NT{port} ::= [0-9]+
 \end{alltt}
--- a/src/clroot.ml
+++ b/src/clroot.ml
@@ -113,24 +113,15 @@ let getUser s =
     (Some beforeAt,afterAt)
   else (None,s)
 
-(*ipv6 support*)
-let hostWithBracketsRegexp = Str.regexp "\\[.*\\]"
 (* Hostname, IP or Unix domain socket path *)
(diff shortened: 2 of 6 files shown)


gdt:
This is great to see.   I will review the code.   I'm happy to merge the feature without the test changes, and I'd then love it if someone(tm) changed all TCP socket tests to unix-domain socket tests, at least on !windows.

tleedjarv:
(Pushed again, fixed additions in the manual.)

Issue #325, opened 2020-03-07
URLs in PDF version of manual run off end of line and are truncated


rickhg12hs:
It would be nice if all URLs in the PDF document were readable.  After building docs from master branch (March 2020), I get unfortunate chopped URLs.

![Screenshot from 2020-03-07 23-21-30](https://user-images.githubusercontent.com/6000188/76153310-7cb33100-60ca-11ea-8784-47adb306ba7c.png)

vaukai:
There are some more "Overfull" lines: (ebuild is from https://github.com/gentoo/gentoo/pull/17989/commits/e2d1e06)
USE='doc gtk' ebuild unison-2.51.3_p20201024.ebuild clean compile | grep Overfull | sort | uniq
Overfull \hbox (109.6327pt too wide) in paragraph at lines 806--812
Overfull \hbox (1.46704pt too wide) in paragraph at lines 2113--2121
Overfull \hbox (1.86284pt too wide) in paragraph at lines 125--126
Overfull \hbox (219.5766pt too wide) in paragraph at lines 119--121
Overfull \hbox (22.49097pt too wide) in paragraph at lines 1339--1339
Overfull \hbox (27.74092pt too wide) in paragraph at lines 1339--1339
Overfull \hbox (61.6224pt too wide) in paragraph at lines 319--322
Overfull \hbox (93.30688pt too wide) in paragraph at lines 1044--1050

PR #641, opened 2022-02-18
doc: Small fixes

### LINKED ISSUE #416: Building in doc with HEVEA=false still calls hevea
When compiling with "HEVEA=false" the second  "hevea unison-manual.tex" in [doc/Makefile](https://github.com/bcpierce00/unison/blob/master/doc/Makefile#L53-L55) still calls hevea.

What is the purpose of the "ifdef HEVEA ..."?

tleedjarv:
Various small fixes to documentation. Re-generate strings properly in UTF-8.

Closes #416 #325


gdt:
I'm not sure what the UTF-8 issue is, but if the underlying problem is not fixed (that the doc build process does not properly control locale, or something like that), it isn't going to really help as the other way is going to keep happening.

tleedjarv:
No, the underlying problem is not fixed. I don't know what it is.

tleedjarv:
I figured out the UTF-8 issue. It's lynx, not hevea (duh!). Patch and rebase coming soon.

PR #642, opened 2022-02-18
GHA: Add CI for docs

### LINKED ISSUE #412: Add CI for docs
We build src/ but not doc/ in CI, and it would be nice to build the docs, both to verify that they build, and to have the build products.

The build should use havea  to make html, but any steps forward are of course useful.

Probably we only need to do this on GNU/Linux, and only for more recent versions of ocaml; 4.09 and 4.11 might be enough.

Related to #411

tleedjarv:
This adds a CI job to run `make docs`. The output is picked up by all build jobs but only packaged when publishing the artifacts. To get the docs from a non-publishing CI run, download the separate 'unison-docs' artifact.

txt, html and pdf format manuals are added in the package.

This makes building docs a pre-requisite for building binaries. It's not ideal but it's pretty fast when caches are hot. Failure in `make docs` will not stop the builds, so that CI can still function for primary builds.

Closes #412

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -10,7 +10,36 @@ on:
   - push
 
 jobs:
+  docs:
+    runs-on: ubuntu-20.04
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Use OCaml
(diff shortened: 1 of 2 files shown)


gdt:
enough has passed that I'm sure this is ok.

PR #648, opened 2022-02-21
Fix HTML TOC links

### LINKED ISSUE #610: broken TOC links in unison-manual.html
The links in the TOC have no target, there is no `#news` for example. Unclear when it broke.

tleedjarv:
I don't know what I'm doing but this workaround seems to fix HTML TOC links without any other side-effects...

Fixes #610


gdt:
Seems reasonable to me.

PR #658, opened 2022-03-02
provide packages in dune-project

olafhering:
This will allow `dune build @install -p unison,unison-gtk2,unison-fsmonitor`

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -1,2 +1,77 @@
 (lang dune 2.3)
+
 (name unison)
+
+(version dev)
+
+(generate_opam_files false)
+
+(license GPL-3.0-or-later)
+
+(maintainers "maintainers")
+
+(authors "Trevor Jim, Benjamin C. Pierce and J\195\169r\195\180me Vouillon")
+
+(source
+ (github bcpierce00/unison))
+
+(homepage https://www.cis.upenn.edu/~bcpierce/unison/)
+
+(documentation "https://github.com/bcpierce00/unison/wiki")
+
+(package
+ (name unison)
+ (synopsis "File-synchronization tool for Unix and Windows")
+ (description "
+Text based userinterface for Unison.
+
+Unison is a file-synchronization tool for Unix and Windows.
+It allows two replicas of a collection of files and directories
+to be stored on different hosts (or different disks on the same host),
+modified separately, and then brought up to date by propagating
+the changes in each replica to the other.
+")
+ (depends
+  (ocaml
+   (>= 4.03))
+  (dune
+   (>= 2.3))))
+
+(package
+ (name unison-gui)
+ (synopsis "File-synchronization tool for Unix and Windows")
+ (description "
+Graphical userinterface for Unison.
+
+Unison is a file-synchronization tool for Unix and Windows.
+It allows two replicas of a collection of files and directories
+to be stored on different hosts (or different disks on the same host),
+modified separately, and then brought up to date by propagating
+the changes in each replica to the other.
+")
+ (depends
+  (ocaml
+   (>= 4.03))
+  (dune
+   (>= 2.3))
+  (lablgtk2
+   (>= 2.18.6))))
+
+(package
+ (name unison-fsmonitor)
+ (synopsis "File-synchronization tool for Unix and Windows")
+ (description "
+This helper application monitors file system changes on all given
+(relative to root) paths.
+
+Unison is a file-synchronization tool for Unix and Windows.
+It allows two replicas of a collection of files and directories
+to be stored on different hosts (or different disks on the same host),
+modified separately, and then brought up to date by propagating
+the changes in each replica to the other.
+")
+ (depends
+  (ocaml
+   (>= 4.03))
+  (dune
+   (>= 2.3))))
--- a/src/dune
+++ b/src/dune
@@ -18,12 +18,14 @@
 (executable
  (name linktext)
  (public_name unison)
+ (package unison)
  (modules linktext)
  (libraries unison_lib))
 
 (executable
  (name linkgtk2)
  (public_name unison-gui)
+ (package unison-gui)
  (flags :standard -w -3-6-9-27-32-52)
  (modules linkgtk2 uigtk2)
  (libraries threads unison_lib lablgtk2))
--- a/src/fsmonitor/linux/dune
+++ b/src/fsmonitor/linux/dune
@@ -20,6 +20,7 @@
 (executable
  (name watcher)
  (public_name unison-fsmonitor)
+ (package unison-fsmonitor)
  (enabled_if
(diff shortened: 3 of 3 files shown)


tleedjarv:
I don't know enough about dune to properly review, so here are some naive comments. Perhaps you're still working on it and would have addressed some of these points anyway.

 - I don't know what the "package" concept is, but shouldn't the description (and synopsis in case of fsmonitor) be different for each package, specifying what's in it (text vs graphical UI vs fsmonitor)?
 - Seeing dependencies here is a bit scary. Currently no-one maintains the opam and dune files. This means no-one is going to keep the dependencies correct. Also, why the duplication with the opam file?
 - The dependencies are probably incorrect (also in the opam file). I don't think dune builds work on OCaml < 4.08 (they only work with `make`).

olafhering:
The existing `dune-project` file lists dune 2.3 as minimal version. This version of dune in turn requires at least OCaml 4.07. In the context of a local opam workspace this also defines the dependencies for opam packages which unison depends on, such as `lablgtk`. However, I think opam got a mode where one can build the latest dune binary and reuse this binary to work with old OCaml versions in a different workspace. I'm not an user of opam, so no idea how it works in practice.

The exact values of package dependencies can certainly be refined at some point, right now they just repeat what is already in `unison.opam`.

I will try to find a proper description for fsmonitor.

olafhering:
This change preserves the status-quo of package versions. AFAIK they are only relevant within opam-repository anyway.

Essentially this PR offers a way to build selected binaries via dune, as explained in the commit message.

In the future it can be used as a vehicle to generate opam files, for submission to the opam repository.

I'm unsure what to put into the maintainer field. After all it should not contain the one who maintains the opam file, but who maintains the code. But also here I can preserve the state and copy the string from the opam file.

dune-project
```diff
@@ -1,2 +1,77 @@
...
+
+(homepage https://www.cis.upenn.edu/~bcpierce/unison/)
+
+(documentation "https://github.com/bcpierce00/unison/wiki")
+
+(package
+ (name unison)
+ (synopsis "File-synchronization tool for Unix and Windows")
+ (description "
+Text based userinterface for Unison.
```
tleedjarv, line 26: I think correct is "user interface", two words. (ditto below)

dune-project
```diff
@@ -1,2 +1,77 @@
...
+   (>= 2.3))
+  (lablgtk2
+   (>= 2.18.6))))
+
+(package
+ (name unison-fsmonitor)
+ (synopsis "File-synchronization tool for Unix and Windows")
+ (description "
+This helper application monitors file system changes on all given
+(relative to root) paths.
```
tleedjarv, line 65: I don't have a good wording suggestion right now but I think it should be made even more clear that this is an _optional add-on_ to the main application. Perhaps by adding that without this helper the syncing, while fully functional, is done manually, repeated with a set time interval, or initiated externally (cron, for example).

tleedjarv:
> I'm unsure what to put into the maintainer field. After all it should not contain the one who maintains the opam file, but who maintains the code.

Could you please figure out what is correct?

liyishuai:
> [`maintainer: [ <string> ... ]`](https://opam.ocaml.org/doc/Manual.html#opamfield-maintainer) (mandatory): A contact address for the package maintainer (the format `"name <email>"` is allowed).

I'd read it as dev members with write access.

dune-project
```diff
@@ -1,2 +1,77 @@
...
+
+(version dev)
+
+(generate_opam_files false)
+
+(license GPL-3.0-or-later)
+
+(maintainers "maintainers")
+
+(authors "Trevor Jim, Benjamin C. Pierce and J\195\169r\195\180me Vouillon")
```
liyishuai, line 13:
```suggestion
(authors "Trevor Jim" "Benjamin C. Pierce" "J\195\169r\195\180me Vouillon")
```

dune-project
```diff
@@ -1,2 +1,77 @@
 (lang dune 2.3)
+
 (name unison)
+
+(version dev)
+
+(generate_opam_files false)
```
liyishuai, line 7: Set  this to `true` so that `dune build` updates the OPAM files as well.
```suggestion
(generate_opam_files true)
```

dune-project
liyishuai, line 26: Avoid empty line in description.
```suggestion
 (description "Text based userinterface for Unison.
```

dune-project
```diff
@@ -1,2 +1,77 @@
...
+  (ocaml
+   (>= 4.03))
+  (dune
+   (>= 2.3))))
+
+(package
+ (name unison-gui)
+ (synopsis "File-synchronization tool for Unix and Windows")
+ (description "
+Graphical userinterface for Unison.
```
liyishuai, line 44:
```suggestion
 (description "Graphical userinterface for Unison.
```

dune-project
```diff
@@ -1,2 +1,77 @@
...
+  (dune
+   (>= 2.3))
+  (lablgtk2
+   (>= 2.18.6))))
+
+(package
+ (name unison-fsmonitor)
+ (synopsis "File-synchronization tool for Unix and Windows")
+ (description "
+This helper application monitors file system changes on all given
```
liyishuai, line 64:
```suggestion
 (description "This helper application monitors file system changes on all given
```

liyishuai (changes requested):

tleedjarv:
> > [`maintainer: [ <string> ... ]`](https://opam.ocaml.org/doc/Manual.html#opamfield-maintainer) (mandatory): A contact address for the package maintainer (the format `"name <email>"` is allowed).
> 
> I'd read it as dev members with write access.

Hmm, I'd read it as _package_ maintainers, not as software authors. Even more so, considering this:

> [`authors: [ <string> ... ]`](https://opam.ocaml.org/doc/Manual.html#opamfield-authors): a list of strings listing the original authors of the software.

I don't know about opam, but in most (all?) other repositories, the software developers are usually not the package maintainers. Perhaps I'm mistaken and mixing up source and binary repositories. Opam being a source repository, it does make sense for package maintainers to usually be the same as developers (software authors). But it does not make sense to list the authors as package maintainers unless they're actively maintaining the package definition. And this is exactly the case here.

gdt:
Another wrinkle is that many packaging systems maintain their own conrtrol files, and some want to put packaging-system-specific files into the upstream repository.

dune-project
gdt: For now, I'm ok with not automatically overwriting checked-in files.

gdt:
I know there are some unresolved comments, but I just merged something that requires rc cycling, and this seems clearly an improvement, so I'm hitting merge and further changes can be considered after 2.52.0.

PR #661, opened 2022-03-05
Spring cleaning: preferences and CLI options

tleedjarv:
This is an assortment of fixes and improvements to preferences. Not everything is tied strictly together, so could be considered for separate merging.

It looks like a lot of change but it's really not. It's just tediously spread about in many files, and much of it is just mechanic change.

#### Removing preferences

Until now, there was no way to remove preferences. This limitation is because the client always sends the full set of preferences to the server and if the server doesn't know a preference it will complain and break the connection. This issue is fixed by remembering the removed preferences so that the complaining and breaking the connection part can be skipped.

With this new ability, remove the following preferences: `sshversion`, `rshcmd`, `rshargs` (all three for obvious reasons).

Anything else that should be removed?

#### Removing rsh protocol

With the rsh-related preferences removed, just go ahead and remove the `rsh` protocol completely. For obvious reasons. While at it, clarify in the manual that this is not the end of the world and one can use the `sshcmd` option to use any remote shell program they want.

#### Deprecating preferences

Closely related to removing preferences, it would be good to have a mechanism to warn and prepare users for upcoming removals. Deprecation does not change the functioning of a preference, it will only nag the user every time a deprecated preference is found in a profile or on the command line. "Deprecated" label is added in the listing of options and in the preference documentation.

With this new ability, mark as deprecated the following preferences: `backups`, `stream`, `halfduplex`.

`backups` seems to be a relic, and while it works (by getting translated into new preferences), it is confusing because it is supposed to be exclusive with the new preferences. Has the time come to prepare for removing it? That's my proposal.

`stream` is probably a preference that's not used by anyone. It is defaulted to true (since it was created), yet its description states that it is experimental. It was probably created as a way out of the new feature if it broke horribly. Clearly it didn't break. No point in having it as a preference any longer.

`halfduplex` is yet another preference that's probably not used by anyone. The description was incorrect (it is not automatically enabled on Windows) and the need for it is questionable.

Anything else that should be deprecated?

#### CLI-only preferences

The preferences have this (very welcome) dual ability that they can be set either in a profile or on the command line. This has caused the implementation to register all command line options as preferences, making pure command line options available to be included in profiles. This is not desired as some options never make sense in a profile and also should not be sent to a server. Add the ability to specify CLI-only preferences that are never sent to server and that will cause an error if included in a profile.

Not included in this PR, but if desired a note "command line only" could be added to such preferences in the listing of options, just like with "deprecated".

With this new ability, mark as CLI only the following preferences: `version`, `doc`, `prefsdocs`, `server`, `socket`, `listen` (and alias `host`), `ui`, `selftest`, `testserver`, `i`, `dumparchives`.

#### Preference categories

(Read the commit message for description.) I came up with and have proposed some categories but that is of course up for discussion and those could be changed any time with ease. The number of categories, their exact topics and titles, all is completely free to change. Check `unison -help` to see the new output with categories. The same output is also included in the manual.

Having real type-checked categories makes it possible to remove any reliance on magic numbers (like looking for character '!' in the beginning of doc strings, and other hacks like that) and to filter preferences based on category.

#### Error handling

The error handling and reporting from profile loading and parsing was quite terse. Not to mention incorrect. Lines in a file are parsed in reverse, from bottom to top; but the line numbers were still counted from 1 (so, 1 is actually the _last_ line of the file). Fixed it and improved the error messages a lot. Now also the chain of includes is shown, not just a random error message (but where did it come from??).

#### GUI (gtk only) changes

CLI-only preferences (see above) are not shown in the GUI profile editor.

Currently all the preferences were listed in one long alphabetical list. That is pretty unusable. To make it a bit more manageable, the "basic preferences" were displayed separately and the user could turn on "all preferences".

Since the preferences now have categories (see above), make the list in GUI a tree list, with group levels and topic levels. I don't know if this is more user-friendly, but I hope so. Feedback is welcome.

There are currently two separate commits for this, the first one introduces the tree but keeps the "all preferences" switch. The second one removes the switch. I believe the second ocmmit is the correct final state, but I wanted to keep commits separate for now just so anyone interested can build the GUI with either commit and compare how it looks like.


gdt:
I intend to merge this just after 2.52.0 is released.  Reviews appreciated.

bcpierce00:
Nice.

tleedjarv:
Rebased. Also deprecated `halfduplex` preference.

jirogoro:
I hope `halfduplex` not to be deprecated since that option is useful for WIndows clients to avoid random deadlocks.
https://stackoverflow.com/questions/9764504/unison-sync-between-windows-linux-hangs-randomly-during-transfer

tleedjarv:
`halfduplex` is not going to be removed any time soon, don't worry. But I wonder if it's still needed with the latest versions (latest Unison version compiled with the latest OCaml version). Those issue reports are quite old.

`halfduplex` is marked deprecated precisely to hear from users who legitimately need it.

gdt:
Comments on stackoverflow don't count as bug reports :-)  I can't quickly find a bug that explains this - but we have a lot, and that's part of  why I'm pushing to close invalid bugs, no longer valid bugs, and things that are not bugs.

But seriously, if you can reproduce a bug with the latest release, please file a bug with the details.  A comment here will be effectively lost.

PR #665, opened 2022-03-07
adjust enabled_if in fsmonitor to match all Linux targets

### LINKED ISSUE #657: %system can not be used to detect the system
[This discussion](https://discuss.ocaml.org/t/a-zoo-of-values-for-system/8525) revealed `(enabled_if (= %{system} "linux"))` can not give the expected result. Specifically, building works for x86 and aarch64, but not ppc:

````
[   37s]      ; architecture = "arm64"
[   37s]      ; model = "default"
[   37s]      ; int_size = 63
[   37s]      ; word_size = 64
[   37s]      ; system = "linux"
````
````
[   44s]      ; architecture = "power"
[   44s]      ; model = "ppc64le"
[   44s]      ; int_size = 63
[   44s]      ; word_size = 64
[   44s]      ; system = "elf"
````

olafhering:
"system" is a synthetic value which is constructed in
ocaml.git/configure, based on the configure --host=triple option.

The current single pattern matches just x86 and arm64 Linux targets.
Extend the list to cover all other targets as well.

Fixes #657

Signed-off-by: Olaf Hering <olaf@aepfle.de>

### DIFF
--- a/src/fsmonitor/linux/dune
+++ b/src/fsmonitor/linux/dune
@@ -3,7 +3,13 @@
 (library
  (name fswatcher)
  (wrapped false)
- (enabled_if (= %{system} "linux"))
+ (enabled_if
+  (or
+   (= %{system} "linux")
+   (= %{system} "linux_elf")
+   (= %{system} "elf")
+   (= %{system} "linux_eabihf")
+   (= %{system} "linux_eabi")))
  (modules :standard \ watcher)
  (flags :standard -w -3-27-39)
  (foreign_stubs
@@ -14,7 +20,13 @@
 (executable
  (name watcher)
  (public_name unison-fsmonitor)
- (enabled_if (= %{system} "linux"))
+ (enabled_if
+  (or
+   (= %{system} "linux")
+   (= %{system} "linux_elf")
+   (= %{system} "elf")
+   (= %{system} "linux_eabihf")
+   (= %{system} "linux_eabi")))
  (modules watcher)
  (flags :standard -w -27)
  (libraries fswatcher))


tleedjarv:
This comment is not really related to this change but the PR kind of makes it very visible. What is the reason for having both the library and the executable definitions? Wouldn't it be simpler and cleaner (and less error prone) to remove the library definition and move the foreign_stubs and libraries down into executable?

gdt:
A fair question, but if this is to make 2.52.0, adding targets seems acceptable (but also could be deferred) and refactoring is too scary at rc time.   (I am happy to release a 2.52.1 with all accumulated fixes before the disruptive changes start for 2.53.   I just don't want to keep delaying 2.52.0 as there will be a worthy fix every week, or at least I hope so.)

olafhering:
I briefly tried to merge `executable` and `library` into a single `executable` rule. As a result the inotify stubs are not found during build. Maybe there is a way to have an `executable` rule which also contains stubs, I have to read the dune documentation carefully.

gdt:
As I see it: dune is not a project-supported build method, and I am inclined to hit merge on PRs to improve dune stuff that seem sensible and not to worry about it otherwise.    So @olafhering totally up to you to submit any such changes or not.

olafhering:
I will try to finish the `dune-project` PR today. In addition I will try to handle Windows and Solaris with dune as well in separate PRs, now that I know how to deal with `enabled_if`.

gdt:
ok, but no rush because I don't want to merge things that aren't fixes for regressions and aren't really tiny before release.

Does dune really need to be done per operating system?  (That seems like a major regression in build systems, but I am pretty clueless about dune so far.)

olafhering:
Certainly not. But OS specific things need a condition, which need to be handled somehow.

PR #566, opened 2021-07-14
GTK 3

tleedjarv:
After the relative simplicity of #557, I thought to myself, why not just try compiling with lablgtk3 and see what happens. Turns out it wasn't so difficult, and here's the result. ~~(It's based on #557, the first commit belongs to that PR -- keep that in mind when reviewing.)~~ Rebased now.

As previously, I don't have any GTK knowledge so this is more a mechanical change just to get it compiling.

### The results

- It seems to work. Functionality seems to be all there but there may be some visual quirks.
- ~~I also briefly tested on Windows and it seems to work as long as you have roots in the profile or provided on command line. Yet, since GTK is not properly installed then stock icons are missing and file/dir chooser makes it crash with error message that GSettings schemas are missing. I assume this is only a problem when downloading the CI binaries. I don't have any Windows knowledge nor Windows dev or build environment, so can't do anything about it. Perhaps @rivy you can work your magic like you did with the gtk2 GUI GHA packaging?~~
- I don't have access to macOS to test.
- I did not attempt to keep gtk2 and gtk3 GUI in parallel. This is a wholesale change from gtk2 to gtk3.

### Known deficiencies

(I think all but the first one are acceptable and may very well remain so permanently.)

- ~~Many widgets, most notably the progress bar and any text views (documentation, diff, sync summary details) have incorrect size (too small). I don't know whether this is a bug in Unison code, in lablgtk3 or in GTK.~~
- Icons in menus are gone. (GTK manual tells you that this is how it should be.) May just as well leave it like that because I think those using standard themes did not see any icons in menus anyway.
- ~~Option menu has been removed in GTK (was used in documentation window). I currently replaced it with a menu bar, which works nearly identically to the option menu. Perhaps a better solution could be proposed here.~~ Replaced with a notebook.
- ~~The transfer rate diagram is gone from the statistics window (statistics in text form are still available). Perhaps that's ok and it could be officially removed for good.~~

Closes #358
Closes #69
Closes #254

## What next?

This patch can be merged as it is; by itself it works. ~~There are some related changes/fixes needed like the Windows packaging issue and widget sizes but they don't need to be included in this PR to make it complete.~~

I am not a frequent GUI user myself so I have probably tested only a fraction of all functionality available in the GUI. And I probably wouldn't notice if something is visually amiss. This is a call for testers/helpers.

[...]

PR #671, opened 2022-03-08
polish dune-project

liyishuai:
Follow up #658, addressing some review comments.

Now `dune build` generates the `.opam` files, avoiding potential syntax errors caused by handwriting.

Better check in the generated `.opam` files to the repository, since:
1. `dune-project` is not updated frequently
2. Hosting `.opam` files allows `opam publish` to release to OPAM more automatically.


gdt:
Thank you for your submission, and I just now saw #674.   After 2.52.0 I will have to figure out how these two relate, but if you and @olafhering could end up with a consensus set of changes that would be great.

liyishuai:
I'm releasing control of this PR branch. Please (force-)push at will.

gdt:
Thanks.   I  will look at this after release and afterwards if you want to sort through changes and rebase I will consider, or if you think whatever I do is good enough that's fine too.

gdt:
Release made.  Please either rebase on top of current  master or close, if everything was in #674.   (I'm trying to avoid any work I can as I feel like my time on unison is a bottleneck.)

gdt:
I don't want to change to generated opam files without a proper discussion about why or why not, and I think that belongs on unison-hackers@.    And, if we do  change to generated, and the decision is made to have generated files checked in (like strings.ml is now), I would like commits separated into  logical changes and regen operations.   I'm not asking you to do all this work; to me the first thing is agreement on what we ought to do and why.   I don't know enough about dune/opam to have a strong opinion yet.

If you want to just have the lablgtk min version change, that seems ok.   With any luck that is the version that the build instructions say.

liyishuai:
I'm still proposing to generate OPAM files from `dune-project` and check them in. This is actually two questions:
1. Q: Write or generate OPAM files?
	A: Generating them from `dune-project` helps keeping them synchronized.
2. Q: Check in OPAM files or ignore them?
    A: [opam-publish](https://github.com/ocaml-opam/opam-publish) requires OPAM files hosted in the repo. If we publish to OPAM manually, then there's indeed no need to check them in, and the current `unison.opam` should be removed to avoid confusion.

tleedjarv:
A few random questions.

What does `(version dev)` mean? Should it be something else (or just removed)?

Should `(wrapped_executables false)` be added?

olafhering:
`(version )` is supposed to carry the version of the project. This variable can be referenced in other dune files, if needed.
I think changing it should be wired up to the release process. In case a new version gets tagged, also this file should be changed. No idea if the release process is documented.

But at this point the value, or even the existence, does not matter because nothing consumes the version. Also, unison is no library, which means the version will not be part of the ocamlfind `META` files.

I have to check what `wrapped_executables` actually does.

gdt:
The requested discussion on unison-hackers has not occurred, and there's an argument against it above, so this PR is on hold, and I'm inclined to close it after 30 days of the feedback label.

liyishuai:
I haven’t received post access to that mailing list, and would prefer launching the discussion after depositing my dissertation in May.

liyishuai:
Another for generating `unison.opam` from `dune-project`: #566 updated the OPAM file and left `dune-project` untouched, making these files inconsistent on dependencies: https://github.com/bcpierce00/unison/blob/2a73c1fc781d2b7242b2820ec680abbff3545d83/dune-project#L52-L57 https://github.com/bcpierce00/unison/blob/2a73c1fc781d2b7242b2820ec680abbff3545d83/unison.opam#L13-L18

gdt:
I see your point, but this is a cultural battle about which file is the one that is primary, and seems to be basically a tussle between two packaging systems.   Unison is an outsider to this tussle, and I am inclined not to wade into it.

gdt:
I understand that, but I have not accepted that goal, so far.  I'm not opposed; it just seems like a more complicated issue.

liyishuai:
Dune is not yet a packaging system, but a build system that allows exporting package descriptions. A dune-based project has its standardized build procedure, so writing `dune-project` and `unison.opam` in parallel poses confusion that which file to trust.

gdt:
This PR is still pending a discussion on unison-hackers sorting out all the issues.

liyishuai:
I've sent an email to unison-hackers on June 24. Has it been distributed or should I resend it?

gdt:
You sent a very brief note, but it did not really explain the issues enough for someone not an opam expert to understand.  I've seen comments that we shouldn't remove opam, and comments that we should because it's now generated, and comments that seem to say that dune and opam are sort of the same thing, and that they are different.

It might help to write the argument for why the change is right into the commit message, vs just saying 'polish', and to rebase as a lot as hit master.   I expect the rebase is trivial, but it takes a potential hiccup off the table.

liyishuai:
I've rebased the commit on top of the current master branch.

I have no idea of how to write the commit message, as I don't know what audience it is for. "Someone not an opam expert" should not be in charge of reviewing the changes here, and I lack the ability to teach OPAM. These changes are proposed as-is. I've granted the maintainers (force-)push access to this branch to change the commit message and/or contents.

olafhering:
I'm not sure if it was already discussed in the context of unison: why are the `build` rules added? I remember most other projects removed it in the recent past from their `dune-project` and `opam` files (ocaml/opam-repository#14266).

Did you add it intentionally?

liyishuai:
The `build` rules were introduced in #698. I've made 7fa617368261deb12553a00c673e55526dbf6b26 a separate commit that we can either squash or drop.

tleedjarv:
Feel free to remove the `build` rules. I don't use opam myself, so trust your knowledge on the matter. That said, it seems to me that the `build` rules were removed in opam repository as a workaround and there never was a proper fix to retire this workaround. Also, probably could be a different approach for libraries and applications, as you mentioned.

unison.opam
```diff
@@ -1,25 +1,38 @@
...
 ]
-synopsis: "File-synchronization tool for Unix and Windows"
-description: """
-Unison is a file-synchronization tool for Unix and Windows. It allows
-two replicas of a collection of files and directories to be stored on
-different hosts (or different disks on the same host), modified
-separately, and then brought up to date by propagating the changes in
-each replica to the other."""
+build: [
+  ["dune" "subst"] {pinned}
```
kit-ty-kate, line 25: This is a bug in the opam file autogeneration fixed in dune 2.7 (see https://github.com/ocaml/dune/pull/3647). I would suggest to bump to at least `(lang dune 2.7)`.
```suggestion
  ["dune" "subst"] {dev}
```

gdt:
Please see https://lists.seas.upenn.edu/pipermail/unison-hackers/2022-November/002061.html and followup on the list.

liyishuai:
I've made the following changes based on the discussions on unison-hackers:
- Update the `maintainers` field of OPAM files to unison-hackers.
- Remove the dependency on Dune from the OPAM files.

## No Not Merge Until:
- The `depends` and `build` rules are fixed in accordance with the Makefile; and
- The OPAM files are tested on CI.

PR #675, opened 2022-03-10
README, NEWS updates for upcoming release

tleedjarv:
This PR is intended to be merged just before or on the day of release.

Release dates in NEWS are not mandatory if it is too much a maintenance burden, but I find it annoying when this information is not available.


gdt:
Saturday seems like a good day for the release, so I'll try to hit tag then :-)

PR #682, opened 2022-03-17
Add a `man` page

### LINKED ISSUE #645: Add man page
unison has a manual, but not a man page (written for mandoc/nroff).   Obviously the entire manual contents does not belong, but "man unison" should be more useful than it is now.

tleedjarv:
I always found it a bit unfortunate that `unison` does not have a `man` page. So here's an attempt at fixing this.

The options list and documentation sections are generated. The rest is mostly copied from the manual and README, with slight modifications. Some sections have been written manually. Ideally, the entire page would be generated, but that task I leave for when (if) the manual is rewritten in another format.

There is a new `make` target: "manpage" and it is also included in the "docs" target. There is no installation for the man page (just because this in general is something that awaits Makefile rewrite).

As always, collaborators are welcome and much expected. To review the entire man page, you have to run `make manpage` (or `make` within `man/`) which will generate `man/unison.1`. It is also possible to revew everything except the options list and documentation in `man/unison.1.in`.

_All_ feedback is welcome. Change the structure, styling, content, you name it.

Closes #645

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,8 @@ doc/unison-manual.pdf
 doc/unison-manual.ps
 doc/unison-manual.txt
 doc/unisonversion.tex
+man/*.tmp
+man/unison.1
 
 _build
 .merlin
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,10 @@ src:
 docs:
 	$(MAKE) -C src UISTYLE=text
 	$(MAKE) -C doc
+	$(MAKE) -C man
+
+manpage:
+	$(MAKE) -C man
 
 include src/Makefile.OCaml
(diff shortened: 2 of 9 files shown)


tleedjarv:
One small update. Use the `Ic` macro for preferences (previously `Cm`).

tleedjarv:
Pushed again. Shortened the introduction text.

gdt:
I didn't really study this, but it's obviously 99.9% right, and easy to adjust if not.  The only hesitation I have is that manpages should be built and installed by default, but as I said earlier, no need to couple that as right now no manpage is built, so not a regression, and it's a larger can of worms

PR #685, opened 2022-03-18
Cleanup: Remove superfluous .depend files

tleedjarv:
This does not change anything in the build process. It only removes `.depend` file placeholders that were required due to a chain of includes from makefiles in different directories. The real and very much needed `src/.depend` remains untouched.


gdt:
Are you implying that the makefiles statically contain all dependencies, and thus unison does not have and does not need any kind of "make depend"?   And therefore that this is just leftover bits from an earlier make depend world?  Or something else? (I can't tell from the PR why this is ok.)

tleedjarv:
`src/.depend` is still there and is where the real dependencies are. Other files where just dummy placeholders because `src/Makefile.OCaml` included `.depend` and other makefiles in turn included `src/Makefile.OCaml` which caused make to look for a `.depend` in every directory.

I changed the `include` to `-include` and now all those dummy placeholders can go away.

Issue #588, opened 2021-09-08
Support DOS device paths "\\?\Volume{GUID}" & UNC paths in Windows


ropery:
In Windows, a volume mounted at `X:` can be addressed in a format called “[DOS device paths](https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats#dos-device-paths)”, like these:

    \\?\X:
    \\?\Volume{b75e2c83-0000-0000-0000-602f00000000}\

The `\\?\Volume{GUID}\` form is useful, because it allows us to reliably identify a volume regardless of its drive letter.

I use Unison on Windows to keep removable disks in sync. For now, I’m using this in my profile:

    root = Y:\
    root = Z:\

I would like to be able to address them by GUID instead:

    root = \\?\Volume{b75e2c83-0000-0000-0000-602f00000000}\
    root = \\?\Volume{2c83b75e-0000-0000-0000-2f6000000000}\

Why is this desirable? Because although we can [assign a “permanent” drive letter](https://docs.microsoft.com/en-us/windows-server/storage/disk-management/change-a-drive-letter) to removable volumes, the association is stored in the registry of the particular Windows system and therefore open to change, etc.; also, this assigned letter is not _guaranteed_ to this volume only: it gets the next unused letter if the assigned letter is in use. With GUID, however, we can be certain that we’re addressing the correct volume.

A further consideration if this gets implemented: in the GUI, displaying the roots (replicas) as GUID would be very human-unfriendly. Instead, it should display the filesystem Lable and/or the drive letter, e.g., `Seagate1 (X:)`; or maybe allow customization, via a syntax like `root = <root> -> <name>`.

PS For example, [TrueCrypt](https://superuser.com/a/465741) supports accessing a disk drive using volume ID instead of a drive letter in Windows.
PS [Example: get drive letter from volume GUID using `wmic`](https://superuser.com/a/1163674).

PPS Currently, trying to use `root = \\?\Volume{GUID}\` results in `Fatal Error: Cannot find canonical name of root directory //?/Volume{GUID}/ (Invalid argument)`. (Notice that `\`s are changed to `/`s.)

https://github.com/bcpierce00/unison/blob/470d36a47681e05f55802134eae9f80bf29a7798/src/fspath.ml#L257

gdt:
[...]

PR #687, opened 2022-03-22
Fix mangling of Windows \\?\ paths

tleedjarv:
This is a bugfix part of #588. Note that there is more to #588 and this PR makes no attempt at covering any of that.

I have done some very brief testing. I have not tested the GUI. **Independent success confirmations must be received from Windows users.** (See #588 for examples of how to test.)


tleedjarv:
More comprehensive fix is included in #757.

PR #689, opened 2022-03-29
Fix Windows build on OCaml 4.14

tleedjarv:
Windows build got broken by a change in OCaml 4.14.

Since older OCaml versions are soon to be removed from the CI anyway, might as well add additional versions now to detect build errors immediately.


gdt:
It seems like we should have an open bug that the Windows code is using a private API, but maybe ocaml is just messy like that.

tleedjarv:
> It seems like we should have an open bug that the Windows code is using a private API, but maybe ocaml is just messy like that.

No, it's Win API that's messy like that :)

PR #691, opened 2022-03-29
Add RPC over Unix domain sockets tests in GHA

tleedjarv:
Fix Unix domain sockets on Windows (see commit message) and add RPC tests over local socket to GHA.

Best reviewed commit by commit.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -209,6 +209,7 @@ jobs:
     - run: opam exec -- make test
 
     - name: Run self-tests over RPC
+      if: runner.os == 'Windows' && !contains(matrix.job.ocaml-version, '4.14')
       shell: bash
       run: |
         # Separate backup dir must be set for server instance so that the central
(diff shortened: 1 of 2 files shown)


gdt:
Is this disabling SO_REUSEADDR on other than Windows?  If so, that seems like an unintended side effect, or at least an explained one.  The second two commits look ok to me.

tleedjarv:
> Is this disabling SO_REUSEADDR on other than Windows?

Yes. Unix domain sockets don't have SO_REUSEADDR anyway but apparently POSIX-ish systems just ignore it. On Windows, `setsockopt` complains loudly. This fix actually applies to all platforms, even though other systems seem to work fine without it. This should not change anything on other platforms; the two `setsockopt` calls should have been noop-s.

gdt:
Could you put that explanation in the commit message?   I'm just paranoid about my future self trying to understand past changes a few years from now.

tleedjarv:
> Could you put that explanation in the commit message?

Sure. Updated now.

gdt:
Thanks for the commit message update.

PR #699, opened 2022-04-15
GHA: Add RPC ABI-check

tleedjarv:
Test the RPC ABI stability and backwards compatibility by executing self-tests between code from tip of the current branch and previous (compatible) releases. This is a minimum smoke test only. More details in the commit message.

This is a franken-workflow but that's mainly due to patches included directly in the workflow definition. I'm not going to change that; these patches do not belong in the tree unless the previous releases are "officially" patched and released (and that's unlikely to happen).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -336,6 +336,323 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+
+  rpc_abicheck:
+    needs: build
+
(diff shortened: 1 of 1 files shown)


tleedjarv:
Build fails on Windows are not related to this PR. Seems to be an opam issue, cause unknown.

tleedjarv:
The cause of Windows fails seems to be the new git version just released to mitigate the latest CVE. Hopefully this will be resolved quickly in `setup-ocaml`.

PR #720, opened 2022-05-11
Drop support for OCaml compilers older than 4.08

### LINKED ISSUE #715: Remove all accomodations for ocaml < 4.08
PRs for this should not be merged until 7 days after 2.52.1, and we decide there are no problems with 2.52.1 that are so serious as to warrant a 2.52.2.

Besides cleaning up, this is an enabling step towards gtk3.

tleedjarv:
Closes #715

Note that support for older compiler versions is not inherently removed (yet). It's just the special accommodations to make the code readily compile with older compilers that are removed.


gdt:
I skimmed and looks fine to me.  Waiting until Monday for 7-day timer to expire.

PR #724, opened 2022-05-13
Makefile cleanup

gdt:
These commits are minor makefile cleanup:
  - pruning code that has not been used in a long time (old release process)
  - typos
  - add a \todo marker about non-portable usage
  - point to LICENSE

I believe these are safe changes ready to merge now.  (Much more is needed, but is harder.)


src/Makefile.OCaml
```diff
@@ -364,6 +365,8 @@ ifeq ($(OSARCH), OpenBSD)
     OCAMLDOT=true
   endif
 else
+  # \todo POSIX does not specify a -t option to type, or the presence
+  # of bash.
   ifeq ($(shell echo type -t ocamldot | bash), file)
```
tleedjarv, line 370: I see two possible ways here. First, get rid of everything related to ocamldot? Alternatively, remove all the logic for setting `OCAMLDOT=true` and replace the `ifdef OCAMLDOT` below with `ifneq ($(shell command -v ocamldot),)` (which is the POSIX way).

gdt:
And there's: create the ocamldot input file with a reasonable name intended to stay, make sure it's removed in clean, and then just run ocamldot and if the user gets not found, so be it.   Don't even check for it.

src/Makefile.OCaml
gdt: Typed in wrong box earlier :-(    My intent is to merge this with the \todo, thus leaving cleaning up the ocamldot stanza as future work.  This PR is intentionally limited to comments and removing content that isn't used.

PR #727, opened 2022-05-14
Prune unportable usage in searching for dot

gdt:
(To be rebased after #724 is merged.)

Remove unportable usage (type -t, bash) in searching for ocamldot, and just run it anyway in make depend, but with failures ignored.   This removes complexity at the expense of a few error lines for people that run make depend on systems without ocamldot.


tleedjarv:
I wonder if this code is really needed. Is there a known user or anybody who will likely use it in future? I'd like to remove it completely or move it to a separate script that can be invoked manually.

gdt:
Good point about separating.   I just changed it to be in its own target, and put the generated file in .gitignore.  So while it may be unnecessary, it is at least not causing trouble, and we can revisit once everything else is cleaned up.

src/Makefile.OCaml
```diff
@@ -342,38 +342,22 @@ system.cmx fspath.cmx fs.cmx: system/$(SYSTEM)/system_impl.cmx
...
-	echo '{ rank = same; "Uarg"; "Prefs"; }'>>dot.tmp
-	ocamldot .depend | tail -n +2 >> dot.tmp
-	-dot -Tps -o DEPENDENCIES.ps dot.tmp
-endif
+           >>.depend.dot.tmp
+	echo '{ rank = same; "Tree"; "Safelist"; }'>>.depend.dot.tmp
+	echo '{ rank = same; "Uarg"; "Prefs"; }'>>.depend.dot.tmp
+	echo 'Attempting to use ocamldot and dot to produce a dependency graph:'
+	-ocamldot .depend | tail -n +2 >> .depend.dot.tmp
+	-dot -Tps -o DEPENDENCIES.ps .depend.dot.tmp
```
tleedjarv, line 360: Forgot that `DEPENDENCIES.ps` should also be added in `.gitignore`.

src/Makefile.OCaml
```diff
@@ -476,6 +460,7 @@ clean::
 	-$(RM) system/generic/*.cm[iox] system/generic/*.{o,obj} system/generic/*~
 	-$(RM) system/win/*.cm[iox] system/win/*.{o,obj} system/win/*~
 	-$(RM) fsmonitor/*.cm[iox] fsmonitor/*.{o,obj}
+	-$(RM) .depend.dot.tmp
```
tleedjarv, line 463: No need, as there is `*.tmp` already. But `DEPENDENCIES.ps` should be added here.

src/Makefile.OCaml
```diff
@@ -342,38 +342,22 @@ system.cmx fspath.cmx fs.cmx: system/$(SYSTEM)/system_impl.cmx
...
-
-# Rebuild dependencies (must be invoked manually)
-.PHONY: depend
+.PHONY: depend dependgraph
 depend::
 	ocamldep $(DEP_INCLFLAGS) *.mli *.ml */*.ml */*.mli */*/*.ml */*/*.mli > .depend
-ifdef OCAMLDOT
-	echo 'digraph G {' > dot.tmp
+
+dependgraph::
```
tleedjarv, line 349:
```suggestion
dependgraph: depend
```

tleedjarv: Or perhaps `dependgraph: .depend`. I don't know which way is better.

src/Makefile.OCaml
gdt: *.tmp does not match .foo.tmp

gdt:
merged with your comments addressed

src/Makefile.OCaml
tleedjarv: > *.tmp does not match .foo.tmp

You are right, of course. It should be the other way around, the tmp file name does not need to begin with a dot.

src/Makefile.OCaml
```diff
@@ -476,6 +460,7 @@ clean::
 	-$(RM) system/generic/*.cm[iox] system/generic/*.{o,obj} system/generic/*~
 	-$(RM) system/win/*.cm[iox] system/win/*.{o,obj} system/win/*~
 	-$(RM) fsmonitor/*.cm[iox] fsmonitor/*.{o,obj}
+	-$(RM) .depend.dot.tmp DEPNDEENCIES.ps
```
tleedjarv, line 463: Merged too soon. There's a typo here.

PR #733, opened 2022-05-19
Makefile: Simplify and prune

gdt:
This
  - removes content that doesn't have a rationale to stay
  - drops :: on rules that don't need them
  - avoids specifying UISTYLE on builds
  - avoids cleaning things that aren't built

Draft for discussion; this may be too aggressive, and we'll see the CI results.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,23 +1,17 @@
 # Unison file synchronizer: Makefile
 # See LICENSE for terms.
 
-.PHONY: all src
-
-default: text
-
-text:
-	$(MAKE) -C src UISTYLE=text
-
-test:
-	./src/unison -ui text -selftest
+.PHONY: all src docs manpage test depend clean install
 
 all: src manpage
 
 src:
 	$(MAKE) -C src
 
--include src/Makefile.ProjectInfo
-
+# It's a wart that docs need "unison" built (vs some docs utility).
+# Having docs build src/unison points out that UISTYLE is a bug; either
+# docs might build the GUI (not wanted as too heavy for docs use), or
+# whatever is built might not be rebuilt later.
 docs:
 	$(MAKE) -C src UISTYLE=text
 	$(MAKE) -C doc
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,47 +1,33 @@
...
 all: src manpage
 
 src:
 	$(MAKE) -C src
 
--include src/Makefile.ProjectInfo
-
-docs:
-	$(MAKE) -C src UISTYLE=text
+docs: src
```
tleedjarv, line 11: I'd prefer not. Dependency on `text` (`UISTYLE=text`) is important here because we don't want to build the GUI just to be able to generate docs (and we most certainly don't want any GUI dependencies pulled in). I think this remains important even if in future the build system is changed to by default always build the GUI.

Makefile
```diff
@@ -1,47 +1,33 @@
...
 
 manpage:
 	$(MAKE) -C man
 
-depend::
+test:
+	./src/unison -ui text -selftest
+
+depend:
 	$(MAKE) -C src depend
```
tleedjarv, line 22: I tentatively propose to remove `depend` completely from toplevel Makefile.

gdt: As I see it, the fact that there is a makefile in src is a mere implementation detail, and the normal targets should work at top level.  There's also no good reason why docs couldn't end up with a depend; they are complicated enough.

Makefile
gdt: I guess maybe we should fix things in a different order.   Right now, I think that when src/unison is built, and then a requests to build with a different UISTYLE is made, nothing happens because it is up to date.   Building docs shouldn't affect what happens in src, and it's an artifact of unison that building the docs requires running the binary.   But I do agree about the more general GUI issue.

Perhaps we should set this aside,  resovle the "UISTYLE exists" bug and then clean up.  But I think I'll just revert this part of the change, and the rest is probably ok.

gdt: I have force-pushed without the hunk that you objected to (leaving it as part of the later UISTYLE fix).

tleedjarv: > Right now, I think that when src/unison is built, and then a requests to build with a different UISTYLE is made, nothing happens because it is up to date.

That's not the case because depencencies will change with UISTYLE. (At least works for me and works in the CI.)

> resovle the "UISTYLE exists" bug

I don't consider this a bug. This might still be kept as an implementation detail, just hidden from users. (Think a single make target for users that executes one or more builds with different UISTYLEs to produce multiple binaries, with all the shared dependencies compiled once. This will work already today and is effectively what CI is doing. This is for me what the real bug is about.)

Makefile
tleedjarv: I agree that the end goal is not to have a makefile in src.

gdt: I didn't say that was a goal.  Just that someone should be able to invoke make at top level to do the normal things and that should work.  But, taking depend out of top-level is out of scope for this PR, so we can worry about that later.

Makefile
gdt: Certainly we can build multiple styles.   What is a bug is using a variable that controls things in a way which does not play nice with dependencies.   Once there is unison and unison-gtk, there should be no need for a UISTYLE variable; one can just build one or both of the desired binaries.   And yes, reusing the things in common.    I have observed bad behavior and will add that to the UISTYLE ticket so it isn't lost here.

PR #734, opened 2022-05-19
Simplify src/Makefile

gdt:
This does three things:
  - removes the demo, as proposed on unison-users
  - Removes rules to generate INSTALL, because we don't generate it
  - Simplifies the clean target by having fewer :: rules, when there is no good reason to have them split.


tleedjarv (approved):

gdt:
Status: a positive review and no objections.  Waiting until 24 May to merge to allow call for objections on mailing list to time out.

PR #736, opened 2022-05-22
Prune old test targets

gdt:
These rely on a profile that isn't checked in.   If they are still needed, they should be turned into scripts that can be run during CI.


gdt:
Another open PR included by mistake; I'd rather wait and do this again than juggle conflicts.

PR #746, opened 2022-05-28
NEWS.md: Add 2.53.0 section

gdt:
Note that this does not imply 2.53.0 is imminent; it is merely
housekeeping as NEWS should always be up to date along the branch.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,12 +2,26 @@
 
 This file contains a summary of user-visible or important changes, in
 the style of the GNU coding standards.  By user-visible, we include
-changes relevant for those building unison from source.
+changes relevant for those building unison from source.  We omit
+most bugfixes and minor improvements.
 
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -2,13 +2,27 @@
...
 unison uses git history for that, and thus there is no changelog file.
 
+## Changes in 2.53.0
+
+NOT YET RELEASED
+
+  * Change OCaml requirement to >= 4.08
+  * Clean up a variety of unmaintained and unused bits, mainly
+    build-related
+  * Change GUI to use lablgtk3 instead of lablgtk2
```
tleedjarv, line 20: Perhaps only mention GTK. lablgtk is a library name that few users know.

gdt: Good point - I adjusted the wording and force-pushed, but I think it's important to leave lablgtk* present because that is a build dependency, and NEWS is written for an audience including those that build from source.

PR #757, opened 2022-06-15
Reduce the amount of Windows-specific code in the System module

tleedjarv:
### Why?

Windows-specific code is barely maintained and it's not going to get better. 

The code removed in this PR is no longer needed because OCaml's Unix library has provided the same or improved functionality since OCaml 4.06.

### What?

Best reviewed commit by commit.

Commits 1, 2, 4, 5. Use the Unix library and remove as many Win32 API stubs as possible. The unfortunate exception is `stat` (due to OCaml not providing proper `ctime` in Windows). Maybe that'll have to go in the end, too, with ctime not being valuable enough to justify the complex Windows-only code. The usage of Win32 Unicode API (as opposed to Win32 ANSI API) was controlled by the `unicode` preference, but that has done nothing since OCaml 4.06. Remove that (ab)use of the `unicode` preference.

Commits 3, 6, 7, 8. While most of the Win32 API stubs are removed, the System module itself is kept because Windows-only differences can't be fully avoided. But now the System module itself and its usage are greatly simplified.

Commits 9, 10. Finally, remove some dead code and fix some bugs with Windows-only paths.

These changes are not intended to reduce any functionality. Quite the contrary, there are improvements in some areas (like the `\\?\` paths, for example).


gdt:
I skimmed the commits. With passing CI and no comments from anyone else, I'm going to hit merge.

PR #769, opened 2022-07-20
Fix potential fastcheck miss

tleedjarv:
I've stumbled on something I consider a bug. See more details in the commit message.

Summary of the issue is that if mtime has changed but file contents has _not_ changed (and something similar for directories) then mtime is updated directly in the archive. This works as long as mtime is not synced (`times` preference is off). This breaks when mtime is to be synced but for some reason is not synced (propagation failure, for example).

I noticed this when testing xattrs and ACL branches because there the breakage is quite evident, as opposed to just the mtime sync failing.

You can easily check the incorrect behaviour with following script:
```sh
mkdir test{1,2}
touch test1/a
unison -times test{1,2} -batch
touch test1/a
unison -times test{1,2}  ## make sure not to use -batch
## SKIP the change, do not propagate it
unison -times test{1,2}
## mtime change will no longer show up, even though it was never propagated
```
(repeat with this patch applied)

I propose a fix here and believe it to be correct. I've also applied the fix to fastcheck on directories.

### DIFF
--- a/src/update.ml
+++ b/src/update.ml
@@ -1540,7 +1540,7 @@ let rec noChildChange childUpdates =
 (* Check whether the directory contents is different from what is in
    the archive *)
 let directoryCheckContentUnchanged
-      currfspath path info archDesc childUpdates scanInfo =
+      currfspath path info propsChanged archDesc childUpdates scanInfo =
   if
     noChildChange childUpdates
       &&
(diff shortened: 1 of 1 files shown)


gdt:
It strikes me that there are further nuances, because whether *this run* is syncing mtimes or not is separate from whether some other run that uses the archives will sync mtimes.   However that is probably a much larger issue, and this fix is clearly a step in the right direction.

PR #779, opened 2022-08-21
Add WinOS console+GUI hybrid back into build/production

### LINKED ISSUE #778: 🐛(WinOS) `unison` (gui/gtk) has lost all console output (between v2.51.4 and v2.51.5)
The WinOS GUI version of `unison` no longer outputs anything to the console. Unfortunately, this makes using `-ui text`, `-help`, `-version` essentially unusable, and no errors are ever displayed when incorrect argument/option text is supplied.

```shell
>"unison-v2.51.4+ocaml-4.12.0+x86_64.windows\bin\unison-gtk2.exe" -version
unison version 2.51.4 (ocaml 4.12.0)

>"unison-v2.51.5+ocaml-4.12.0+x86_64.windows\bin\unison-gtk2.exe" -version

>"unison-712599e7+ocaml-4.14.0+x86_64.windows\bin\unison-gui.exe" -version
```

In looking through the difference (https://github.com/bcpierce00/unison/compare/v2.51.4...v2.51.5), I suspect the problem is in the new console handling code which is likely not piping output back to the original console.

rivy:
Fixes #778.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -238,7 +238,7 @@ jobs:
       if: ${{ ! ( contains(matrix.job.ocaml-version, '+msvc') || contains(matrix.job.ocaml-version, '+musl') || contains(matrix.job.ocaml-version, '+32bit') ) }}
       run: opam depext --install --verbose --yes lablgtk3 && opam install ocamlfind
 
-    - if: steps.vars.outputs.STATIC != 'true' ## unable to build static gtk for linux or windows/Cygwin MinGW platforms
+    - if: steps.vars.outputs.STATIC != 'true' ## unable to build static gtk/gui
       shell: bash
       run: |
         opam exec -- make src OSTYPE=$OSTYPE UISTYLE=gtk3 STATIC=${{ steps.vars.outputs.STATIC }}
@@ -246,6 +246,19 @@ jobs:
         # * copy only main/first project binary
         project_exe_stem=${PROJECT_EXES%% *}
         cp "src/${project_exe_stem}${{ steps.vars.outputs.EXE_suffix }}" "${{ steps.vars.outputs.PKG_DIR }}/bin/${project_exe_stem}-gui${{ steps.vars.outputs.EXE_suffix }}"
+
+    - name: "Build WinOS text+gui hybrid"
+      if: runner.os == 'Windows' && steps.vars.outputs.STATIC != 'true' ## WinOS, non-static (unable to build static gtk/gui)
+      shell: bash
+      run: |
+        # create and stage text+gui hybrid for Windows
+        # * copy only main/first project binary
+        project_exe_stem=${PROJECT_EXES%% *}
+        # * clean/remove build artifact(s)
+        rm "src/${project_exe_stem}${{ steps.vars.outputs.EXE_suffix }}" ##.or.# opam exec -- make -C src clean #.or.# opam exec -- make clean
+        # * re-create (with hybrid text+gui UI)
+        opam exec -- make src OSTYPE=$OSTYPE UISTYLE=gtk3 UI_WINOS=hybrid STATIC=${{ steps.vars.outputs.STATIC }}
+        cp "src/${project_exe_stem}${{ steps.vars.outputs.EXE_suffix }}" "${{ steps.vars.outputs.PKG_DIR }}/bin/${project_exe_stem}-text+gui${{ steps.vars.outputs.EXE_suffix }}"
 
     - uses: actions/upload-artifact@v2
       with:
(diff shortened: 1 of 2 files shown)


rivy:
Essentially all tests are successful; the only failure was a network timeout failure during preliminary OCaml install/setup.

.github/workflows/CI.yml
```diff
@@ -246,6 +246,16 @@ jobs:
         # * copy only main/first project binary
         project_exe_stem=${PROJECT_EXES%% *}
         cp "src/${project_exe_stem}${{ steps.vars.outputs.EXE_suffix }}" "${{ steps.vars.outputs.PKG_DIR }}/bin/${project_exe_stem}-gui${{ steps.vars.outputs.EXE_suffix }}"
+        # create and stage text+gui hybrid for Windows
+        case '${{ matrix.job.os }}' in windows-*)
+          # * clean/remove build artifact(s)
+          opam exec -- make -C src clean ## .or.# opam exec -- make clean #.or.# rm "src/${project_exe_stem}${{ steps.vars.outputs.EXE_suffix }}"
```
tleedjarv, line 252: I think `rm "src/${project_exe_stem}${{ steps.vars.outputs.EXE_suffix }}"` is plenty sufficient here. Just to avoid the entire rebuild. (The designation as console or GUI application happens at link time, so rebuild is not needed.)

tleedjarv:
In general it looks fine to me. Just one comment (see above) but even that is not too important. It would mainly help with GHA build times.

.github/workflows/CI.yml
rivy: Yes, since it's only a link step change, I thought that myself. But I wasn't completely sure, so I implemented it as a half measure of rebuild. With just the 'src' clean, the recompilation literally completes within seconds and doesn't change the total build time in any real way, so I left it.

I'll change to the `rm ...` and re-push if you think its warranted.

tleedjarv: Thinking a bit more about it, perhaps a more relevant improvement is to make this new code a separate job step guarded by an `if: runner.os == 'Windows'`. I think this makes the build log easier to follow should anything fail. It also makes the scripts simpler (trying to go for a universal build script with GTK on Win is kind of futile at this point, something we have to live with).

rivy:
I pushed a change. Let mw know if that's the direction you meant.

.github/workflows/CI.yml
```diff
@@ -247,6 +247,20 @@ jobs:
...
+      shell: bash
+      run: |
+        # create and stage text+gui hybrid for Windows
+        # * copy only main/first project binary
+        project_exe_stem=${PROJECT_EXES%% *}
+        # * clean/remove build artifact(s)
+        rm "src/${project_exe_stem}${{ steps.vars.outputs.EXE_suffix }}" ##.or.# opam exec -- make -C src clean #.or.# opam exec -- make clean
+        # * re-create (with hybrid text+gui UI)
+        opam exec -- make src OSTYPE=$OSTYPE UISTYLE=gtk3 UI_WINOS=hybrid STATIC=${{ steps.vars.outputs.STATIC }}
+        project_exe_stem=${PROJECT_EXES%% *}
```
tleedjarv, line 261: nit: `project_exe_stem=${PROJECT_EXES%% *}` is duplicated a couple of rows above.

tleedjarv:
This looks good to me (assuming the build completes).

Just one remark: make sure not to change whitespace in the later portions of the `yml` as those are literal patches there any may fail to apply when changed.

gdt:
I'm watching and am fuzzy on the details, but that's ok -- when you two agree it's ready and CI passes I'll hit merge.

rivy:
> Just one remark: make sure not to change whitespace in the later portions of the `yml` as those are literal patches there any may fail to apply when changed.

I've got auto-format on save enabled on my editor, so I have to specifically "save without reformatting". Any, sometimes my hardwired `ctrl-s` fires off without actual my conscious thought. 😄

I'll reformat the PR commit(s), keeping whitespace otherwise intact.

rivy:
All tests passing. No extraneous whitespace changes.
I believe done.
OK to merge?

tleedjarv:
You missed one nit comment (https://github.com/bcpierce00/unison/pull/779#discussion_r956751366), or maybe not :)

Either way, looks good to me.

I'm not too sure about the confusion that the users must have with three exe-s. Then again, that confusion is already there with two exe-s, I guess. It could be simpler for first-time (click-starting) users. It's not in scope of this PR, of course.

.github/workflows/CI.yml
rivy: Removed it; thanks!

rivy:
I didn't see the note (likely, I was posting from an un-updated page).
Anyway, final update with the change is pushed and now in CI.

gdt:
Ci hasn't quite passed but I'm going to merge anyway, as the Windows  CI infrastructure just seems too flaky.

PR #784, opened 2022-08-31
Make GUI-only Windows binary more discoverable

tleedjarv:
Another fix for #778.

The following normally text-only output is now handled differently by the GUI-only binary:
- `-version` output is displayed in a GUI message box.
- `-doc` produces a generic message explaining the GUI-only binary.
- `-ui text` produces a generic message explaining the GUI-only binary.
- `-help` produces a generic message explaining the GUI-only binary.
- command line parsing errors are displayed in a GUI message box.
- GTK init errors are displayed in a GUI message box.

@rivy could you also test and see how this fits into your workflow.

### DIFF
--- a/src/main.ml
+++ b/src/main.ml
@@ -164,6 +164,13 @@ let catch_all f =
   with e ->
     Util.msg "Unison server failed: %s\n" (Uicommon.exn2string e); exit 1;;
 
+let gui_safe_printf fmt =
+  Printf.ksprintf (fun s ->
+    if System.has_stdout ~info:s then Printf.printf "%s" s) fmt
+
+let verify_stdout () =
+  if not (System.has_stdout ~info:"") then exit 37
+
 let init () = begin
   ignore (Gc.set {(Gc.get ()) with Gc.max_overhead = 150});
   (* Make sure exception descriptions include backtraces *)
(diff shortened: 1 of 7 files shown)


acolomb:
My only concern here is about errors involving the GUI system itself, such as when GTK cannot be initialized. I think those are dumped to the console either way, and obviously a GUI window for those is impossible.

Could we just pop up a console window in that last case? I'd assume it's something like GApplication failing to initialize or other error return codes from early GTK functions.

tleedjarv:
> My only concern here is about errors involving the GUI system itself, such as when GTK cannot be initialized. I think those are dumped to the console either way, and obviously a GUI window for those is impossible.
> 
> Could we just pop up a console window in that last case? I'd assume it's something like GApplication failing to initialize or other error return codes from early GTK functions.

Will give it a try. Don't know yet how to test that, though.

I don't think it's going to be too useful for a user because there will be no meaningful error message. `lablgtk` uses the `gtk_init_check()` function https://docs.gtk.org/gtk3/func.init_check.html and this only returns `FALSE` on errors, without actual error message. But it might still be better than just "silently crashing". At least the user knows that gtk init failed, even though it's impossible to tell why.

tleedjarv:
Now GTK init errors are also displayed as a message box.

acolomb:
Alright, that sounds fair. Sorry I didn't have a chance to look at the code changes while traveling.

tleedjarv:
> As for testing GTK errors, you might have luck setting some invalid configuration via environment variables. `GDK_BACKEND` comes to mind for example, which should make it impossible to open any GUI when using an invalid value.

Thanks for the hint @acolomb. This worked (to make GTK init fail). Now, a native Windows message box is displayed with the error. Note that this is strictly an improvement because previously (since always) the error would have been displayed in the console window but the console window would be closed immediately with application exit (unless run from command line) and the user wouldn't see anything.

PR #786, opened 2022-09-03
manual: Enhance description of backups

gdt:
manual: Enhance description of backups
    
Note that the lack of description of mechanisms to clean backup files
other than maxbackups really does mean that there are no other
mechanisms.
    
Note that central backups put backup files on possibly different
filesystems, with possibly different security properties and available
space.

### DIFF
--- a/doc/unison-manual.tex
+++ b/doc/unison-manual.tex
@@ -1685,7 +1685,7 @@ preference.  For example,
 \begin{verbatim}
     backup = Name *
 \end{verbatim}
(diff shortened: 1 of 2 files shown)


tleedjarv:
I think you forgot to re-regenerate strings.ml after your further edits.

gdt:
Actually I forgot that running make at top-level does not regenerate strings.ml.  I reran make in docs and then at top-level and pushed that, so I think it's ok now.  Thanks for catching this.

PR #791, opened 2022-09-09
Small improvements to status display in text UI

tleedjarv:
First, display ETA as hh:mm:ss instead of current mm:ss to improve the progress status during large long-running syncs.
What was previously displayed as `1234:56 ETA` is now displayed as `20:34:56 ETA`.

Second, add a display of transferred bytes. Likewise, to improve the progress status during large syncs.
What was previously displayed as `14%  xx:xx ETA` is now displayed as `14%  (175 MiB of 1.22 GiB)  xx:xx ETA`.

A small bike-shedding moment is the choice of units, MB (base 10) or MiB (base 2). I've currently chosen a hybrid: base 2 for display and base 10 for the thresholds. In other words, 999 B is followed by 1 KiB (rounded) and so on.

### DIFF
--- a/src/ubase/util.ml
+++ b/src/ubase/util.ml
@@ -328,6 +328,23 @@ let percentageOfTotal current total =
 
 let percent2string p = Printf.sprintf "%3d%%" (truncate (max 0. (min 100. p)))
 
+let gib = 1073741824.
+let mib = 1048576.
+let kib = 1024.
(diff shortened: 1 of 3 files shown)


gdt:
I decided not to worry about units, as this seems like a big improvement and we can argue about that separately if anybody wants to, disconnected from achieving the larger win.

PR #792, opened 2022-09-12
Improve GTK performance when clearing the list

tleedjarv:
Large operations on the backing model of the main list in GUI can take extremely long time when done while the model is attached to the list. These operations can take tens of minutes when the list is tens of thousands or hundreds of thousands items long. This happens even with bulk operations like clearing the list.

Make clearing the list almost instantaneous by detaching the model from the list, clearing it, and then re-attaching it to the list.


gdt:
Seems like a higher-level fix might be good, where the gui has a suspend/resume to avoid intermediate processing, but this seems like a safe improvement.

PR #805, opened 2022-10-08
Improve the etags `make` recipe

tleedjarv:
I've noticed that some packaging repositories list `emacs` as a build dependency for `unison` (yes, really!). I don't know how that could have happened but if this recipe had something to do with it then better fix it for good.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -204,7 +204,7 @@ testmerge:
 .PHONY: tags
 
 tags:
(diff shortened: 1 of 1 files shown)


gdt:
It doesn't make sense that this would lead to emacs being a build dependency, but evaluating this on its own merits, command is specified by POSIX and which is not.

Issue #352, opened 2020-06-23
Initiial sync of 650,000 files in text UI can use 5kB per file, exhausting 4G of memory


powertoaster:
server started
Unison failed: Uncaught exception Out of memory
Raised at file "./lwt/lwt.ml", line 135, characters 12-13
Called from file "list.ml", line 73, characters 12-15
Called from file "./lwt/lwt.ml", line 31, characters 2-37
Called from file "./lwt/lwt.ml", line 83, characters 17-46
Called from file "./lwt/win/lwt_unix_impl.ml", line 239, characters 6-40
Called from file "./remote.ml", line 1465, characters 11-57
Called from file "./ubase/util.ml", line 170, characters 6-9
Called from file "./main.ml", line 131, characters 4-9

I have used this for a while now and not having any problems, then today on a new windows server I am getting this.
It may be related to the large number of files I am trying to sync but it is not orders of magnitude larger than the others.

I am doing a socket transfer between two Windows servers:

c:\unison\unison.exe C:\dir\subdir -prefer C:\dir\subdir   socket://1.1.1.1:2222/

gdt:
Please explain what version, what ocaml version, and how big your roots are (#files).

gdt:
See #378

tleedjarv:
@powertoaster You are using a rather old version. Could you test again with latest Unison code (2.51.3rc1 at the moment)?

How much available memory do you have on servers before running Unison? Could you also try to track Unison memory usage while the process is running? The root size of almost 650,000 files should not be any problem at all memory-wise.

tleedjarv:
As I had already built a test setup with 1.5 million files to investigate another issue, I gave it a go over a socket to try to reproduce this issue.

Memory consumption does increase constantly (unlike when testing between folders on the same machine). This seems to suggest that there might be a memory leak somewhere in the remoting code. Still, it is not enough to cause an Out of memory exception, unless you have very little available memory on your Windows servers.

I also tried running Unison with `repeat` (with new updates being constantly created) to see if available memory would eventually be exhausted. Interestingly, that does not happen. Once the maximum allocation has been reached, true, it is not released while Unison is still running, but subsequent syncs appear to be using the same memory and never allocate more.

It currently seems that this issue is the result of having very little available memory on the servers and the real issue is #378. The reporter mentions hitting the problem on a new Windows server, which would explain the change in memory available to Unison.

powertoaster:
[...]

PR #809, opened 2022-10-24
Improved parallel transport mechanism and improved cleanup

### LINKED ISSUE #780: Stack usage grows with number of updated files
Version on server and client: Unison 2.52.1 (ocaml 4.14.0)

**default.prf**
```
auto = true
batch = true
repeat = watch
owner = true
group = true
debug = verbose
silent = true
root = /files
root = ssh://syncer@unison2//files
prefer = /files
```

**Error**
```
 97%  01:46 ETA
 99%  00:45 ETA
Uncaught exception Stack overflow
Raised at Lwt.poll in file "/home/runner/work/unison/unison/src/lwt/lwt.ml", line 126, characters 16-23
Called from Lwt_unix_impl.run in file "/home/runner/work/unison/unison/src/lwt/generic/lwt_unix_impl.ml", line 102, characters 8-23
Called from Uitext.Body.doTransport in file "/home/runner/work/unison/unison/src/uitext.ml", line 880, characters 2-113
Called from Uitext.Body.interactAndPropagateChanges.doit in file "/home/runner/work/unison/unison/src/uitext.ml", line 942, characters 38-66
Called from Uitext.Body.synchronizeOnce in file "/home/runner/work/unison/unison/src/uitext.ml", line 1146, characters 6-50
Called from Uitext.Body.synchronizePathsFromFilesystemWatcher.loop in file "/home/runner/work/unison/unison/src/uitext.ml", line 1184, characters 6-125
Called from Uitext.Body.synchronizeUntilDone in file "/home/runner/work/unison/unison/src/uitext.ml", line 1232, characters 8-47
Called from Uitext.Body.start in file "/home/runner/work/unison/unison/src/uitext.ml", line 1445, characters 21-43
```

No any errors in syslog.

tleedjarv:
This PR includes two features that are logically related. See motivation for a single PR below.

The first commit (Move common update propagation code to Uicommon) is common to both features.

The second commit (Reduce memory usage during update propagation) is the first feature, inspired by #352 and #780. This covers the green path of update propagation.

The remaining commits are the second feature, which is a continuation of improvements introduced in #649 with fatal errors not stopping the GUI and #764. This covers the error path of update propagation.

The motivation and goals of these changes are: improved resource usage, fixing/prevention of resource leaks and making the UI more error-tolerant. See more detailed description in commit messages and in code comments.

**For reviewers:** Reviewing commit-by-commit is the simplest. Ignoring whitespace changes and detecting moves (also ignoring whitespace or indentation changes for move detection) may help.

### Testers needed!

Volunteer testing is very welcome. Regarding fear of data loss/corruption, it is as safe to test as any other Unison version. In fact, there are improvements in error handling code that are expected to result in a cleaner and safer recovery from errors.

**How to test?** Just use Unison as normal. There are no user-visible changes as long as update propagation completes successfully. If you want to go further then you can simulate errors/interruptions during update propagation. For example, kill the server process, break the network connection, suspend the client, kill the ssh process; anything you can come up with, really. This should stop the propagation with a fatal error but should not result in uncaught exceptions or additional errors caused by the cleanup. If you're using the GUI then you should be able to continue using the GUI normally and either rescan or change the profile. Please report any deviations.

Text UI and the GTK GUI have received testing (but please test more!). The macOS native GUI has not had any testing yet.

----

With this merged, I think #352 can be closed. I see memory consumption reduced by > 1 kB and there have been other improvements before this. I think memory consumption of < 2 kB per update (in text UI) is reasonable to expect now. (But it depends _a lot_ on the length of path names and synced properties.)
It will probably fix #780, too, but I'm not going to test it.

Why two features in a single PR? Given the resources available to reviewing and testing, having two (or more) separate PRs would result in _less_ review and testing. The included features make changes in the same process (update propagation), one in the green path and the other in the error path. It makes sense to test these changes together and the risk of masking/hiding errors due to too much change at once is minimal.


gdt:
I'd like to see test reports, but am inclined to just hit merge 2 weeks after 2.53.0 is released if not.

Issue #554, opened 2021-06-23
Make SIGINT be more of a "graceful shutdown" than it is now


bruceg:
When running `unison -repeat watch -watch`, unison will continuously look for new changes (via the watcher) and synchronize them. This all works fine, but I would like a way to interrupt this process between synchronization batches. Currently, the only way I am aware of (using the text UI) is to wait for unison to report "Synchronization complete" and then quickly hit Control-C before it can be notified of changes. I know unison can be interrupted during synchronization, but I would prefer to avoid leaving temporary files around or having to resolve issues with partial synchronizations or leftover lock files.

Is there another way of stopping `unison -repeat` other than killing it? If not, could some mechanism be added?

gdt:
I would ask: when you run that, and decide to stop unison because you need to reboot a computer or something, what do you do?  I wonder if this is really "there is no graceful shutdown in repeat mode" and not particularly about wanting a pause between batches.

bruceg:
It is the former (no graceful shutdown). My current process is to watch the text output and quickly hit Control-C when I see the `Synchronization complete …` or `Nothing to do` line, as there's usually a second or two pause before it starts the loop again.

bruceg:
I had two thoughts for implementation, though I am not familiar with the internals so I cannot comment on the feasibility of either. First would be to use some non-interrupt signal (say SIGHUP) to set an internal flag that is checked before and after synchronizations to trigger a graceful shutdown. The other would be to check for a configurable shutdown file to do similar. Doing a graceful shutdown during a synchronization strikes me as a much more difficult problem to solve.

gdt:
I guess the other question is about graceful shutdown of a sync that doesn't have -repeat.  Really, SIGINT should lead to things being cleaned up and left in a state where if reran nothing bad will happen.

I also wonder if the command-line could read from stdin, and if it sees EOF enter graceful shutdown at the end of the current sync.  Or take q\n etc.

bruceg:
Right, reading something (like EOF) from stdin would work as well and is a good idea, at least for an interactive stop. It's a little harder to set that up to trigger automatically from another program, but it could still be done with a named pipe on stdin. My other suggestions are harder for interactive mode, so this latter one is likely better for the common case.

I agree SIGINT should be safe as well, but I have run into two situations where interrupting unison causes problems.
[...]

PR #810, opened 2022-10-24
Make it possible to stop update propagation by user request

### LINKED ISSUE #812: Interrupting the text UI terminates the child process for `ssh` too early
When interrupting the text UI with Ctrl-C (or any ^C/break in terminal input) then the generated SIGINT will terminate `ssh` before `unison` manages to do cleanup.

The issue is that `unison` will execute `ssh` as a child process. If `unison` is executed from a terminal then the `ssh` child process will be in the same process group and using the same terminal (the same session). A SIGINT generated by the terminal is sent to all processes in the same process group. The end result is that the `ssh` process dies before `unison` manages to do the cleanup with the server. You can witness this by pressing Ctrl-C (with `unison` in the foreground) and then receiving an error "Failed due to lost connection".

I'm not sure what the solution to this is. ~~Creating a new process group for the child process before `exec`'ing ssh is an option. Even if this works, I wonder how portable it is (it's POSIX, but still) and whether this will cause endless unforeseen side-effects.~~ Creating a new pty like is done for ssh by the GUI is another option and probably a better one. This could also work in Windows.
_Edit_: Just remembered that executing `ssh` in a different process group wouldn't work because `ssh` may need input from the terminal.

_Originally posted by @tleedjarv in https://github.com/bcpierce00/unison/issues/810#issuecomment-1292485699_

tleedjarv:
Hot on the heels of #809 here's a quality-of-life improvement for users.

So far, it has been a bit tricky to stop an ongoing update propagation. The only way is basically by killing the process.

This PR makes it possible for the user to request stopping the propagation and also makes the propagation actually responsive to this request. Stopping is orderly and both the client and server processes remain in a consistent state and continue working normally. The only side effect is that temporary files may be left in one or both replicas. (These temp files are harmless and can be used by the next propagation attempt or will be deleted after 30 days if not.)

In GUI, there is a new button. In text UI, stopping is requested by pressing Control-C. Additionally, pressing Control-C multiple times will try to forcefully terminate the process. (As a side note, Control-C any time when _not_ propagating will still terminate the process.)

This is an improvement for #554 but not enough to close it yet. There's more that can be done for the repeat mode.

Also fixes #812.

This PR is based on #809 (review only the last six commits here).

**Testers are welcome, as always!** If you are going to test #809 then you may just as well test only this PR as it includes the entire #809.


tleedjarv:
There is one little issue with ssh and text UI. Not caused by nor made worse by changes in this PR, so I'm not in a hurry to fix it. The issue is that `unison` will execute `ssh` as a child process. If `unison` is executed from a terminal then the `ssh` child process will be in the same process group and using the same terminal (the same session). A SIGINT generated by the terminal is sent to all processes in the same process group. The end result is that the `ssh` process dies before `unison` manages to do the cleanup with the server. You can witness this by pressing Ctrl-C (with `unison` in the foreground) and then receiving an error "Failed due to lost connection".

I'm not sure what the solution to this is. Creating a new process group for the child process before `exec`'ing ssh is an option. Even if this works, I wonder how portable it is (it's POSIX, but still) and whether this will cause endless unforeseen side-effects. Creating a new pty like is done for ssh by the GUI is another option and probably a better one. This could also work in Windows.

Just wanted to mention it here. Fixing this issue will not have any role to play in merging (or not) of this PR.

gdt:
Could you open an issue about the ssh/pgrp coupling and ^C?   That way it will survive this being moved.

(I am not at all sure that pty is better than pgrp, if pgrp is POSIX-specified and ptys are not.)

gdt:
I am inclined to merge this after #809 has settled, barring test reports that say there's an issue.

PR #814, opened 2022-10-30
Fix one more resource leak

tleedjarv:


gdt:
I'll merge this as soon as 2.53.0 is released.

PR #817, opened 2022-11-01
CI.yml: Drop old ocaml versions on all but ubuntu

gdt:
Drop ocaml 4.08 to 4.13 on macos and windows, on the theory that building on ubuntu is sufficient to show portability to older ocaml, and because windows builds in particular are too flaky.  With the new scheme, it is possible that there might be an undetected issue with older ocaml in OS-specific code, but basically I'd rather take that risk than have to intervene in most CI runs.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -47,32 +47,19 @@ jobs:
       matrix:
         job:
         - { os: macos-11      , ocaml-version: 4.14.0 }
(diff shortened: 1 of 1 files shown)


gdt:
I'm open to leaving in 4.08 for mac and windows, if people think that's necessary.   But, I'm also not sure there is any real value in building with old ocaml at all, now that there is no longer a wire protocol compat issue.

tleedjarv (approved):

PR #818, opened 2022-11-03
Make sed invocation more compatible

tleedjarv:
Some versions of `sed` require a space between a command and an argument. The space is actually a POSIX requirement, with no space being an optional extension.


gdt:
I'm inclined to merge this post-release, as it doesn't seem to be causing trouble -- I don't think it's serious enough to reroll and rc and restart the timer.  (Given how many things are accumuulating, I am expecting a .1 in a small number of months anyway.)

PR #820, opened 2022-11-07
opam file: Remove wrong build tag and unnecessary dependency

kit-ty-kate:
* ocamlfind is not used when using dune as far as i can see
* the use of the `{build}` tag makes unison completely disjoint with the rest of the opam state (e.g. if someone updates lablgtk they expect the packages that are using it to also use the new lablgtk). Using it this way is a hack rather than a feature.

### DIFF
--- a/unison.opam
+++ b/unison.opam
@@ -11,10 +11,9 @@ bug-reports: "https://github.com/bcpierce00/unison/issues"
 dev-repo: "git://github.com/bcpierce00/unison.git"
 build: ["dune" "build" "-p" name "-j" jobs]
 depends: [
-  "ocaml" {build & >= "4.08"}
-  "ocamlfind" {build}
-  "dune" {build & >= "2.3"}
-  "lablgtk3" {build & >= "3.1.0"}
+  "ocaml" {>= "4.08"}
+  "dune" {>= "2.3"}
+  "lablgtk3" {>= "3.1.0"}
 ]
 synopsis: "File-synchronization tool for Unix and Windows"
 description: """


unison.opam
```diff
@@ -11,10 +11,9 @@ bug-reports: "https://github.com/bcpierce00/unison/issues"
 dev-repo: "git://github.com/bcpierce00/unison.git"
 build: ["dune" "build" "-p" name "-j" jobs]
 depends: [
-  "ocaml" {build & >= "4.08"}
-  "ocamlfind" {build}
-  "dune" {build & >= "2.3"}
-  "lablgtk3" {build & >= "3.1.0"}
+  "ocaml" {>= "4.08"}
+  "dune" {>= "2.3"}
+  "lablgtk3" {>= "3.1.0"}
```
gdt, line 16: I think I understand this.  lablgtk3 is required to be present if unison gui is run, to provide the libs.

unison.opam
```diff
@@ -11,10 +11,9 @@ bug-reports: "https://github.com/bcpierce00/unison/issues"
 dev-repo: "git://github.com/bcpierce00/unison.git"
 build: ["dune" "build" "-p" name "-j" jobs]
 depends: [
-  "ocaml" {build & >= "4.08"}
-  "ocamlfind" {build}
-  "dune" {build & >= "2.3"}
-  "lablgtk3" {build & >= "3.1.0"}
+  "ocaml" {>= "4.08"}
```
gdt, line 14: Please bear with me as i am not clueful about opam but it's becoming clear somebody in unison needs to be.  As I read the docs, this is declaring that ocaml (which has compiler and runtime) is necessary at build time, but not at run time.   And the commit drops `build &` to indicate that it is necessary at run time.   In my build (with pkgsrc, not opam, NetBSD amd64), the unison binary does not depend on any compiler files.  But, it's a native build, not byte code, and it seems possible that a byte code build could be done, in which case I would expect a runtime lib would be needed.

Therefore, *in general*, ocaml is needed at run time.

Is that right?

unison.opam
```diff
@@ -11,10 +11,9 @@ bug-reports: "https://github.com/bcpierce00/unison/issues"
 dev-repo: "git://github.com/bcpierce00/unison.git"
 build: ["dune" "build" "-p" name "-j" jobs]
 depends: [
-  "ocaml" {build & >= "4.08"}
-  "ocamlfind" {build}
-  "dune" {build & >= "2.3"}
```
gdt, line 16: Can you explain why dune would be needed at runtime?  (or why I'm confused?)

gdt:
I have pinged the opam $MAINTAINER via email, asking for review comments.

unison.opam
kit-ty-kate: Because of https://github.com/ocaml/dune/issues/2147 dune is not backward compatible and would cause problem when downgrading.

gdt:
Also I am having trouble understanding opam docs which have
```
build dependencies are no longer needed at run-time: they won't trigger recompilations of your package.
```
and this is blurring:
  1) if the dep is not installed at runtime the package will fail
  2) some sort of interpackage dependency rule that is not clearly articulated

In general, if any build tool has changed, then it is necessary to rebuild the package because the output might be different.  This could happen because a bug was fixed in gmake, or bison, or really anything.   Since that's always true, this wouldn't seem to be talking about this, but it also seems to be that.   I do understand that this is a common pain point in packaging systems.

unison.opam
gdt: Why is dune involved after the package is built?   Is this about a dune metadata file that is interpreted if some other package depends on unison?  (Are there any such packages?)   Or because people expect to invoke some dune commands on installed opam packages in general, or unison in particular?

kit-ty-kate:
> In general, if any build tool has changed, then it is necessary to rebuild the package because the output might be different. This could happen because a bug was fixed in gmake, or bison, or really anything. Since that's always true, this wouldn't seem to be talking about this, but it also seems to be that. I do understand that this is a common pain point in packaging systems.

yes precisely. `{build}` is more of a hack to avoid unnecessary recompilation and should be avoided except for packages whose content shouldn't matter. e.g. `conf-*` and well as some build tools since if the package compilation succeeded it should be safe to keep the package in the same state (cue some annoying exceptions though..)

unison.opam
gdt: This is ok by the "lablgtk3 change requires rebuild" consistency property, so I won't object.

unison.opam
gdt: This one I still don't understand.

It seems like some people are telling me that opam is primary and others are telling me that dune-project should generate unison.opam, and my read is that there are differences of opinion within the ocaml community.  So I'm trying extra hard to understand.

kit-ty-kate: Dune installs a `dune-package` file that have the version of dune it has been compiled with:
```
$ cat ~/.opam/default/lib/unison/dune-package
(lang dune 3.5)
(name unison)
(version dev)
(sections (lib .) (bin ../../bin) (doc ../../doc/unison))
(files
 (lib (META dune-package opam))
 (bin (unison))
 (doc (LICENSE README.md)))
```

kit-ty-kate:
If you want to read up more on the issues with the build tag there is this excellent comment from David summarizing the situation: https://github.com/ocaml/opam/issues/3850#issuecomment-882632871

tleedjarv:
May I suggest this be discussed together with (and then included in) #671.

Also of interest may be the fact that the upstream project does not publish in the opam package repository. An independent maintainer is publishing the package. Thus the opam file in the upstream repository (if the generated file is committed at all) is not the same as it is in opam package repository.

kit-ty-kate:
> Also of interest may be the fact that the upstream project does not publish in the opam package repository. An independent maintainer is publishing the package. Thus the opam file in the upstream repository (if the generated file is committed at all) is not the same as it is in opam package repository.

Yes, that’s been myself (as an opam & opam-repository maintainer) for the past 4 releases. I try to use the upstream opam file whenever possible. I did simply use opam-publish (which takes the upstream opam file) all those 4 times and returned the fixes upstream (which is this here PR).

Be it in opam-repository or not, it shouldn’t matter to the discussion anyway given it would still affect users who choose to install it from here (e.g. via `opam pin` or `opam install .`)

tleedjarv:
> Be it in opam-repository or not, it shouldn’t matter to the discussion anyway given it would still affect users who choose to install it from here (e.g. via `opam pin` or `opam install .`)

Right. The discussion should include the question of who is maintaining the opam file. Currently in the `maintainer` field is the person who originally created the file (I think) but is apparently no longer maintaining it.

gdt:
I think there are separable issues and we can make progress on them separately

This is about minor fixes to the opam file, and I at this point do not have a problem with the proposed changes.  I do care about reproducible builds, even though unison isn't there, and I basically now believe that the build feature in dune is a bug.  Certainly it is not documented well enough.  Thus i'm ok with just removing it.

That leaves to be addressed:
  - maintainer field is incorrect (I have pinged and am waiting for response or timeout)
  - dune-project files vs opam

but it's ok to fix one thing and leave the others for another day.   So I'm going to merge this.

PR #832, opened 2022-11-15
opam/dune: Reset maintainer

gdt:
The previous maintainer is no longer active, so replace the email address with the hackers list as a way to denote that they are maintained within the project.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -8,7 +8,7 @@
 
 (license GPL-3.0-or-later)
 
-(maintainers "juergen@hoetzel.info")
+(maintainers "unison-hackers@lists.seas.upenn.edu")
 
 (authors "Trevor Jim" "Benjamin C. Pierce" "J\195\169r\195\180me Vouillon")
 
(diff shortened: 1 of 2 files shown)


gdt:
2 of 2 commenters on unison-hackers agree.

jhjourdan:
What is the policy of unison-hackers? Is anyone allowed to send to this mailing list?

gdt:
Subscribers only, usual anti-spam.   This is documented at https://github.com/bcpierce00/unison/wiki/Mailing-Lists

jhjourdan:
Well, then it is probably not great to use this in the maintainers field. This field is used by e.g., maintainers of opam-repository to ask for changes when there are issues with a package.

gdt:
I could make a wiki page and put that in the field instead.    In practice, though ,kit-ty-kate knows where to find us and has been filing PRs, and the github repo is listed as well.   I hope this situation is temporary as we are still in the process of working this all out (but making progress now, after a long time of not), so I lean towards not worrying about the rough edges too much, as its obviously a mailinglist and that's a big clue that there isn't really a maintainer.    (I'm not hostile to dune/opam; it's just not something I use so I haven't been putting time into it.)

PR #835, opened 2022-11-17
Detect Wayland or X11 displays before choosing text UI

mxmgh:
Otherwise, Unison will not start its graphical interface when running in a pure-Wayland environment.

### DIFF
--- a/src/uigtk3.ml
+++ b/src/uigtk3.ml
@@ -4477,11 +4477,14 @@ let start = function
       let displayAvailable =
         Util.osType = `Win32
           ||
-        try System.getenv "DISPLAY" <> "" with Not_found -> false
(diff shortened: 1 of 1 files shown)


gdt:
I am conceptually ok with this and will let you two resolve the ocaml style issues.

In the glorious future, we'll have unison and unison-gui, which I think won't do this, but in the meantime this seems to extend in the right way.

gdt:
It looks like this needs != converted to <> at least.   @mxmgh are you ok with the suggested change?  If you so you push the commit suggestion button, or redo the branch yourself.

tleedjarv:
I can't test this but the code looks fine. Squash the commits (no need to keep the Co-authored-by tag), or @gdt you can also do it when merging.

PR #836, opened 2022-11-17
Get the POSIX strerror_r(3) definition with glibc

tleedjarv:
glibc includes a non-POSIX version of strerror_r(3). POSIX-compliant version of strerror_r(3) is the default but some feature test macros must be defined in case CFLAGS change the default settings.

The same result could be achieved by just defining _DEFAULT_SOURCE but this is available only since glibc 2.19.

### DIFF
--- a/src/props_xattr.c
+++ b/src/props_xattr.c
@@ -160,6 +160,16 @@
  *
  */
 
+/* To get the correct definition of strerror_r(3) with glibc. */
+#if !defined(_POSIX_C_SOURCE) || defined(_POSIX_C_SOURCE) && _POSIX_C_SOURCE < 200112L
+#undef _POSIX_C_SOURCE
+#define _POSIX_C_SOURCE 200809L
+#endif
+#ifdef _GNU_SOURCE
+#undef _GNU_SOURCE
+#endif
+
+
 #include <caml/memory.h>
 #include <caml/alloc.h>
 #include <caml/fail.h>


gdt:
We need to tread carefully here.   I have seen a lot of cases where feature macros are defined, but the code uses things not defined by that feature.   Generally, systems hide everything not required by a defined macro.

I don't follow where CFLAGS that ask for something else are coming from.   That seems like an error compared to what BUILDING.md should say.    I wonder about detecting bad flags and #error, but I'll try to build and understand this better.

tleedjarv:
This was just a precaution. We may just leave it be and deal with it when any actual issues surface. The more I think about it, the more this seems like the right way to go.

> We need to tread carefully here. I have seen a lot of cases where feature macros are defined, but the code uses things not defined by that feature. Generally, systems hide everything not required by a defined macro.

My thinking is this. We don't target glibc, hence it is not only safe but also desirable to undefine _GNU_SOURCE. We do target POSIX, including XSI and anything non-POSIX is always guarded. Defining _POSIX_C_SOURCE should not only be safe but also desirable. Or it could be (sometimes must be) replaced by a definition of _XOPEN_SOURCE. Perhaps it is still useful to define _DEFAULT_SOURCE, too.

As to hiding everything not required, this is expected. This specific file is using non-standard interfaces and is very heavily guarded already. `strerror_r` is almost the only standard thing in there :D It uses _very_ few library functions overall.

> I don't follow where CFLAGS that ask for something else are coming from. That seems like an error compared to what BUILDING.md should say. I wonder about detecting bad flags and #error, but I'll try to build and understand this better.

Those CFLAGS come from packaging systems. In two ways. Many (most? all?) packaging systems provide their own CFLAGS when building projects. Unison's Makefile currently uses CFLAGS incorrectly but I have a patch fixing that (PR coming soon).

Second, the OCaml compiler itself will record CFLAGS used to compile it and will use those flags when itself compiling other files. You can see this with `ocamlc -config` or if you add `-verbose` to CAMLFLAGS in Makefile.OCaml.

tleedjarv:
I'll close this PR for now. It can be reopened whenever there's an actual issue.

PR #839, opened 2022-11-22
Fix another resource leak

tleedjarv:
Operator `;` has higher precedence than `try`. The code looked like it was doing the right thing but it would actually execute each following `close` only if the previous `close` failed. In the normal case all of these descriptors were leaked.

This code is currently used only by the native macOS GUI. I do not have the possibility to test this GUI.


gdt:
Given that we ~never have testing by other's I'm going to merge this as fixing an obvious bug, and if there is a latent bug exposed we'll deal with that later.

PR #844, opened 2022-12-06
Add logotype to README.md

llucps:
I was having fun re-creating the Unison logo from scratch and I though it would be a "silly" and nice to add the logo on the README.md file.

P.D. By the way I could add the Illustrator source file and a SVG if needed.


gdt:
Thanks for contributing.   I let this sit a bit to think about it.  Right now, we don't have a logo in the repo, and when I first saw this I had no idea what "Unison logo" you meant.

As I understand it, you have created a derived work of the logo from https://www.cis.upenn.edu/~bcpierce/unison/Unison.gif and have a "source file" which can be processed by a proprietary tool to create SVG, PNG, etc.   I think it's clear that if this were to be merged, we'd want the sources.   Assuming my guesses are correct, I am uncomfortable with both proprietary tool use as not meeting the project's license, and because most people will not be able to regenerate it.   I am curious why you didn't use inkscape -- but this is definitely not a request to reimplement.

To go full Debian on you :-) I'm also unclear on licensing of the original logo.  I will inquire about that and related things.

llucps:
Oh I assumed that was the "official" Unison logotype, since is the one used in the manual and documentation. 

I recreated it from scratch using Illustrator since is one of the tools I've been using for the last 25 years. But I've recreated a SVG version using Inkscape, so the source file is a SVG and no longer on a proprietary format.

I have no idea who created the original logotype and I assumed that it would be under the GPLv3 license since the projects is under this license?

P.D. As a side note, there is currently an Adobe Illustraor file called unison-icon.ai which is not mine in the same icons folder.

gdt:
Thanks for de-Illustratoring and pointing out what's in the repo; that helps a lot.    Until 2023, I have limited time for unison, basically only things that are particularly important or that I can deal with in a few minutes.   I will look at this again when I have time to dig into the history and how we should fully straighten things out.

llucps:
I totally understand, this is just a cosmetic "improvement". I don't really want to burden you with unnecessary work. I just had fun doing it and thought it would be nice to add the logo to the README file.

bcpierce00:
I'm happy to see a more modern representation of the logo!

I am the original author, and I'm totally happy to have it distributed
under whatever license terms.

    - B

gdt:
Great, then lets consider this logically approved and I will look into details in early 2023.  (I would just as soon clean up everything related as we do this, and I don't think it's going to be hard.)

Issue #827, opened 2022-11-09
Clarify and document the status of opam packages, and publish them in the opam repository


jhjourdan:
How should opam be used to install unison?

Looking at the sources, it seems like the opam file is written to provide two packages at once: `unison` and `unison-gui`. Could this be documented somewhere?

If the purpose of the `unison` package is to drop the dependency to lablgtk, then it should not mention it as a dependency.

Is there a reason to have two packages when opam features "optional dependencies", which are precisely designed to only enable a feature (e.g., a GUI) if the corresponding dependencies are enabled? Could this be documented?

The most recent unison package in the opam repository is 2.52.0. It does not contain any version of the `unison-gui` package. Would it be possible to have a more recent version (2.53), both with and without the GUI?

jhjourdan:
Yet another complain (sorry...): if the same opam file is used for two different packages, then it should be called `opam` and not `unison.opam`: otherwise, opam does not find it for the `unison-gui` package.

Also, it is weird that when the GUI is enabled, the executable produced by the `Makefile` is `unison`, while it is `unison-gui` when using `dune`.

gdt:
Basically unison.opam is not really maintained, and the people maintaining unison do not really understand opam.  unison's build system is simply make.   Arguably the opam file should be hosted in the opam world.

gdt:
See also #824

jhjourdan:
I don't understand, then: why is there a parallel dune build system for opam packages? Does any of the maintainer use dune?

I could volunteer becoming the opam package maintainer, but then I will choose the solution that requires the least maintenance effort: I would drop all the dune build system, and have a simple opam file that call make. Sure, this is perhpas a bit sad for e.g., parallelization, but at least we will have a working opam package...

Cc @olafhering, @kit-ty-kate, @ejgallego, @liyishuai, who seem to be involved in opam/dune in unison.

olafhering:
There is indeed a life outside of opam.
If I were in charge, all and everything would be built with dune, simply because it gets everything right (AFAICT).

Regarding unison itself, there is enough of dune support in unison.git to be able to build binaries with `dune` instead of `make`. However, some pieces are still missing, like generating the documentation. This is just a matter of man power to get this sorted out.

[...]

PR #845, opened 2022-12-14
Use make as a build system for opam files.

### LINKED ISSUE #842: Change opam packaging to have separate CLI/GUI packages and maybe fsmonitor
Currently, there is only one opam file.  I think it requires gtk3 on the host and builds a CLI, but the point is that one should be able to build CLI or GUI as one wishes.

Somewhere between perhaps and probably, have an opam file for fsmonitor (on the set of platforms for which there is an fsmonitor implementation, with sensible behavior on other ones).

While doing this, align with the `unison` and `unison-gui` plan in #667.

jhjourdan:
There are now two opam files:
- unison.opam is the main opam file. It has an optional dependency to lablgtk3, so that the GUI is built only if the lablgtk3 package is installed.
- unison-gui.opam is a pseudo-package, with both unison and lablgtk3 as dependencies. It can be used to force the GUI to be installed.
    
As a side effect, remove the package information from the dune-project files, to avoid redundancy.

Fixes #842, #827


gdt:
Thanks for submitting,  I will not be able to dig into this until January sometime.

jhjourdan:
Sure. Sorry for the delay!

jhjourdan:
One potential subject of discussion is the status of `unison-fsmonitor`. Up to now, this executable is not installed by `make install`, but seems to be needed by some features of unison. So I considered that a bug of the Makefile, and decided to install it as part of the `install` Makefile target.

Is this the right thing to do?

gdt:
unison-fsmonitor is built on some places and not others, not just !mac.  Also not on netbsd.

I am not sure we want to go down this path.  I do not have time to pay attention until January.

jhjourdan:
For the record, I rebased this branch, removing my change about the installation of fsmonitor.

And then I noticed that the install target has been removed, making this change more subtle because then the opam file needs to contain the installation script.

I'm not going to continue until anyone tells he is interested.

tleedjarv:
The install taget is in #939. Your testing and/or review would be very helpful.

jhjourdan:
Thanks!

I rebased this PR on top of #939, and this seem to work well.

gdt:
Can you rebase to master?  I came to this intending to read what I expected to be a small number of commits.  I'm guessing a rebase is trivial and leads to that.  I did kick off the workflow.

gdt:
Thanks.  Now I can be 100% sure easily this only touches opam.  I have kicked off  a workflow run and will hit merge if it finishes green.

jhjourdan:
Perfect, thanks!

I'll do my best to publish releases on OPAM as soon as they appear here.

PR #850, opened 2023-02-05
GHA: Update OCaml versions

tleedjarv:
Add some new OCaml versions to CI. While at it, move some Mac build configuration out of source/makefile into the CI script to make it easier to adjust according to the environment.


src/Makefile.OCaml
```diff
@@ -216,17 +216,11 @@ else
   buildexecutable:: $(NAME)$(EXEC_EXT)
 endif
 
-MINOSXVERSION=10.6
-ifeq ($(OSARCH),osx)
-  CAMLCFLAGS+=-ccopt -mmacosx-version-min=$(MINOSXVERSION)
-  XCODEFLAGS=-arch $(shell uname -m)
-endif
-
```
gdt, line 224: What happens if someone builds without adding in what is in CI?   Do they get a working native build, with defaults for target and arch, that will work on the system it was built on?  It's not immediately clear that this is CI glue vs necessary gyrations for building on mac.

tleedjarv: The deployment target (macosx-version-min) is not relevant for regular users building for themselves. It _might_ be relevant for packagers or users with special needs, in which case they know what they're doing and would need to set their environment properly. In my view, this should definitely be excluded from static build conf.

Arch is a bit trickier. Build conf should not include a static arch (even if "static" here means `uname -m`); it should be the user requesting whatever they want (and fail if it's not supported). As I understand it, OCaml's cross-compilation story is quite weak, but still, why have the arch in the makefile.

This is where it gets tricky. I think that macOS (since ver 11?) by default wants to build universal binaries (both amd64 and arm64), which means that there will _always_ be cross compilation involved. This will break builds that don't request an arch explicitly. I don't know why a user would build universal binaries to run on their own machine (as opposed to distributing them) but I think this is the Xcode default... Setting `ONLY_ACTIVE_ARCH = YES` should solve this, but it does not seem to work (at least not in GHA environment) and it would also count as having a static arch in the build conf. Oh well...

That said, I don't have a Mac myself, so this is all just theory. Unless someone can confirm one way or another then I will roll back the arch changes but keep the deployment target changes.

gdt:
I will try to run the build on a mac in the next few days.

PR #852, opened 2023-02-09
Added detail on defining a path on MacOS

bunk3m:
@gdt suggested that I add an edit (Aug 2022) to the manual that shares the answers that @tleedjarv graciously provided for me. I was having problems defining a root on OSX. 
OSX generally requires a backslash escape when there is a space in the path.  However, Unison doesn't need the backslash escape when defining paths on OSX.  I hope the addition helps describe the OSX special case.

### DIFF
--- a/doc/unison-manual.tex
+++ b/doc/unison-manual.tex
@@ -1498,9 +1498,15 @@ for a boolean flag or
 \end{verbatim}
 for a preference of any other type.
 
-Whitespaces around {\tt p} and {\tt xxx} are ignored.
-A profile may also include blank lines and lines beginning
+A profile may include blank lines and lines beginning
 with {\tt \#}; both are ignored.
+
+Spaces and tabs before and after {\tt p} and {\tt xxx} are ignored.
+Spaces, tabs, and non-printable characters within values are not
+treated specially, so that e.g. \verb|root = /foo bar| refers to a
+directory containing a space.
+(On systems using newline for line ending, carriage returns are
+currently ignored, but this is not part of the specification.)
 
 When Unison starts, it first reads the profile and then the command
 line, so command-line options will override settings from the
--- a/src/strings.ml
+++ b/src/strings.ml
@@ -2350,8 +2350,14 @@ let docs =
       \n\
       \032  for a preference of any other type.\n\
       \n\
-      \032  Whitespaces around p and xxx are ignored. A profile may also include\n\
-      \032  blank lines and lines beginning with #; both are ignored.\n\
+      \032  A profile may include blank lines and lines beginning with #; both are\n\
+      \032  ignored.\n\
+      \n\
+      \032  Spaces and tabs before and after p and xxx are ignored. Spaces, tabs,\n\
+      \032  and non-printable characters within values are not treated specially,\n\
+      \032  so that e.g. root = /foo bar refers to a directory containing a space.\n\
+      \032  (On systems using newline for line ending, carriage returns are\n\
+      \032  currently ignored, but this is not part of the specification.)\n\
       \n\
       \032  When Unison starts, it first reads the profile and then the command\n\
       \032  line, so command-line options will override settings from the profile.\n\


tleedjarv:
I don't think this is specific to OSX. You can leave out any mention of an OS and just state that spaces in paths don't need to be escaped.

(Haven't tested this, but I think this also applies to other special characters; in fact, there is no way to escape anything.)

tleedjarv:
> Is the "no need to escape" something that would be explained in the oocaml compiler docs so we can confirm our belief?

Actually the opposite. The "no need to escape" is the default and the "need to escape" is something exceptional that is required by the shell, for example. I quickly looked at the code (it's spread in too many separate locations to post here for reference). It does no other processing than trimming whitespace from beginning and end when reading from a profile. So there _are_ paths that can't be specified in the profile file (but they should still be possible when specified on the command line directly!).

So:
no escaping needed in the profile file;
escaping needed on the command line, but this is required _and_ processed by the shell - none of this escaping reaches Unison.

gdt:
I squashed the three commits together, because I only want to merge clean history.

I did some experiments (on NetBSD) and found that while spaces (and tabs!) in the middle of the line are read, initial spaces and tabs and trailing spaces and tabs are dropped.  A BEL (^G, 007) also worked.   And, CR is dropped everywhere.   I edited to reflect this more complicated situation (which I didn't understand before as I am someone who does not believe in putting spaces in directory names!)  Some of this behavior seems perhaps buggy, and some happenstance, so the manual leaves most behavior outside the spec to avoid precluding cleanup.

I then force-pushed to your branch.

I did not dig into UTF-8.

@bunk3m This commit lists you as author, which I think is good because you did most of the digging and bringing this to our attention.   Let me know if you think this is accurate and if you're ok with this result.   Anyone else, the usual review is appreciated.

tleedjarv:
I'm fine in general but the location in the manual is a bit difficult. The section is about paths and then all the examples are about the `root` preference. Of course, processing of paths and roots is the same, so nothing's factually wrong, just seems somewhat out of place. A separate subsection is probably the best, making it very clear that this applies to both roots and paths.

(No, I don't think it's bikeshedding because the "path" section in the manual is about a specific concept, not about fs paths in general. Also, since the manual does eventually need to be updated and overhauled, I would not want these interim additions make that future work more difficult.)

gdt:
I see; paths are a concept distinct from roots even though they both relate to unix pathnames.  And agreed that this is about how the profile file reader reads unix pathnames for both roots and paths.  I'll have a look at moving it.

bunk3m:
> @bunk3m This commit lists you as author, which I think is good because you did most of the digging and bringing this to our attention. Let me know if you think this is accurate and if you're ok with this result. Anyone else, the usual review is appreciated.

Thank you and yes, it's OK.  

This was great as this is my first contribution on github.  I appreciate everyone's help in sorting it out.  I'm happy to help again.  Thank you!

gdt:
> I'm fine in general but the location in the manual is a bit difficult. The section is about paths and then all the examples are about the `root` preference. 

I moved this to the section that introduces profile files.

doc/unison-manual.tex
```diff
@@ -1535,6 +1535,23 @@
 case, all actions that have been selected for a set of changes
 currently being displayed will be discarded.
 
+When unison reads roots from a profile, spaces and tabs immediately
```
tleedjarv, line 1538: I'm fine overall, just one last nitpick. This sentence is about roots specifically, but it's actually true for all pathnames. Also, there is a more generic sentence about the same thing ~40 lines earlier (line 1501).

gdt: Good spot on earlier text.   I merged the space caution into that and simplified a lot.  Needs a strings regen, after an issue on master is resolved (either needs a strings regen or we have some odd environment regen issue again).

gdt: strings issues are resolved.   I'll merge this tomorrow or sooner if I hear you are happy with it.

Issue #865, opened 2023-03-09
rsync >= 3.2.4 quotes arguments itself, making unison fail when it quotes arguments


rrthomas:
See the discussion of `--old-args` in `rsync(1)` for how to handle this; also:

> If you use scripts that have been written to manually apply extra quoting to the remote rsync args (or to require remote arg splitting), you can ask rsync to let your script handle the extra escaping.  This is done by either adding the --old-args option to the rsync runs in the script (which requires a new rsync) or exporting RSYNC_OLD_ARGS=1 and RSYNC_PROTECT_ARGS=0 (which works with old or new rsync versions).

Or `copyquoterem`'s default could be made to depend on the rsync version.

rrthomas:
(On reflection this issue must surely have been reported already, as rsync 3.2.4 was released nearly a year ago; but I have searched closed issues and can't find a similar one!)

gdt:
More importantly, bugs in rsync are not bugs in unison!!
But seriously: looks like you opened an issue in wrong repo.

gdt:
If it's a unison issue, please make it clear what is wrong in unison, and which version of unison you find the bug in, and a reproduction recipe.  My system has rsync 3.2.7 and I have seen no problems.  I don't claim there are none -- but it's necessary to explain what the problem is.

rrthomas:
Sure thing, sorry:

```
$ unison -version
unison version 2.52.1 (ocaml 4.08.1)
$ rsync --version
rsync  version 3.2.7  protocol version 31
…
$ cat .unison/test.prf
root = /home/rrt/tmp
root = ssh://localhost//home/rrt/tmp2
copythreshold = 0
$ pwd
/home/rrt
$ ls -l tmp
total 4.7M
-rw-r--r-- 1 rrt rrt 4.7M Mar  5 22:33 foo.opus
$ ls -l tmp2
total 0
$ unison -confirmbigdel -batch -ui text test
Unison 2.52.1 (ocaml 4.08.1): Contacting server...
Connected [//dwks//home/rrt/tmp -> //dwks//home/rrt/tmp2]
 
Looking for changes
  Waiting for changes from server
Reconciling changes
deleted  ---->            .bak.0.foo  
local        : deleted
dwks         : unchanged file     modified on 2023-03-09 at 23:49:11  size 0         rw-r-----
new file ---->            foo.opus  
local        : new file           modified on 2023-03-05 at 22:33:24  size 4891675   rw-r--r--
dwks         : absent
[...]

PR #866, opened 2023-03-10
Avoid problems with quoting arguments to copyprog

rrthomas:
Use Unix.open_process_args_{full,in} instead of Unix.open_process_{full,in},
passing command-line arguments as an array, so they are not parsed by the
shell and do not need to be quoted.

Add the function runExternalProgramArgs to src/external.ml to implement
this.

This is motivated by recent changes to rsync:
https://download.samba.org/pub/rsync/NEWS#3.2.4

See also https://github.com/bcpierce00/unison/issues/865

This obviates the need for the copyquoterem preference, which is ignored for
backwards compatibility, but its documentation is removed.


gdt:
Have you tested with old rsync also?

tleedjarv:
A few quick comments before further review:
 * I'm not sure about removing `copyquoteterm` entirely. Copyprog could be something other than `rsync` and even though the need to quote may be unlikely, the quoting code is already there so why not keep it? (The default=true for rsync should be removed, though.)
 * Instead of doing a full copy for `runExternalProgramArgs`, it's better to make the existing `runExternalProgram` into a universal auxiliary function and then add new `runExternalProgram` and `runExternalProgramArgs` to call the auxiliary func with proper params for args/non-args. (in the end, the plain non-args version could be removed completely?)

tleedjarv:
This is from OCaml's manual re `Unix.open_process_args_{full,in}`:
> The executable file prog is searched in the path. This behaviour changed in 4.12; previously prog was looked up only in the current directory.

If I read this correctly then this is serious and basically makes this PR not work when built with OCaml < 4.12.

rrthomas:
Thanks very much for the comments so far; some replies:

>     * I'm not sure about removing `copyquoteterm` entirely. Copyprog could be something other than `rsync` and even though the need to quote may be unlikely, the quoting code is already there so why not keep it? (The default=true for rsync should be removed, though.)

I think that to retain `copyquoterem` there should be a common use case, at least. It's a hack that works around an implementation problem (which has now been fixed), and it was in rsync, the default `copyprog`. I don't think it should be Unison's problem to work around similar flaws in other programs (which, as in rsync, are probably in any case security bugs). Fixing Unison not to use the shell 

>     * Instead of doing a full copy for `runExternalProgramArgs`, it's better to make the existing `runExternalProgram` into a universal auxiliary function and then add new `runExternalProgram` and `runExternalProgramArgs` to call the auxiliary func with proper params for args/non-args. (in the end, the plain non-args version could be removed completely?)

Absolutely. I have not done so in the first pass purely because I'm quite new to OCaml, and because I wanted to ensure that the initial implementation was correct.

> Generally we have a commit with real changes and then a "regen strings" commit separately. If that's reasonably easy for you to split out please do.

I'll do that!

> If I read this correctly then this is serious and basically makes this PR not work when built with OCaml < 4.12.

Good catch, I guess that means this change would have to wait to be merged until the minimum requirement is lifted from the current 4.08 to 4.12?

src/copy.ml
```diff
@@ -914,14 +914,8 @@ let copythreshold =
...
     ~category:(`Advanced `General)
-    "add quotes to remote file name for copyprog (true/false/default)"
-    ("When set to {\\tt true}, this flag causes Unison to add an extra layer "
-     ^ "of quotes to the remote path passed to the external copy program. "
-     ^ "This is needed by rsync, for example, which internally uses an ssh "
-     ^ "connection requiring an extra level of quoting for paths containing "
-     ^ "spaces. When this flag is set to {\\tt default}, extra quotes are "
-     ^ "added if the value of {\\tt copyprog} contains the string "
-     ^ "{\\tt rsync}.")
+    "ignored for backward compatibility"
```
tleedjarv, line 917: Instead of this, use  `Prefs.markRemoved`. Example: 7a28f01c0

(Alternative is to first use `~deprecated:true` for a couple of releases but in this case it may be ok to remove directly.)

gdt:
I think it's going to be a very long time before we required >= 4.12.   The change to 4.08 was pretty recent and even that had some gnashing of teeth.

So I guess that means for now probably changing the default, and more importantly enhancing the documentation to make it easier to understand.  Right now I think someone coming to this is not going to follow what's going on.

rrthomas:
> I think it's going to be a very long time before we required >= 4.12. The change to 4.08 was pretty recent and even that had some gnashing of teeth.

Sure, I can put a note in my diary for however long that is. OCaml 4.08 was released 3½ years before that change; a conservative 4 years after 4.12 is around 2 years from now.

> So I guess that means for now probably changing the default, and more importantly enhancing the documentation to make it easier to understand. Right now I think someone coming to this is not going to follow what's going on.

If the change is made at a time when most users will have a suitable version of rsync, then the only people who will need an explanation are those stuck with old rsync. A big "heads up, if you're using unpatched rsync < 3.2.4, you will need `copyquoterem = true`"?

src/copy.ml
```diff
@@ -1037,28 +1031,18 @@ let transferFileUsingExternalCopyprog
...
       Prefs.read copyprog
   in
-  let extraquotes = Prefs.read copyquoterem = `True
-                 || (  Prefs.read copyquoterem = `Default
-                    && Util.findsubstring "rsync" prog <> None) in
-  let addquotes root s =
-    match root with
-    | Common.Local, _ -> s
-    | Common.Remote _, _ -> if extraquotes then Uutil.quotes s else s in
+  let progTokens = Str.split (Str.regexp "[ \t]+") prog in
```
tleedjarv, line 1034: Just a nit: Since this code is executed for every transfer then it may be beneficial to compute `Str.regexp "[ \t]+"` only once (make it a top-level value). Probably not worth it...

Since you don't use progTokens until below in the code, just move this line there.

gdt:
It's not really strictly time; it's about arriving at a place where the benefits of dropping the old versions outweigh the pain, and that ends up being about Laggard Term Stable distributions where people expect to build new unison, which we gradually decide not to support.  But I have no idea where we'll be in two years.

I didn't mean an explanation of the change.   I meant that the existing explanation of quoting rules was not adequately clear.   I've been using sh for a really long time and I didn't follow it the first time.   Really this is exposing an implementation detail that shouldn't be exposed in the first place.  I'll continue in the ticket, since I think this PR is not going to get merged any time soon.

src/copy.ml
```diff
@@ -1037,28 +1031,18 @@ let transferFileUsingExternalCopyprog
...
-             ^ (Uutil.quotes fromSpec) ^ " "
-             ^ (Uutil.quotes toSpec) in
-  Trace.log (Printf.sprintf "%s\n" cmd);
+    ^ (Fspath.toString (Fspath.concat fspathTo pathTo)) in
+  Trace.log (prog ^ " " ^ fromSpec ^ " " ^ toSpec ^ "\n");
   Lwt_util.resize_region !copyprogReg (Prefs.read copymax);
   Lwt_util.run_in_region !copyprogReg 1
-    (fun () -> External.runExternalProgram cmd) >>= fun (_, log) ->
+    (fun () -> External.runExternalProgramArgs (List.hd progTokens)
+                 (Array.of_list (List.append (List.tl progTokens) [fromSpec; toSpec]))) >>= fun (_, log) ->
```
tleedjarv, line 1045: Here I would suggest a style change.
Instead of using `List.hd`, `List.tl` and `List.append`, you could write
```ocaml
  let prog :: args = Str.split (Str.regexp "[ \t]+") prog in (* or whatever names you choose instead of prog and args *)
  Lwt_util.run_in_region !copyprogReg 1
    (fun () -> External.runExternalProgramArgs prog
                 (Array.of_list (args @ [fromSpec; toSpec]))) >>= fun (_, log) ->

File: src/external.ml [hunk 4]
```diff
@@ -114,3 +120,38 @@ let runExternalProgram cmd =
          else "\n\n" ^ Util.process_status_to_string returnValue)))
       (* Stop typechechecker from complaining about non-exhaustive pattern above *)
       | _ -> assert false)
+
+let runExternalProgramArgs cmd args =
```
tleedjarv, line 124: It's probably OK to just remove `runExternalProgram` completely and keep just `runExternalProgramArgs`.

The current `runExternalProgram` is used in very few locations and it may be worthwhile just changing those invocations to use the args-variant.

tleedjarv (changes requested):
I, too, am ok with the concept.

The path problem with OCaml < 4.12 could be solved by doing own path search. This way the code could be merged even before the minimum becomes 4.12. Not sure if it's worth the effort, though.

rrthomas:
Thanks very much for the review, which I'll now work through.

> The path problem with OCaml < 4.12 could be solved by doing own path search. This way the code could be merged even before the minimum becomes 4.12. Not sure if it's worth the effort, though.

I agree: it's not worth the effort. This is a nice improvement to have when it's available, but it doesn't seem to be a huge problem in practice.

src/copy.ml
rrthomas: I've used `Prefs.markremoved`.

src/copy.ml
rrthomas: Thanks! `@` was something I was unable to find with quite a lot of searching.

With this change, I now get a warning of a non-exhaustive match, as the list could be empty. I looked into this possibility before, and it seems that the list can't in fact be empty, because if `copyprog` is set to an empty value, then the default is used. I assume there is some idiom for this case?

src/external.ml
rrthomas: I looked into this before, and I think it's tricky, because currently `runExternalProgram` is used for example for `merge`. Now, I have the following setting:

```
merge = emacsclient -q --eval '(ediff-merge-files-with-ancestor "CURRENT1" "CURRENT2" "OLD" nil "NEW")'
```

This won't work if it's parsed as a series of space-separated tokens; the parser needs to understand quoting. For `copyprog` I don't think this is a problem, as its syntax is fixed: the arguments will always be added to the end (and as we've said elsewhere, using anything other than rsync is dubious). For `merge` and `merge2`, they are substituted into the middle.

While it would be nice to avoid calling the shell, I don't think swapping the user's shell for a custom quote parser is a good idea. For this sort of case, it would be better IMO to change `merge` so that it too puts its arguments in a fixed place, and if you want something like my complex setting above you have to write a short shell-script or similar.

src/copy.ml
tleedjarv: > With this change, I now get a warning of a non-exhaustive match, as the list could be empty.

Right, of course. And this is why the code should be written in this style in the first place; to get the compiler to point out the mistakes.
Using, `List.hd` and `List.tl` is worse because both of these can raise runtime exceptions.
The better way would then be something like:
```ocaml
  let prog, args (* or whatever names you choose instead of prog and args *) = match Str.split (Str.regexp "[ \t]+") prog with [] -> assert false (* still fails at runtime *) | h :: t -> h, t in
  Lwt_util.run_in_region !copyprogReg 1
    (fun () -> External.runExternalProgramArgs prog
                 (Array.of_list (args @ [fromSpec; toSpec]))) >>= fun (_, log) ->
```

As you can see, this can still fail at runtime. If this is unacceptable then validation must be performed to make sure that the original value of `prog` is never empty. Maybe this validation already exists, I haven't checked.

tleedjarv: What is the correct way of specifying args? Right now you exclude the executable itself from the list of args, but isn't that supposed to be the first argument? (that is, instead of extracting the tail of list, just use the entire list)

src/external.ml
tleedjarv: You're right, we can't sensibly parse values of such preferences.

I just noticed that something like this may happen with the remote shell args as well:
https://github.com/bcpierce00/unison/blob/a66be8529f75abfeb66b8a54e72a7a1df3d527d1/src/remote.ml#L1817-L1832
(here, `create_process` and `exec*` functions are used)

The value of `shellCmdArgs` comes from a user preference and is later simply split by spaces. Nothing to do with this PR. I just noticed that the same issue lurks within that code, although mostly theoretical. As you've shown, for the merge preference this issue is very much real.

Let's keep these functions separate, as they are.

src/copy.ml
rrthomas: > As you can see, this can still fail at runtime. If this is unacceptable then validation must be performed to make sure that the original value of `prog` is never empty. Maybe this validation already exists, I haven't checked.

As I said above, `prog` can't be empty, because if it is, the default value is used. So I think it would be sufficient to add a comment to this effect.

rrthomas: > What is the correct way of specifying args? Right now you exclude the executable itself from the list of args, but isn't that supposed to be the first argument? (that is, instead of extracting the tail of list, just use the entire list)

That didn't occur to me; in any case, it doesn't appear to be the case, because the code works as-is.

The documentation is ambiguous; `open_process_args` refers the reader to `open_process_args_out`, which refers the reader to `open_process_args_in` (a rather eccentric ordering IMO: why isn't `open_process_args` first in the documentation, with the other two functions referring directly to it? and there are other problems, as `open_process_args_out` is not the same as `open_process_args_in` except for redirecting stdin to a pipe, as it doesn't redirect stdout to a pipe…). Anyway (I shall file an upstream bug report!). `open_process_args_in` says it "runs the program `prog` with arguments `args`. This is the bit that's ambiguous. As I say, from the fact that the code I wrote works, it clearly doesn't allow `argv[0]` to be set differently from `prog`, but this is not clear.

tleedjarv: > > As you can see, this can still fail at runtime. If this is unacceptable then validation must be performed to make sure that the original value of `prog` is never empty. Maybe this validation already exists, I haven't checked.
> 
> As I said above, `prog` can't be empty, because if it is, the default value is used. So I think it would be sufficient to add a comment to this effect.

It still can be empty. All it takes is for someone to change the default to empty (unlikely, but still) or just passing an whitespace-only value from command line. It's a minor issue and the added validation would still have to terminate the program which will also happen with `assert false`, except in the latter case you can't specify the error message. This is not a problem for me.

> > What is the correct way of specifying args? Right now you exclude the executable itself from the list of args, but isn't that supposed to be the first argument? (that is, instead of extracting the tail of list, just use the entire list)
> 
> That didn't occur to me; in any case, it doesn't appear to be the case, because the code works as-is.

All these functions work by calling `posix_spawn` or `execve`. The man pages have this to say:
for execve(2):
```
By convention, the first of these strings (i.e., argv[0]) should contain the filename associated with the file being executed.
```
for posix_spawn(3):
```
The arguments to the created process come from the argv argument, which is processed as for execve(2).
```

I don't know what is correct...

rrthomas: Sorry, I crossed paths with you above and came to the same conclusion; I will now revisit my PR.

rrthomas: Oh, of course my code seemed to work because the default value for `copyprog` contains multiple options, so the first of these became `argv[0]`.

rrthomas: I've updated the PR to check the list is non-empty (with an `assert false`), and to use all the `args` as arguments.

rrthomas: I have filed https://github.com/ocaml/ocaml/pull/12105 to try to improve the documentation of `open_process_args*`.

src/copy.ml
```diff
@@ -1031,43 +1022,36 @@ let transferFileUsingExternalCopyprog
...
-             (Path.toString pathFrom)
-             l (if l="" then "" else "\n"));
-  Uutil.showProgress id (Props.length desc) "ext";
-  finishExternalTransferOnRoot rootTo rootFrom
-    ((snd rootFrom, pathFrom, fspathTo, pathTo, realPathTo),
-     (update, desc, fp, ress, id))
+  let args = Str.split (Str.regexp "[ \t]+") progWithArgs in
+  match args with
+  | [] -> assert false
+  | prog :: _ ->
```
tleedjarv, line 1042: I'm fine with what you have done here but I usually like to keep the diff small, unless the changes are a clear improvement. You don't have to change this but I would propose to use a one-liner that will keep the indentation of all following lines unchanged.
```suggestion
  let prog = match args with [] -> assert false | h :: _ -> h in
```
The same works with multi-line as well, of course.

src/external.ml
```diff
@@ -114,3 +120,38 @@ let runExternalProgram cmd =
...
+  if Util.osType = `Win32 && not Util.isCygwin then begin
+    debug (fun()-> Util.msg "Executing external program windows-style\n");
+    let c = openProcessArgsIn cmd args in
```
tleedjarv, line 127: Since the only difference in these two functions is the `openProcessArgs*` call in two locations then I propose an implementation in spirit of this:
```diff
diff --git a/src/external.ml b/src/external.ml
index cefb9ea9..bbe03d48 100644
--- a/src/external.ml
+++ b/src/external.ml
@@ -80,10 +80,10 @@ let readChannelsTillEof l =
        >>= (fun res -> return (String.concat "\n" (Safelist.rev res))))
     l
 
-let runExternalProgram cmd =
+let runExternalProgramAux ~winProc ~posixProc =
   if Util.osType = `Win32 && not Util.isCygwin then begin
     debug (fun()-> Util.msg "Executing external program windows-style\n");
-    let c = openProcessIn ("\"" ^ cmd ^ "\"") in
+    let c = winProc () in
     let log = Util.trimWhitespace (readChannelTillEof c) in
     let returnValue = closeProcessIn c in
     let resultLog =
@@ -95,7 +95,7 @@ let runExternalProgram cmd =
          "") in
     Lwt.return (returnValue, resultLog)
   end else
-    let (out, ipt, err) as desc = openProcessFull cmd in
+    let (out, ipt, err) as desc = posixProc () in
     let out = Lwt_unix.intern_in_channel out in
     let err = Lwt_unix.intern_in_channel err in
     readChannelsTillEof [out;err]
@@ -114,3 +114,13 @@ let runExternalProgram cmd =
          else "\n\n" ^ Util.process_status_to_string returnValue)))
       (* Stop typechechecker from complaining about non-exhaustive pattern above *)
       | _ -> assert false)
+
+let runExternalProgram cmd =
+  runExternalProgramAux
+    ~winProc:(fun () -> openProcessIn ("\"" ^ cmd ^ "\""))
+    ~posixProc:(fun () -> openProcessFull cmd)
+
+let runExternalProgramArgs cmd args =
+  runExternalProgramAux
+    ~winProc:(fun () -> openProcessArgsIn cmd args)
+    ~posixProc:(fun () -> openProcessArgsFull cmd args)
```

tleedjarv (commented):
I'm fine with the patch and I would happily see it merged (no point in letting it rot for years).

I do have two remaining code comments but both are about style. Yet, the second one (avoid copying the entire function in external.ml) is kind of important.

Do also remember to regen strings.ml.

I doubt that discussions on whether and how to have external rsync support will reach any decision quickly, if at all. I do support merging this PR as an immediate fix but there are two remaining conditions:
 - for OCaml 4.12 the path search code (like in fswatch.ml) must be added (it should be clearly commented as being for compat only and must be removed once 4.12 becomes the minimum) (the path search should be done only once, not for every synced file)
 - the patch must be tested in Windows (even though I doubt there are many rsync users on Windows, but we can't allow a regression here)

src/copy.ml
rrthomas: Thanks, I've made this change; I agree of course about small diffs.

rrthomas: I have also updated `strings.ml`.

src/external.ml
rrthomas: I've applied this patch; much nicer, thanks!

rrthomas:
>     * for OCaml 4.12 the path search code (like in fswatch.ml) must be added (it should be clearly commented as being for compat only and must be removed once 4.12 becomes the minimum)

I've added the code from `fswatch.ml` as highlighted above, also checking the case that the command name contains `/`, in which case we don't check `PATH` (exactly as for `execvpe`).

> (the path search should be done only once, not for every synced file)

Hmm, I'm not sure how to do this safely. I have added calls to `search_in_path` in `openProcessArgs{In,Full}`, which will indeed be called once per file. But since this will happen anyway with `open_process_args_in`/`full` in OCaml >= 4.12, it seems OK to do it once per file, and in the mean time any other callers of `openProcessArgs*` will work.

>     * the patch must be tested in Windows (even though I doubt there are many rsync users on Windows, but we can't allow a regression here)

I looked at doing this, but I see Cygwin is needed for building, so gave up (I do build and test other packages regularly on Mingw-64 using MSYS2).

tleedjarv:
> > (the path search should be done only once, not for every synced file)
> 
> Hmm, I'm not sure how to do this safely. I have added calls to `search_in_path` in `openProcessArgs{In,Full}`, which will indeed be called once per file. But since this will happen anyway with `open_process_args_in`/`full` in OCaml >= 4.12, it seems OK to do it once per file, and in the mean time any other callers of `openProcessArgs*` will work.

I'm not sure about the performance penalty. You're right that this will happen anyway (even with current `openProcess` which goes through the shell) so maybe it's not worth thinking about this.

> > ```
> > * the patch must be tested in Windows (even though I doubt there are many rsync users on Windows, but we can't allow a regression here)
> > ```
> 
> I looked at doing this, but I see Cygwin is needed for building, so gave up (I do build and test other packages regularly on Mingw-64 using MSYS2).

Ah, no worries. Once the CI runs then you can just download the built binaries; no need to build yourself.

src/copy.ml
```diff
@@ -1031,39 +1022,30 @@ let transferFileUsingExternalCopyprog
...
+                 (Array.of_list (args @ [fromSpec; toSpec]))) >>= fun (_, log) ->
   debug (fun() ->
-           let l = Util.trimWhitespace log in
-           Util.msg "transferFileUsingExternalCopyprog %s: returned...\n%s%s"
-             (Path.toString pathFrom)
-             l (if l="" then "" else "\n"));
+      let l = Util.trimWhitespace log in
+      Util.msg "transferFileUsingExternalCopyprog %s: returned...\n%s%s"
+        (Path.toString pathFrom)
+        l (if l="" then "" else "\n"));
```
tleedjarv, line 1048: One more nit: there is still indentation change here.

gdt:
I'll look at this after the release happens.    We were already close and this is too scary to drop in at the last minute, but I won't feel that way early after a  point release.

tleedjarv:
> I did this, but I couldn't work out how to test the patch, because I couldn't find out how to use `rsync` with Unison on Windows. I tried using rsync (from msys, for Mingw), but it assumes that files containing a colon are remote, and even if I specify the local path as `\foo\bar`, unison translates that to `C:/foo/bar`.
> 
> Searching online did not enlighten me; rather, most tutorials for using rsync on Windows use WSL.

Do we conclude from this that this functionality never really worked with Unison on Windows? (or rather that rsync itself doesn't work outside some POSIX-y container like WSL or Cygwin)

Then we have no regression either. That's good enough for me.

tleedjarv (approved):
I have not tested myself but the code looks good.

gdt:
I'm ok with merging this, given my understanding:
  - the code will work on both old and new rsync on unix-y platforms
  - with the old code, external rsync doesn't seem to work on Windows, so it's not a regression
  - copyprog usage is non-default anyway
 
(merges are on hold for a bit in case of post-release disasters)

rrthomas:
> I'm ok with merging this, given my understanding:
> 
>     * the code will work on both old and new rsync on unix-y platforms

I have tested this, and confirm that it does.

>     * with the old code, external rsync doesn't seem to work on Windows, so it's not a regression

I confirm that I couldn't get it to work on Mingw. It should work on Cygwin, but I've not tested that. Cygwin should however be "unixy".

PR #869, opened 2023-03-14
NEWS: Add NEWS for 2.53.1

gdt:


NEWS.md
```diff
@@ -10,6 +10,17 @@ As of 2022, this file (NEWS.md at top level) is used for news.
...
 unison uses git history for that, and thus there is no changelog file.
 
+## Changes in 2.53.1
+
+Release 2023-03-17
+
+  * Improve stopping of update propagation.
+  * Enable VT input escape codes in Windows.
+  * Respect user-provided CFLAGS, CPPFLAGS, LDFLAGS, LDLIBS.
+  * Add build instructions in INSTALL.txt (and drop from manual).
```
tleedjarv, line 20:
```suggestion
  * Add build instructions in INSTALL.md (and drop from manual).
```

tleedjarv (approved):
Overall looks good. I would consider a few additions.

I think it is important to state that the repeat mode is now more fault tolerant. Previously it used to stop on any error; now errors that are expected to be temporary, like connection errors, do not stop the repeat mode.

Preferences 'force', 'prefer' and related ones no longer require specifying the full root string (see manual for more information).

Should we also add something like "Minor bugfixes and improvements"? At the very top of NEWS it is stated "We omit most bugfixes and minor improvements." but maybe it is good to repeat it here so that users don't take this list as exhaustive.

NEWS.md
```diff
@@ -10,6 +10,17 @@ As of 2022, this file (NEWS.md at top level) is used for news.
 Some software has a "changelog" file that records all changes, but
 unison uses git history for that, and thus there is no changelog file.
 
+## Changes in 2.53.1
+
+Release 2023-03-17
```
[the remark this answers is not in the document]
gdt: adopted, thanks.

NEWS.md
gdt: fixed, thanks.

gdt:
I think I've addressed all the comments.

tleedjarv:
Looks good to go.

PR #873, opened 2023-03-19
Print sync summary before starting propagation

tleedjarv:
When status messages are not disabled, a summary like the following is printed (and logged) before prompting for confirmation/starting propagation.
```
100017 items will be synced, 21 skipped
7.6 KiB to be synced from testa to testb
65.4 MiB to be synced from testb to testa
```


gdt (approved):
I think this is more useful than noise by a fairly wide margin, so I'm ok with it.

Issue #65, opened 2017-02-14
Slow performance with large files e.g. virtual machine disk images


klemens-u:
Hello, after some investigation I was happy to see that unison uses an intelligent algorithm which transfers only the modified part e.g. of a 15GB large virtual disk image.

But still, the progress is quite slow and generates a lot of I/O.

My test scenario:
- A 15GB virtual machine disk image on my local machine
- Already transfered once via unison to a remote machine
- Local start and shutdown of the virtual machine to provoke changes in the disk image file
- Running unison again.

This process takes more 10 minutes although only a few megabytes are transferred over the network.
Therefore I monitored I/O on both machines and the network traffic.
Here is what I found:
- Unison detects changes to the local file -> good.
- Full read of the 15GB file on the local machine -> ok, surly necessary to detect changes
- Full read of the 15GB file on the remote machine -> for comparison and changeset calculation?
- Network-transfer of the changed data (just a few MB) and simultaneous creation of a temp 15GB file on the remote machine. Since only little data is transferred over the network I assume most of the data is copied from the original remote file. This creates massive I/O on the target machine, since 15GB are read and written to the same disk... -> why isn't there a rsync-like "inplace" option?
- Full read of the 15GB file on the local side and full read of the 15GB file on the target side -> why? verify? This is very time and I/O consuming...

Apropos rsync - in reference to "Making Unison Faster on Large Files" (http://www.cis.upenn.edu/~bcpierce/unison/download/releases/stable/unison-manual.html#speeding):

Regarding the "copyprog" option: I couldn't detect any reference to rsync in the debug output (debug=all), or as a separate process. How can I make sure Unison is delegating sync of large files to rsync?

Furthermore, as I understand, rsync is only used for the initial first-time transfer. Is it possible touse rsync also for subsequent syncs?

Thanks!

klemens-u:
For comparison: rsync --inplace takes about half the time (around 5minutes) for the example above.

alkuzad:
@klemens-u Have you checked if you have `fastcheck=true` enabled ? And - if you need only one way synchronization - rsync can be less problematic. Unison is best used for multiple replicas.

brabalan:
@klemens-u `copyprog` only works for new files.

klemens-u:
@alkuzad: 
- yes, fastcheck is enabled. 
- I use Ubuntu Linux. Local machine = 14.04, remote machine 16.04, both 64bit. 
[...]

PR #876, opened 2023-03-20
Use reflink copy API within rsync protocol

tleedjarv:
Based on #577 (the first commit is from that PR).

With #577, it was very simple to extend this support to the rsync delta copy. This is performance-wise and by space requirement like rsync's `--inplace` but it's actually _safe_ to use. This would solve #65 in some configurations (see the caveat below).

The caveat is that it only works on platforms that export an API for fs block-level links (so-called reflinks) and on a filesystem that supports reflinks. Other platforms/filesystems continue using the old code.

(Note on implementation: the reflink mode currently works the same way as the plain old copy mode, by overlapping writes for data from the original file and for data from the network. This works well enough and makes the fallback to old mode extremely simple. An alternative implementation could be to reflink clone the entire file and then overwrite the changed blocks. This is not implemented because it would make the code more complicated and it's hard (at the moment) to see any general performance benefits.)

### DIFF
--- a/src/transfer.ml
+++ b/src/transfer.ml
@@ -468,7 +468,7 @@ struct
 
   (* For each transfer instruction, either output a string or copy one or
      several blocks from the old file. *)
-  let rsyncDecompress blockSize infd outfd showProgress (data, pos, len) =
+  let rsyncDecompress blockSize infd outfd ?copyFn showProgress (data, pos, len) =
     let decomprBuf = Bytes.create decomprBufSize in
     let progress = ref 0 in
     let rec copy length =
@@ -480,10 +480,23 @@ struct
         let _ = reallyRead infd decomprBuf 0 length in
         reallyWrite outfd decomprBuf 0 length
     in
+    let copyBlocks' offs length =
+      LargeFile.seek_in infd offs;
+      copy length
+    in
     let copyBlocks n k =
-      LargeFile.seek_in infd (Int64.mul n (Int64.of_int blockSize));
+      let offs = Int64.mul n (Int64.of_int blockSize) in
       let length = k * blockSize in
-      copy length;
+      begin match copyFn with
+      | None -> copyBlocks' offs length
+      | Some f ->
+          let fallback copied =
+            let offs = Int64.add offs (Uutil.Filesize.toInt64 copied)
+            and length = length - (Uutil.Filesize.toInt copied) in
+            copyBlocks' offs length
+          in
+          f (Uutil.Filesize.ofInt64 offs) (Uutil.Filesize.ofInt length) ~fallback;
+      end;
       progress := !progress + length
     in
     let maxPos = pos + len in
--- a/src/transfer.ml
+++ b/src/transfer.ml
@@ -356,7 +356,12 @@ struct
       only 31 bits.)
   *)
   (* Block size *)
-  let computeBlockSize l = truncate (max 700. (min (sqrt l) 131072.))
+  (* Block size (including the minimum and maximum limits used here) is
+     calculated similar to the rsync reference implementation. The main
+     difference here is rounding down to a power of 2 to allow block-level
+     links on filesystems that support it. *)
+  let computeBlockSize l =
+    1 lsl (truncate (max 10. (min (Float.round (log (sqrt l) /. log 2.)) 17.)))
   (* Size of each strong checksum *)
   let checksumSize bs sl dl =
     let bits =
--- a/src/copy.ml
+++ b/src/copy.ml
@@ -832,6 +832,14 @@ let transferFileContents
                   fspathTo pathTo fileKind srcFileSize outfd id in
               let eof =
                 Transfer.Rsync.rsyncDecompress blockSize ifd fd showProgress ti
+                  ~copyFn:(fun in_offs len ~fallback ->
+                    (* Flush the buffered output channel just in case since
+                       we manipulate the channel's underlying fd directly. *)
+                    flush fd;
+                    copyFileRange
+                      (Unix.descr_of_in_channel ifd)
+                      (Unix.descr_of_out_channel fd)
+                      in_offs len fallback (fun _ -> ()))
               in
               if eof then close_all infd outfd))
       else
--- a/src/transfer.mli
+++ b/src/transfer.mli
@@ -94,6 +94,13 @@ module Rsync :
            int                   (* block size *)
         -> in_channel            (* old file descriptor *)
         -> out_channel           (* output file descriptor *)
+        -> ?copyFn:              (* function for optimized copying *)
+           (   Uutil.Filesize.t  (* input file offset *)
+            -> Uutil.Filesize.t  (* data length *)
+            -> fallback:         (* default function for copying *)
+               (Uutil.Filesize.t (* bytes copied before fallback *)
+                -> unit)
+            -> unit)
         -> (int -> unit)         (* progress report *)
         -> transfer_instruction  (* transfer instruction received *)
         -> bool


src/Makefile.OCaml
```diff
@@ -129,6 +129,8 @@ ifneq ($(strip $(LDLIBS)),)
   CLIBS+=-cclib '$(LDLIBS)'
 endif
 
+CAMLCFLAGS+=-ccopt -D_FILE_OFFSET_BITS=64
```
gdt, line 132: Why is that define ok in general, vs on some subset of systems?   That feels like a non-POSIX thing that has to be guarded, but I'm unsure.

src/Makefile.OCaml
```diff
@@ -274,7 +277,7 @@ OCAMLOBJS+=main.cmo
 OCAMLLIBS+=unix.cma str.cma
 INCLFLAGS+=-I +unix -I +str
 
-COBJS+=osxsupport$(OBJ_EXT) pty$(OBJ_EXT) bytearray_stubs$(OBJ_EXT) hash_compat$(OBJ_EXT) props_xattr$(OBJ_EXT) props_acl$(OBJ_EXT)
+COBJS+=osxsupport$(OBJ_EXT) pty$(OBJ_EXT) bytearray_stubs$(OBJ_EXT) hash_compat$(OBJ_EXT) props_xattr$(OBJ_EXT) props_acl$(OBJ_EXT) copy_stubs$(OBJ_EXT)
```
gdt, line 280: spurious whitespace change?

src/copy_stubs.c
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  CAMLreturn(Val_long(ret));
+}
+
+
+#else /* defined(__linux__) || defined(__sun) */
+
+
+CAMLprim value unsn_copy_file(value in_fd, value out_fd, value offs, value len)
```
gdt, line 195: This looks like it fails always because of lack of support for better-than-copy.  Whether I guessed right or not a comment would be good.

gdt:
I don't really follow "safe".    Is the syscall that writes the new bits atomic, and only called once, so we are guaranteed that even with a crash the user will get the new bits or the old bits?   If that is the case, it would be good to make that argument in comments.  If not, I don't understand.

src/Makefile.OCaml
tleedjarv: It will do no harm on systems that don't support this. This is required to request `off_t` being a 64-bit type. The new function calls added with this PR use `off_t` only on platforms which have this as a 64-bit type. It will do nothing bad on 32-platforms that don't support it (but you won't get large file support either).

The OCaml runtime itself (and by proxy unison) has relied on this since 2002, on all platforms, unconditionally.

src/Makefile.OCaml
tleedjarv: No, long line. Scroll to the end, ` copy_stubs$(OBJ_EXT)` was added.

src/copy_stubs.c
tleedjarv: Yes, it's something of an idiom. I added comments on expected return values and exceptions, so I included this too.

gdt:
This is still draft, but having thought I am not comfortable with defining LARGEFILE macros on systems that are not known to use that.  While i see the argument that defining random macros on other systems shouldn't change behavior, it's unclean and feels unsafe in the future.

In general, we need a way to feature test and adapt, as autoconf would do.   That has lots of issues, mainly windows, but we could have a config.h that has #ifdef OS-that-uses-LFS and then #define, and include that widely, to centralize and have a comment home for such logic, if it isn't easy do to in the makefile.  But we already do stuff like that.

tleedjarv:
Is your worry about defining it globally, or about defining it unconditionally? These specifications have been in place for over 20 years (https://unix.org/version2/whatsnew/lfs.html and https://unix.org/version2/whatsnew/lfs20mar.html). Are there many systems that have not implemented these or are known to be broken?

Unison has already depended on this specific macro for over 20 years because OCaml and its Unix library define this macro unconditionally and Unison explicitly uses functions that are enabled for large files and require this macro to be defined. If this wasn't done then Unison couldn't handle files larger than 2 GB on 32-bit systems. The macro definition hasn't been needed in build scripts explicitly so far only because there has been no C code directly depending on it, and now there is in this PR. But indirectly it has always been there and it is required to support >2 GB on 32-bit systems.

Of course, since the new C code requiring this macro is highly platform-specific then it's maybe possible to rewrite it to not require this macro in the first place. And then there's also the possibility that we don't support 32-bit systems for functionality in this PR. That's not a bad option either, I think.

I did just learn that the portable way is to use `getconf LFS_CFLAGS` (and `getconf LFS_LDFLAGS` and `getconf LFS_LIBS`) instead of manually defining the macro. This is maybe equivalent to testing the feature set autoconf style like you mentioned? This is at least something I can fix easily.

gdt:
Where I'm coming from is that the entire notion of "LFS" is only applicable to some systems.  It just isn't true that all "32-bit" systems cannot handle large files without it.  Some, e.g. NetBSD, have chosen instead to define off_t etc. to be types that are big enough, rather than using int, and did so a really long time ago.  Just as time_t became int64_t.     So on NetBSD, there is no LFS option; there are just native syscalls that work.   (There are also old binary codepoints for ancient binary compat, not a unison issue).   The name _LARGEFILE_SOURCE shows up only in a comment under our /usr/include.

It's perfectly ok to define this on systems where that is what you are supposed to do.  But it is perpetuating the confusion that this is the sole path to being able to use large files.

I will have a deeper look, given that this is a standardized confusion-generating macro.

tleedjarv:
Ok, I see. I thought the notion of "LFS" is in SUS, but that does not mean that every SUS-compatible system needs to provide a _switch_. "LFS" may very well be its natural and only state. In that case `getconf LFS_CFLAGS` will just return nothing. This is perfectly valid.

tleedjarv:
> ... and did so a really long time ago.

Hehe, "LFS" is from 96 and was included in SUS v2 (97). Of course it's sad that it would still be opt-in on some systems (and I don't even know if that's the case).

Anyway, I think avenues worth exploring are using `getconf` (and it _will_ return nothing on most systems), and possibility to rewrite the affected code so that it can't be ambiguous on 32-bit systems. Lastly, I wouldn't exclude just disabling this code if the target is discovered to be 32 bits.

gdt:
My top-level point is that a system having a 32-bit CPU and there being an issue with large files that needs to be mitigated are not the same thing.   This is the fundamental confusion that I am bothered by.  The "large file issue" is because off_t is "int" on 32-bit CPUs (or rather, on ILP32 systems); it is not intrinsic to 32-bit CPUs.  So switching on CPU type is wrong.  Indeed, on PDP-11 Unix, one could use 2GB files in theory, on a machine where int was 16 bits.  Of course, nobody had a disk that big.

I will try to understand this better -- and I certainly have no problem defining this when appropriate.   I just don't understand when that is and isn't yet.  It may be that what POSIX says is that if you define that, you are guaranteed that off_t is at least 64 bits, and if you don't, you are only guaranteed that it is some integral type.  In which case, it's ok to define it.  (One could build a system on an x86_64 that uses 32 bits for off_t without that define and be conforming, I think.)

tleedjarv:
After some misunderstandings, I think we're converging now. It's not a hardware question, so not related to 32-bit CPUs (I used the word "system" in a very lax way). It's a libc/kernel ABI question primarily (and possibly API question secondarily, if it comes to that). Even though I basically meant ILP32 when I wrote "32-bit system", it's not directly related to ILP32 either. An ILP32 can have a working ABI and a non-ILP32 system can have a broken ABI. The expectation is that the ABI supports large files by default. If that's not the case then it may be possible to coerce it into a supporting state -- that's what this macro is for. And `getconf` is a way for the host itself to tell us (I don't know about cross-compilation but there's hardly any support for that as it is).

The original LFS rationale includes this statement:
```
Typical core utilities must be compiled in a "large" off_t compilation environment or must use the transitional APIs.
```

And these instructions for enabling "large" off_t (remember, it may be the default anyway): https://unix.org/version2/whatsnew/lfs20mar.html#A.3.2.4

> One could build a system on an x86_64 that uses 32 bits for off_t without that define and be conforming, I think.

I don't know for sure but would expect it to be true, so it's better to be prepared (and as stated, this has been the case with OCaml and its Unix library).

tleedjarv:
Rebased on #577 (the first commit is from that PR).

gdt:
I'd like to drop the '("inplace")'.   While I am now clear on the reflink concept, what's really going on is that a FS feature for deduplication/cow is being used, and inplace gives the impression (at least to me) that the existing file is being modified -- which if true (it's not) would be unsafe.    reflink is a word that may not be familiar to people, but it's what this is called, and it conveys that the code uses a feature that the user doesn't yet understand, rather than the wrong impression.

Also in commit 3 it says this is like rsync's --inplace but safe.  But it's not; rsync's inplace is about writing to the existing file, and this creates a new file.   With respect to FS semantics and rename(2), it's just like the old way, but with block-level references.  Which is great.   I don't want people to end up mislead, even if they are fuzzy to start with.  So I would say

"This has similar performance to rsync's "--in-place", but because a new file with references is created, rather than modifying the existing file, the atomicity guarantee provided by rename(2) is not compromised."

I am guessing that if the length copied is less than the native filesystem block size, then the reflink is going to fail, but that's ok because a) it's efficiency only and b) data less than the blocksize is small.   It's really MB that we are worried about and due to block size constraint to power of 2, it will be a multiple of things like 4KB or whatever the reflink granularity is.

Overall, this looks pretty safe, assuming it survives a bit of testing.

src/transfer.ml
```diff
@@ -356,7 +356,10 @@ struct
       only 31 bits.)
   *)
   (* Block size *)
-  let computeBlockSize l = truncate (max 700. (min (sqrt l) 131072.))
+  (* Make sure block size is a power of 2 to allow block-level links on
+     filesystems that support it. *)
+  let computeBlockSize l =
+    1 lsl (truncate (max 10. (min (Float.round (log (sqrt l) /. log 2.)) 17.)))
```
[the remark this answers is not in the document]
tleedjarv: I'd like to avoid adding a comment like this. It reads very much like:
```c
/* Assign 3 to a */
a = 3;
```

I checked the original rsync algorithm and the algorithm itself does not prescribe a block size. But it looks like you are right, the plausible explanation is that it's just "a good tradeoff of num blocks and block size". It does not allow the hash table to grow indefinitely while keeping the block size small enough to reduce the number of hash collisions.

I also checked rsync sources, and sure enough, there it is. The minimum of 700 has been there since the first commit in 1996.  Max block size used to be 2^29 but was changed to 2^17 in 2007. The block size itself is calculated as (quoting the comment) "The block size is a rounded square root of file length."

So I'd propose a comment like this: "The block size (including the minimum and maximum limits) is calculated similar to the rsync reference implementation. The main difference here is rounding down to a power of 2 to allow block-level links on filesystems that support it.".

tleedjarv:
I have updated the comment, commit message and the PR description. I've clarified that it is like "inplace" in terms of performance and space requirements, not that the target file is actually directly updated.

tleedjarv:
Marking this as draft until #577 is merged.

tleedjarv:
Rebased. Ready to merge.

PR #877, opened 2023-03-21
Add requirements for building the user manual

### LINKED ISSUE #545: Build instructions do not explain that lynx is needed
Perhaps they don't even explain how to bulld the docs, but I couldn't  find lynx mentioned in the manual.  Trying to build the docs, strings.ml wasn't updated and there was a lynx not found error.

tleedjarv:
Closes #545 (although still nothing about LANG/LC_TYPE plan)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -32,6 +32,16 @@ Please refer to instructions provided by the repository.
 
 - lablgtk3 and its prerequisites (ocamlfind, GTK 3 and its dependencies)
 
+##### Optional, for building the user manual
+
+- LaTeX
+- dvips
+- ps2pdf (included with Ghostscript, for example)
+- HEVEA (https://hevea.inria.fr/) (optional, for HTML and text formats)
+- Lynx (optional, for text format)
+- (for developers only) HEVEA and Lynx are required to re-build the manual
+  included in the Unison binary
+
 #### Building
 
 Building from source is as simple as executing:


INSTALL.md
```diff
@@ -32,6 +32,16 @@ Please refer to instructions provided by the repository.
...
 - lablgtk3 and its prerequisites (ocamlfind, GTK 3 and its dependencies)
 
+##### Optional, for building the user manual
+
+- LaTeX
+- dvips
+- ps2pdf (included with Ghostscript, for example)
+- HEVEA (https://hevea.inria.fr/) (optional, for HTML and text formats)
+- Lynx (optional, for text format)
+- (for developers only) HEVEA and Lynx are required to re-build the manual
```
gdt, line 42: I think this means "regen strings" needs these.   I feel like I'm nitpicking but it would be nice to have these hoisted to a section that addresses what it takes to prepare a PR.   Maybe that's just something earlier that explains that the manual dependencies are not optional for PR submission (but I guess latex/dvips/ps2pdf are not actually needed for that).

tleedjarv: Yes, it's the "regen strings" thing.
Agree, but I think such information is better kept in HACKING.md or CONTRIBUTING.md or similar. If this line is too provocative for this document then I can just remove it.

> but I guess latex/dvips/ps2pdf are not actually needed for that

I think so, but not sure. dvips/ps2pdf should definitely not be needed. I don't think hevea needs latex, but haven't checked.

gdt:
We don't have CONTRIBUTING.md now.  I just would like it to be clear that you need these for strings regen, even if you don't care about the manual.

PR #879, opened 2023-03-21
Allow syncing POSIX ACLs on Linux via xattrs

tleedjarv:
There is no direct support for syncing POSIX ACLs on Linux via the `acl` preference. Since Linux stores POSIX ACLs as extended attributes on files and directories, an easy way to sync POSIX ACLs (between Linux systems only) is to enable xattr syncing support for these attributes (`system.posix_acl_access` and `system.posix_acl_default`).

A user can sync POSIX ACLs on Linux by enabling the `xattrs` preference and adding an `xattrignorenot` preference with value `Path !system.posix_acl_*`. To optionally prevent other xattrs from being synced, also add an `xattrignore` preference with value `Path *` (or `Regex .*`).

### DIFF
--- a/doc/unison-manual.tex
+++ b/doc/unison-manual.tex
@@ -2193,6 +2193,19 @@ Unison currently supports the following platforms and ACL types:
 Not all filesystems on the listed platforms support all ACL types
 (or any ACLs at all).
 
+Synchronizing POSIX ACLs on Linux is not supported directly. However, it is
+possible to synchronize these ACLs with another Linux system by synchronizing
+extended attributes (xattrs) instead, because POSIX ACLs are stored as xattrs
+by Linux. This is disabled by default (see \sectionref{xattrs}{Extended
+Attributes - xattrs}).  A simple way to enable syncing POSIX ACLs on Linux is
+to enable the preference \verb|xattrs| and add a preference
+\verb|xattrignorenot| with a value \texttt{Path !system.posix\_acl\_*}.  The
+\verb|*| will be expanded to include both \verb|posix_acl_access| and
+\verb|posix_acl_default| attributes -- if you only want to sync either one,
+just remove the \verb|*| and type out the attribute name in full.  If you want
+to prevent other xattrs from being synced then add an \verb|xattrignore| with a
+value \texttt{Path *} (value \texttt{Regex .*} will also work).
+
 
 \SUBSECTION{Extended Attributes - xattrs}{xattrs}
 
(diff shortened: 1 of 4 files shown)


gdt (commented):
I would like to understand this better.   It seems like Linux implements the functionality of POSIX ACLs but does not implement the specified API.    This is synchronizing the storage format via another interface.  Thus, there is no path to synchronizing POSIX ACLs between a conforming system and Linux, and that doesn't seem like a great situation.

Does rsync deal with this by basically implementing the missing API? Something else?  Can it sync ACLs between POSIX systems and Linux?   Should we be doing that instead?

tleedjarv:
So you want the long story.

> It seems like Linux implements the functionality of POSIX ACLs but does not implement the specified API.

I believe it does implement the API. The reason for going the xattr way is that I don't intend to ever make the 'acl' preference support POSIX draft ACLs. Instead of just saying "sorry, you can't sync those", I offer the xattr way as an alternative. (This is a Linux issue. Most other platforms offer something better than POSIX draft ACLs anyway.)

Mind you, this is actually the same thing that `cp` in GNU coreutils does (via gnulib), so I'd call it a pretty good alternative. https://git.savannah.gnu.org/gitweb/?p=gnulib.git;a=blob;f=lib/qcopy-acl.c;h=0f4159b7fd954646d27946823e937264235bdfbb;hb=HEAD#l64 for those interested. Quoting from gnulib source:
> Rather than fiddling with acls one by one, we just copy the whole ACL xattrs

and then, if xattrs are not supported then this comment:

> no XATTR, so we proceed the old dusty way

So it's safe to say that this is what everyone using POSIX draft ACLs on Linux has been doing anyway, just not being aware of it.

This alternative is actually available for all (pseudo-)ACLs that are not implemented by the 'acl' preference, such as Samba server-side (attributes `security.NTACL` and `user.SAMBA_PAI`).

> Does rsync deal with this by basically implementing the missing API? Something else? Can it sync ACLs between POSIX systems and Linux? Should we be doing that instead?

I don't know what rsync does but a quick look at the source makes me believe that they use platform-specific APIs to get/set ACLs and transform those to/from their own internal serialization format. In my opinion this is a dead end for rsync. They can't support better ACLs with this code.

Very early on I made a decision not to have a semantic representation of ACLs within Unison with transformations to/from all platform-specific representations. That would be just too much work and possible maintenance and compatibility issues. With current design it is still possible to improve cross-platform syncing in future in backwards-compatible way if it starts making sense.

I don't intend to ever add direct support for POSIX draft ACLs (although I driveby-added limited support on BSDs...). Other contributors are welcome to do it of course but then I'd recommend them they better spend their effort migrating away from POSIX ACLs to something better.

Should Linux implement NFSv4 ACLs then that I will add into the 'acl' preference.

Issue #186, opened 2018-03-25
Bundle Windows download with required DLLs?


gitwah:
Hello! 
First of all, thanks for unison, I have been using it for quite some time on my linux machine! Recently I wanted to set it up on my parents' windows computer.
After downloading it and starting the .exe the program exits telling me it is missing some .dll. I don't know that much about windows but just downloading random .dll files from all over the web didn't seem like a good idea. My google-fu wasn't sufficient to find any other solution.
So basically what I am suggesting is: Could you include the missing .dll files in the download for the windows version / tell me why that would be a bad idea?

yannickmodahgouez:
Could you provide a more detailed error message ?

gitwah:
"This application has failed to start because libgtk-win32-2.0-0.dll was not found. Re-installing the application may fix this problem."

I remember more than one .dll-files missing, though. After getting one .dll and placing it in the right directory the error message returned requesting another dll file

JackeJR:
on the same note, I think the instructions for compiling for windows is quite outdated.
compiling with mingw is doable but I have never been successful with msvc.

qouoq:
I happen to stumble on the same problem today too.

Following the instructions provided on the [Unison wiki](https://alliance.seas.upenn.edu/~bcpierce/wiki/index.php?n=Main.UnisonFAQOSSpecific#windows), installing the [GTK+ for Windows Runtime](https://gtk-win.sourceforge.io/home/index.php/Main/Downloads) should install all the necessary libraries.

yannickmodahgouez:
@gitwah I assume you're using the graphical interface. Did you try the text interface ?

@JackeJR I just compiled the text version using Visual studio 2017 Community. Here is a PR that helped making it work : https://github.com/bcpierce00/unison/pull/195 and here is the binary : https://github.com/sticksh/unison/releases.
I agree that the windows documentation needs a rewrite.

PeterMosmans:
Slightly related: Here are some adventures on how to compile Unison on Windows using msys2 / mingw64: https://www.onwebsecurity.com/configuration/its-about-the-journey-compiling-64-bit-unison-gtk2-on-windows.html

gdt:
There is work underway to have a reliable windows build recipe which will supercede this request.

MelleNi:
Trying to run the latest unison-gui.exe I just downloaded, I am missing:
-libcario-2.dll
-libfreetype-6.dll
-libfontconfig-1.dll
[...]

PR #880, opened 2023-03-21
GHA: Fix Windows packaging (include missing DLLs)

### LINKED ISSUE #747: Revert workaround for mingw64-binutils 2.38 on Windows
In #745 there is a workaround to pin binutils because #744.  This ticket is just a reminder to remove that, at some point in the future when the underlying issue has been resolved.

tleedjarv:
As pointed out by https://github.com/bcpierce00/unison/issues/186#issuecomment-1477978644 the CI artifacts for Windows were not quite complete and couldn't be run on a clean install.

The CI Windows packaging step was broken. The reasons are unknown but one symptom which caused the packages to be incomplete was _something_(what?) mangling the PATH. Fixing the PATH manually also fixed the packaging (however, setting this PATH in earlier steps where other PATH entries are set does not work).

The binutils workaround is no longer available (Cygwin no longer ships binutils verson 2.37) so remove it. Fortunately, the issue that caused the workaround to be added seems to be gone.
Closes #747

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -295,6 +295,12 @@ jobs:
       name: "Windows: Package gtk"
       shell: bash
       run: |
+        # [2023-03] Setting this PATH here (and it has to be right here) is
+        # a workaround for an unknown issue (most likely something with GHA)
+        # causing MinGW binutils not to be found (while other binutils
+        # pre-installed in the GHA images may be found instead), which in turn
+        # causes the DLL extracting functions to silently fail.
+        export PATH="/usr/${{ steps.vars.outputs.MinGW_ARCH }}-w64-mingw32/bin":${PATH}
         ## package artifact(s)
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -295,6 +295,7 @@ jobs:
       name: "Windows: Package gtk"
       shell: bash
       run: |
+        export PATH="/usr/${{ steps.vars.outputs.MinGW_ARCH }}-w64-mingw32/bin":${PATH}
```
gdt, line 298: If I'm following, this is a workaround for a bug (which is not clearly understood) in GHA-land, in that PATH should have that included already, but has gotten broken.   This seems like a fine fix, but I'd like to see a comment in the code (rather than just the commit message) to explain (to future me, plus others) why a setting that shouldn't have to exist is present.

Issue #396, opened 2020-10-14
Root names are cropped after 9 characters in console mode


skjardenCode:
Hello,

I'm using Unison version 2.48.4 with Linux Debian 10 (Buster). My two configured roots in my profile are:

 - `root = /media/ironwolf-nas-data`
 - `root = /media/ironwolf-nas-lore`

When I start unison (using the console interface), the roots are displayed as follows:

```
[...]
Reconciling changes

ironwolf-...   ironwolf-...
dir      ---->            archive  [f]
```
This is making it hard to judge the correct direction of propagation of changes.

May I suggest giving the root names a bit more space, so they are not cropped at character 9?

Maybe there is an option for it or is it even possible to configure some kind of alias / alternate name for a root via the profile file?

Thanks a lot!

gdt:
(In general, bug reports have to be against 2.51.2 or later.)

This behavior seems to be present in 2.51.3_rc2.  Awaiting PR, with rationale for the choices made.

skjardenCode:
Thanks @gdt for settings labels and improving the issue title.

There are some other occasions where Unison, in my opinion, should not trim the lines so much, for example during the sync process when detecting changes:

```
Contacting server...
Looking for changes
/ MyRootDir/Files/Import/2018-01 - K....blckpud.SynchronizerService.java
```

The current examined path is, I think, about 120 characters long, but Unison just shows 70 characters. But other lines Unison is printing are way longer, character-wise, like

```
[BGN] Copying DevOps from /media/ironwolf-nas-16tb_data to /media/ironwolf-nas-16tb_lore
[END] Copying DevOps
UNISON 2.48.4 finished propagating changes at 22:49:10.69 on 14 Oct 2020
```

Should I open a new issue (improvement) for this or should we consolidate all line-abbreviations in one issue?

gdt:
You should create a change, test it, and submit a pull request.   In all seriousness, there is no army of people fixing bugs that don't personally other them, and this is pretty minor, so unless someone fixes it, I don't see it happening.   This is probably fairly hard to get right, as what matters to who varies, and opinions about super long lines also vary.

skjardenCode:
> In all seriousness, there is no army of people fixing bugs that don't personally other them, and this is pretty minor, so unless someone fixes it, I don't see it happening. 

[...]

PR #883, opened 2023-03-27
Increase max length of status messages in uitext

tleedjarv:
Based on some discussion in #396, and considering that any bigger changes are realistically not going to happen quickly, I propose this intermediate solution. (Temporary things that become permanent, you know the story...)
It is _not_ a solution to #396, just to one bit of it.

### What's in this patch?

With VT100 terminal emulators we can control automatic line wrapping mode per line, and this is exactly what this patch does for the status line while scanning updates. By disabling line wrapping, the length cap (currently at 70) can be removed. Instead, status lines will be chopped off beyond the width of terminal.

This is supposed to work on all proper VT100 emulators (more info in commit message). For some, I've explicitly disabled this function; for others, very long status lines will wrap and be half-overwritten by the next. This may cause ugly output but is not harmful in any other way.

### DIFF
--- a/src/uitext.ml
+++ b/src/uitext.ml
@@ -1163,19 +1163,40 @@ let displayWaitMessage () =
   if not (Prefs.read silent) then
     Util.msg "%s\n" (Uicommon.contactingServerMsg ())
 
+(* Most modern VT100 terminal emulators (and some ANSI) are able to switch
+   automatic line-wrapping off and on by control sequences ESC[?7l and ESC[?7h.
+   This here is a very blunt heuristic to filter out some that can't do it or
+   use a different control sequence. It does not need to be exact, as long as
+   it covers the vast majority of supported systems. *)
+let termNowrapOk =
+  System.termVtCapable Unix.stdout &&
+  let s = try System.getenv "TERM" with Not_found -> "" in
+  not (
+    s = "dumb"
+    || s = "emacs"
+    || Util.startswith s "sun"
+    || Util.startswith s "cons"
+    || Util.startswith s "eterm"
+    || Util.startswith s "cygwin"
+    || Util.startswith s "dvtm"
+  )
+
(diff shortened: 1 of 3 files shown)


src/uitext.ml
```diff
@@ -1163,19 +1163,40 @@ let displayWaitMessage () =
...
+      let mid = (max_len - 3) / 2 in
       let l = String.length path in
       if l <= max_len then path else
       String.sub path 0 (max_len - mid - 3) ^ "..." ^
       String.sub path (l - mid) mid
     in
     let c = "-\\|/".[truncate (mod_float (4. *. Unix.gettimeofday ()) 4.)] in
-    Util.set_infos (Format.sprintf "%c %s" c path)
+    if termNowrapOk && not (Prefs.read dumbtty) then
+      Util.set_infos (Format.sprintf "%c \027[?7l%s\027[?7h" c path) ~clr:"\r\027[K\r"
```
gdt, line 1197: Shouldn't the escape codes be coming from a capability?   It's not just "terminal emulators" but terminals, really, and there are a lot of termcap/terminfo things out there.

tleedjarv: > Shouldn't the escape codes be coming from a capability?

Ideally, yes. In practice, here, I don't think it matters. It appears that pretty much anything modern emulates VT100/xterm and supports these sequences (or at the very least silently ignores them). Some exceptions I've explicitly brought out above.

I thought about it but I don't want to make terminfo, termcap or some other thing a dependency (because it would have to be something like ncurses or s-lang then). It's just not worth it.

I did query the terminfo database and I'm satisfied that the current approach is not going to cause any harm.

> It's not just "terminal emulators" but terminals, really, and there are a lot of termcap/terminfo things out there.

I don't think we need to specially support actual terminals (which ones do you have in mind?). Also, any funny stuff is turned off by `TERM=dumb` or `-dumbtty`.

PR #886, opened 2023-03-29
Improve remote shell terminal handling and password requests in GUI

### LINKED ISSUE #885: unison-gtk waits 30 seconds when server sends text
My university has some silly message that appears every time I log in via SSH.  In the last year, unison-gtk has started to post this message in a new window, seeming to wait for a response:

Screenshot from 2023-03-28 12-01-41

There are two negative effects of this change:

1. The message doesn't appear for _30_ seconds, before which Unison seems to hang so I have to wait 30 seconds before I even get the "Connected" message and the changes start to calculate.
2. I have to click OK within a short period, or else there is a timeout.

Practically speaking, this means that I have to standby for 30 seconds each time I want to synchronize my files, waiting for the message window to open, so that I can press OK.  This was never a problem previous to the existence of this window.  If I use unison in a terminal there is no problem and things proceed normally.

tleedjarv:
This is something that I'd put off for some time, thinking it was not too relevant. #885 made it real. More description in commit messages.

Fixes #885

_For reviewers: remote.ml diff you probably want to view without whitespace, for other files diff with whitespace looks better._

### DIFF
--- a/src/terminal.ml
+++ b/src/terminal.ml
@@ -287,24 +287,114 @@ let close_session pid =
 
 let (>>=) = Lwt.bind
 
-let escRemove = Str.regexp
-   ("\\(\\(.\\|[\n\r]\\)+\027\\[[12]J\\)" (* Clear screen *)
-  ^ "\\|\\(\027\\[[0-2]?J\\)" (* Clear screen *)
-  ^ "\\|\\(\027\\[!p\\)" (* Soft reset *)
-  ^ "\\|\\(\027\\][02];[^\007]*\007\\)" (* Set console window title *)
-  ^ "\\|\\(\027\\[\\?25[hl]\\)" (* Show/hide cursor *)
-  ^ "\\|\\(\027\\[[0-9;]*m\\)" (* Formatting *)
-  ^ "\\|\\(\027\\[H\\)") (* Home *)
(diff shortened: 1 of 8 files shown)


src/terminal.ml
```diff
@@ -287,24 +287,114 @@ let close_session pid =
...
-   ("\\(\\(.\\|[\n\r]\\)+\027\\[[12]J\\)" (* Clear screen *)
-  ^ "\\|\\(\027\\[[0-2]?J\\)" (* Clear screen *)
-  ^ "\\|\\(\027\\[!p\\)" (* Soft reset *)
-  ^ "\\|\\(\027\\][02];[^\007]*\007\\)" (* Set console window title *)
-  ^ "\\|\\(\027\\[\\?25[hl]\\)" (* Show/hide cursor *)
-  ^ "\\|\\(\027\\[[0-9;]*m\\)" (* Formatting *)
-  ^ "\\|\\(\027\\[H\\)") (* Home *)
-
-let escSpace = Str.regexp "\027\\[\\([0-9]*\\)C"
+(* OpenSSH on Windows is known to produce at least the following escape
```
gdt, line 290: While this seems like a big improvement, I wonder if this happens always, or only if TERM supports these.

PR #891, opened 2023-03-31
makefiles: Avoid bashism in make clean

### LINKED ISSUE #808: build/clean leaves files
After building 2.53.0alpha2 on NetBSD 9, and then cleaning, a few files remain, as shown by `git status --ignored`:
```
Ignored files:
  (use "git add -f <file>..." to include in what will be committed)
        src/lwt/generic/lwt_unix_impl.o
        src/lwt/lwt.o
        src/lwt/lwt_unix.o
        src/lwt/lwt_util.o
        src/lwt/pqueue.o
        src/system/generic/system_impl.o
        src/system/system_generic.o
        src/ubase/depend
```

gdt:
src/Makefile.OCaml used "*.{o,obj}" which is a bash extension.  make by default uses the system /bin/sh, and that in general does not support bash extensions.

Fixes #808 as tested on NetBSD.

### DIFF
--- a/src/Makefile.OCaml
+++ b/src/Makefile.OCaml
@@ -454,15 +454,15 @@ clean::
 	-$(RM) -r *.o core gmon.out *~ .*~
 	-$(RM) -r *.obj *.lib *.exp
 	-$(RM) -r *.tmp *.bak?.tmp .*.bak?.tmp
-	-$(RM) system/*.cm[iox] system/*.{o,obj} system/win/*~
-	-$(RM) system/generic/*.cm[iox] system/generic/*.{o,obj} system/generic/*~
-	-$(RM) system/win/*.cm[iox] system/win/*.{o,obj} system/win/*~
-	-$(RM) fsmonitor/*.cm[iox] fsmonitor/*.{o,obj}
+	-$(RM) system/*.cm[iox] system/*.o system/*.obj system/win/*~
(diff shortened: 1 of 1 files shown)


src/Makefile.OCaml
```diff
@@ -484,15 +484,15 @@ clean::
 	-$(RM) -r *.o core gmon.out *~ .*~
 	-$(RM) -r *.obj *.lib *.exp
 	-$(RM) -r *.tmp *.bak?.tmp .*.bak?.tmp
-	-$(RM) system/*.cm[iox] system/*.{o,obj} system/win/*~
-	-$(RM) system/generic/*.cm[iox] system/generic/*.{o,obj} system/generic/*~
-	-$(RM) system/win/*.cm[iox] system/win/*.{o,obj} system/win/*~
-	-$(RM) fsmonitor/*.cm[iox] fsmonitor/*.{o,obj}
+	-$(RM) system/*.cm[iox] system/*.o system/*.obj system/win/*~
+	-$(RM) system/generic/*.cm[iox] system/generic/*.o system/generic.obj system/generic/*~
```
tleedjarv, line 488:
```suggestion
	-$(RM) system/generic/*.cm[iox] system/generic/*.o system/generic/*.obj system/generic/*~
```

gdt: good catch, thanks.

PR #895, opened 2023-04-03
GHA: Simplify published package naming

tleedjarv:
Since there are no longer tens and tens of packages published, there is no need to automagically construct package names. Simplify the logic and form of names. Most importantly, there is no need to include the compiler's version (nor options) in package name.

Previous names looked like:
```
unison-v2.53.2+ocaml-4.08.1+x86_64.linux.tar.gz
unison-v2.53.2+ocaml-4.14.0+mingw32c+i386.windows.zip
unison-v2.53.2+ocaml-4.14.0+mingw64c+x86_64.windows.zip
unison-v2.53.2+ocaml-4.14.1+x86_64.linux.static.tar.gz
unison-v2.53.2+ocaml-4.14.1+x86_64.linux.tar.gz
unison-v2.53.2+ocaml-4.14.1+x86_64.macos-10.15.tar.gz
Unison-v2.53.2.ocaml-4.14.1.macos-10.15.app.tar.gz 
```

Now names would look like:
```
unison-2.53.2-linux-x86_64-static.tar.gz
unison-2.53.2-linux-x86_64.tar.gz
unison-2.53.2-macos-x86_64.tar.gz
unison-2.53.2+ocaml4.08-linux-x86_64.tar.gz
unison-2.53.2-windows-i386.zip
unison-2.53.2-windows-x86_64.zip
```
(dev snapshots)
```
unison-git_91bed2cd-linux-x86_64-static.tar.gz
unison-git_91bed2cd-linux-x86_64.tar.gz
unison-git_91bed2cd-macos-x86_64.tar.gz
unison-git_91bed2cd+ocaml4.08-linux-x86_64.tar.gz
unison-git_91bed2cd-windows-i386.zip
unison-git_91bed2cd-windows-x86_64.zip
```
(I also want to add the underlying release version to the dev snapshots, but that's for future)

The name forms are proposals and subject to discussion (or bikeshedding), of course.


gdt:
This seems conceptually ok, with two comments/questions:
  - I wonder why we are publishing a 4.08 build at all.  (I can see why we build for CI)
  - "Linux" is a bad name (although pre-existing),  because <rms but he is correct>it's the kernel not the operating system</>.   More importantly, these are for particular versions that assume a lot, and thus we should perhaps be labeling it ubuntu if that's what it is.   (The fact that we are building for ubuntu instead of Debian within the GNU/Linux family I view as a bug, but that's out of scope for this PR.)

PR #897, opened 2023-04-06
GHA: ubuntu-18.04 is gone

tleedjarv:
The current workflow won't run anymore.

Instead of just bumping ubuntu-18.04 to 20.04, I took a different route. These builds are for testing build compatibility, nothing more. Publishing the resulting binaries for users to download is a nice bonus only. So I went ahead and used GHA's Docker feature to run these builds on simulated older ubuntu releases (16.04 for now). This seems to work rather well, just sometimes a tiny bit slower due to having to install additional system packages. The published binaries are now also much more widely compatible with different systems.

I based this PR on #895 (the first commit here).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1104,3 +1104,71 @@ jobs:
     - run: opam exec -- make src UISTYLE=text NATIVE=false
 
     - run: opam exec -- make test
+
+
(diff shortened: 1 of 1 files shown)


gdt:
I merged other stuff so this maybe needs a rebase?

Issue #816, opened 2022-10-30
Progress calculator treats deletions as zero-cost (or only considers network transfers?)


gdt:
Reported privately to me by someone running 2.53.0alpha1 on NetBSD, all works, except that the progress meter treats deletions as free, even though it can take a fs a while to sync the metadata.   If one is syncing some data and deleting a lot (e.g. 20 GB), then the reported progress can go to 100% quickly and then stay there for a while.

It may turn out that the progress mechanism is currently only about the network and not the filesystem, and that better estimates are hard, but this is a defect ticket about the user-observed behavior.   Probably it deserves `effort-veryhigh` but we already have too many labels.

tleedjarv:
Yes, the issue is broader than just deletions. It is currently based on bytes copied between replicas (local or over the network). All metadata-only actions are currently excluded from progress. All data-crunching actions (when no copying is done) are also excluded; here I mean fingerprinting (for example, the paranoid check and the resume-partial-transfer check).

I have been working on improving progress calculation and on relieving this issue (because it has been very noticable). There will probably by a PR or two. No ETA, though (pun intended).

donjan:
Had this issue just now after deleting one single 200 GB directory and then syncing over USB with `fastcheck = true`.
The UI was hanging for roughly an hour with no progress indication. System monitoring showed 40-50 MB/s disk read throughout (no disk write).
I assume that Unison needs to read the whole thing first for a safe removal or consistency, and speeding up simple deletion is a different issue anyway.
In the context of this report, a progress meter would have been valuable UX, even a separate one or a scrolling log.
101 words for a "me too", sorry :)

gdt:
I'd appreciate test reports with 2.53.3 (or an RC), given recent improvements.

tleedjarv:
> What in particular were the changes?

Regarding this particular issue, the only change is that progress status now includes the number of updates done/total. This brings some relief if there are many such deletions (for example). It does not help with one big deletion.

> There was a 'starting' line that was new to me, but overwritten by the `[BGN]` one.

Yes, that is new, too. It is displayed until more status can be displayed. Your test with a single deletion does not provide more status (yet, until this issue is closed, that is).

0-wiz-0:
That sounds like good incremental changes, thank you!

gdt:
It seems the basic point is still valid, even if the situation has been substanitally mitigated.   I'll leave this open until we have a %-progress metric that somehow is a composite of network transfer and fs ops.

PR #899, opened 2023-04-12
Improve sync progress status display (mainly TUI)

tleedjarv:
### Motivation

There are more details in commit messages and code comments, but the basic premise is that the current ETA calculation is just not good enough, for various underlying reasons. To improve calculation of ETA, we need to improve calculation of the progress rate.

The nature of actions taken during a sync makes the rate of progress highly volatile and this is not something that can be easily changed (but maybe improved upon in future). This PR focuses on trying to get reasonably usable results out of the volatile data. In other words, it's mostly about algorithms.

### Description

Two new progress rates are calculated: a less stable one which is suitable for displaying to users and a more stable one which is suitable for ETA calculation. The less stable rate is calculated as an exponential moving average (aka exponential smoothing) of the actual sync rate (which is simply calculated as delta synced bytes over delta time). The more stable rate is calculated as an exponential moving average of the former rate (double smoothing) with an additional Gaussian filter to exclude tiny changes. I've done countless tests to end up with these algorithms and these constant values. It's impractical to go any further without an automated setup... and I'm not going to do that (just yet).

The ETA itself is no longer calculated to a resolution of 1 second (except for ETA under 2 minutes) because even with very stable calculated rate, ETA for longer syncs can change several (tens of) seconds for just a tiny rate change.

Further details are included in collapsed sections.

<details>

<summary><b>Open this section for detailed motivation with illustrations...</b></summary>

<br/>

_But is the current state really that bad?_

Yes, it is. This is a graph of a real sync, showing the actual rate and the rate used for ETA calculation. Here and onwards, all graphs represent real syncs not simulations, x-axis is always in seconds, y-axis is in bytes/s, and "Current rate" indicates the actual "raw" sync rate at any moment of time.

image

_But this isn't so bad..._  Sure it is, let's just zoom in.

image

The ETA is calculated at 2-3x times the actual rate. A 3 minute sync is displayed as 1 minute ETA.

_Does it really matter?_  Yes, it does. The examples here are of syncs lasting a few minutes only. It gets much-much worse for syncs lasting hours.

_Are the new algorithms any better?_  Yes, this is the same graph with the ETA rate calculated by the new algorithm.

image

And zoomed in...

image

_Ok, so that does look better. But is the code complexity worth it?_

I believe it is (but simpler code achieving the same or better results is even better!). This is a real-world example of raw input data:

image

Looks like the sync rate is zero most of the time? But it really isn't; look at the scale of y-axis! Zooming in, we see:

image

Simpler algorithms are simply unable to smooth out these fluctuations while at the same time remaining capable of responding to the huge changes (between ~1MB/s to tens and hundreds of MB/s). This is what a simple moving average would look like when zoomed in (the plot lines are thinner to better show detail). This is not good enough for ETA calculation.

image

The most obvious downside of smoothing the rate for ETA calculation is that actual average rate changes (not just fluctuations) are also smoothed out. This can be seen by looking at the huge changes in the current rate on graphs above. The actual rate change looks like a right angle; just look at how long it takes for the average rate to catch up to that change.  The real trick is finding the balance between stability and responsiveness. ETA calculation requires _very_ stable input.

</details>

<details>

<summary><b>Open this section for more illustrations...</b></summary>

uigtk already had implemented a kind of exponential rate smoothing algorithm. But it was not used for calculating the ETA because it was still much too volatile. Here are some comparisons.

Zoomed out:
image

And zoomed in:
image

The zoomed in section also illustrates well why there is a need for double smoothing.

It's not always fun and games, though. Here, the algorithm just can't produce a stable rate because the volatility in input data is too much. Note that in this case the old ETA algorithm would possibly produce a better result, right up until the average rate changes around 225 seconds mark.

image

Here, the old ETA calculation seems to be more stable:
image

Another example of the extreme rate volatility. Note the scale on y-axis.
image

image

Section of previous graph zoomed in, comparing simple moving average (avg1) to the double-smoothed rate (note that the legend has changed):
image

</details>

### There's more

This PR also in a very minor way improves #816 (I didn't see a reason to split it out into a separate PR -- but it can be easily done). The status display in TUI now shows the number of updates in addition to bytes.

Overall, the status display in TUI has become rather busy with this. For short syncs this could be distracting. For longer syncs, I consider the additional info valuable. It now looks like this:
```
 20%  0/2  (11.6 MiB of 57.1 MiB)   6.1 MiB/s    00:00:05 ETA
```
("0/2" indicates number of completed and total items). Just recently the status display looked like this:
```
 20%  00:05 ETA
```

### DIFF
--- a/src/uicommon.ml
+++ b/src/uicommon.ml
@@ -467,6 +467,109 @@ let addIgnorePattern theRegExp =
   Lwt_unix.run (Globals.propagatePrefs ())
 
 (**********************************************************************
+                     Statistics for update progress
+ **********************************************************************)
+
+(* This seemingly very complex code for calculating the progress rate
+   and ETA has partly to do with Unison currently not tracking progress
+   very accurately. Several potentially very time-consuming operations
+   are not tracked at all: hashing files before and after the copy, for
+   example. The entire amount of work may not even be known in advance
+   when continuing partial transfers after the previous sync has been
+   interrupted. This makes it very difficult to provide meaningful rate
+   and ETA information. The code below is the current best approximation.
+   The way to simplify this code here is to first and foremost improve
+   progress tracking and reporting. *)
+
+module Stats = struct
+
+let calcETA rem rate =
+  if Float.is_nan rate || Float.is_nan rem || rem < 0. then "" else
+  let t = truncate (rem /. rate +. 0.5) in
+  (* Estimating the remaining time is not accurate. Reduce the display
+     precision (and reduce more when longer time remaining). *)
+  let h, (m, sec) =
+    if t >= 3420 then
+      let u = t + 180 in u / 3600, (((u mod 3600) / 300) * 5, 0)
+    else
+      0,
+      if t >= 2640 then ((t + 180) / 300) * 5, 0
+      else if t >= 1800 then ((t + 119) / 120) * 2, 0
+      else if t >= 120 then let u = t + 15 in u / 60, ((u mod 60) / 30) * 30
+      else t / 60, t mod 60
+  in
+  Printf.sprintf "%02d:%02d:%02d" h m sec
+
+let movAvg curr prev ?(c = 1.) deltaTime avgPeriod =
+  if Float.is_nan prev then curr else
+  let a = c *. Float.min (1. -. exp (-. deltaTime /. avgPeriod)) 1. in
+  (* Simplified from a *. curr +. (1. -. a) *. prev *)
+  prev +. a *. (curr -. prev)
+
+type t = (* abstract in mli *)
+  { mutable t0 : float;
+    mutable t : float;
+    totalToComplete : int64;
+    mutable completed : int64;
+    mutable curRate : float;
+    mutable avgRateS : float;
+    mutable avgRateDoubleSGauss : float;
+  }
+
+let gaussC = 2. *. (0.025 ** 2.)
+let avgPeriodS = 4.0
+let avgPeriodD = 3.5
+let calcPeriod = 0.25
+
+let init totalToTransfer =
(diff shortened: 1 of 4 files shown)


acolomb:
Very nice work indeed and a big thumbs up for tackling this seemingly unimportant detail which can however be a real PITA when waiting for a sync to finish before leaving work to catch a train (based on actual experience). And you did a great job measuring, understanding and verifying the situation and your improvements, kudos!

Just a thought I had been pondering since many years back, which might serve well here as an inspiration for even better algorithms.

You analyze three different intervals. 1. From start up to the current time. 2. The last minute. 3. The last ten seconds.

For each interval, analyze the regularly sampled progress rate, taking the average (mean) value as well as the standard deviation within the interval. Then pick the interval with the lowest standard deviation and use its mean value for ETA calculation.

Whether the standard deviation is the right factor to decide on and how the intervals should be tweaked is of course still to be examined, but this is the general idea. The goal is to switch between the different estimations and rely on short-term rates only when they are currently rather steady, and fall back to longer periods in case of fluctuations. Actual, longer lasting changes in progress rate will lead to the short-term recent data being preferred again.

Hope it gives you some fresh ideas to try on the data. Happy to answer any questions if I didn't word it clearly enough.

acolomb:
Well if you still have the raw data of your plots available, different algorithms can be tested independent of an OCaml implementation quite easily. At least for your tested scenarios. Not that I'd have time for experiments right now, but do hold on to the data for some later time maybe.

tleedjarv:
Yes, experimenting would be completely independent and I encourage anyone interested in this to give it a shot in any language/environment. I do have the raw data and it's very easy for me to provide a debug build for anyone to collect and save their own raw data.

PR #900, opened 2023-04-15
Add missing version number to a removed pref

tleedjarv:
This was missed in the original commit. **Please verify correct version before merging.**


gdt:
version looks right to me

PR #904, opened 2023-04-22
Fix incomplete sync when `maxthreads = 1`

tleedjarv:
Fix an issue reported on mailing list. Sync would always finish after only 2 transfers and report all other transfers as "not started". This issue only manifests with preference `maxthreads = 1`.


tleedjarv:
Added self-tests for this functionality.

PR #912, opened 2023-04-29
Fix a compiler warning

tleedjarv:
Fix a warning that is obviously incorrect at this location but will nevertheless cause problems with `-Werror` and simply pollute the build logs.

```
./lwt/lwt_unix_stubs.c:409:13: warning: "callback" is deprecated: use "caml_callback" instead
  409 | CAMLprim value init_lwt (value callback) {
```


gdt:
This is minor enough that I'm going to merge it during the 1-week hold for emergency fixes.

PR #916, opened 2023-05-06
Makefile improvements, update build instructions

tleedjarv:
Another series of small incremental quality-of-life improvements. I kept the commits separate so each can be mostly self-explanatory. Review is also easier commit by commit and when moved blocks are highlighted.

For time being I've given up on compatibility with other make implementations since the current makefiles require GNU make in a big way.

It seems like a lot of change but it really isn't. It's mostly quite trivial changes and should be simple to review. This is by no means the end of it either.


src/Makefile.OCaml
```diff
@@ -39,7 +39,9 @@ endif
 # Another try, Feb 2011, suggested by Ron Isaacson
 # OCAMLLIBDIR=$(shell ocamlc -v | tail -1 | sed -e 's/.* //g' | tr '\\' '/' | tr -d '\r')
 # 2023: "-where" has existed since OCaml 3.01 (2000)
-OCAMLLIBDIR=$(shell ocamlc -where)
+# OCAMLLIBDIR=$(shell ocamlc -where)
+# Since OCaml 4.08 (2019) we have an even better option
```
gdt: How much should we leave breadcrumbs?  My bias without thinking is not much, because that's what commit history is for.   Given that we require 4.08+, it seems that using features in the 4.08 interface spec is legitimate without comment.   Or perhaps just "# OCaml 4.08 requires -config-var".   I do lean fairly hard to not having commented-out code.

tleedjarv: I don't think such breadcrumbs should be left either. Looking at the already existing breadcrumbs, I decided to give it a break this time and do a "hybrid" approach. To not give any false impressions when looking at commit history, I added one final breadcrumb here and then removed all of them in the next commit.

PR #920, opened 2023-05-11
Makefile: remove UISTYLE

### LINKED ISSUE #667: Rationalize unison as CLI and unison-gui as GUI for program names in all builds
This state should ensure as a side effect that CI artifacts will have a CLI binary that will work on a system without the GTK libraries, or with the misplaced or wrong kind of GTK libraries.

Keep in mind, thought, that unison doctrine is that CI is for CI and packaging systems should provide packages.  However building from source should be clean and aligned with packaging norms, and that's what this ticket is about.

tleedjarv:
It's exactly what it says on the tin. Remove UISTYLE and build UIs as unique targets, together or separately. As a bonus, I also brushed up rules for building fsmonitor in the same style.

Now, the following applies:
```sh
make           # build the TUI + GUI if lablgtk3 found + macUI if on macOS + fsmonitor if impl available for the platform
make tui       # build just the TUI
make gui       # build just the GUI; fails if lablgtk3 is not available
make macui     # build just the macUI; fails if xcode tools are not available
make fsmonitor # build just the fsmonitor; works only if fsmonitor impl available for the platform
               # (building the fsmonitor can be forced by manually setting FSMDIR but
               # compilation will naturally fail if the code is invalid for the platform)
```
```sh
make src/unison           # (or make unison in the src dir) equivalent to make tui; primarily for internal use
make src/unison-gui       # (or make unison-gui in the src dir) equivalent to make gui; primarily for internal use
make src/unison-fsmonitor # (or make unison-fsmonitor in the src dir) equivalent to make fsmonitor; primarily for internal use
```

The CLI/TUI is named `unison`, the GUI is now named `unison-gui`.

For comparison, the _old rules_ were roughly:
```sh
make                  # if on macOS then build macUI; otherwise, if lablgtk3 is detected then build GUI, else build TUI; the latter two both named `unison`
make UISTYLE=text     # build just the TUI, named `unison`
make UISTYLE=gtk3     # build just the GUI, named `unison`
make UISTYLE=mac      # build just the macUI
make unison           # (only in src dir) same as simply `make` (also can set UISTYLE)
make unison-fsmonitor # (only in src dir) build just the fsmonitor, if impl available
# since the TUI and GUI styles both would build the same target then rebuilding
# in another UI style may become impossible without first strategically removing files
```

There are countless ways of improvement, I'm sure. This is a first step to get things moving, then we can tweak going forward.

**_Note:_** Need to update any wiki pages that still mention UISTYLE (if any).

Some remarks:

The toplevel makefile is a smell but I didn't do anything about it yet, just marked it non-parallel and delegate all "unknown" targets to the src makefile. The src makefile fully works with parallel builds.

It's all GNU make. I don't know if I've made compatibility any worse but it certainly was not compatible with non-GNU make before this either.

Note that there is a minor bug in doc/Makefile and man/Makefile where EXEC_EXT is always undefined (only relevant for Windows). I'm not going to fix it now because in practice, due to way recipes are written, things will work anyway, even on Windows.

Closes #667?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -165,7 +165,8 @@ jobs:
 
     - shell: bash
       run: |
-        opam exec -- make src UISTYLE=text
+        opam exec -- make tui
+        opam exec -- make fsmonitor
         # stage
         # * notes: darwin/macos doesn't build `unison-fsmonitor`
         for file in ${PROJECT_EXES} ; do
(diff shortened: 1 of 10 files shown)


gdt:
tui/gui builds work on NetBSD.  Between that and your testing, this is a vast impovement!

PR #929, opened 2023-05-29
Remove old dev-testing targets from the makefile

tleedjarv:
I think the time has come to remove these targets from the makefile as they're probably no longer useful even to the original authors. If they still are, against all odds, then they should be re-introduced (in a separate include-makefile) _together_ with all other files they need (profiles, for example) so that they could be run by anyone.


gdt:
Fine with me; will wait 72h for objections.

PR #930, opened 2023-05-29
Simplify the docs build process

tleedjarv:
Proposed simplification of the docs building process. While useful on its own, this is in preparation of further rototill of the entire makefile.

There are two changes included in a single commit.

1. `postproc` does not seem to make any difference anymore. Remove it completely. As you can see from `postproc` sources, it could never have been a critical component...

2. Build by pdflatex directly, instead of via dvips and ps2pdf. Still keep producing the PS output, even though that's hardly needed anymore. (DVI output is still produced in the "text mode" but I don't view this as a problem because DVI output should be considered a temporary by-product (also, this output is perfectly usable).)

There's one caveat. The old recipe explicitly produced the output for the letter paper size. The new recipe produces output for paper size based on the locale (I think). I don't think this is an issue but I'm sure this can be changed if needed.


tleedjarv:
Added a commit to remove hardcoded paths from the process.

gdt:
I think dvi is useful in its own right, but since "gmake docs" builds it, that's ok.

This seems like a good change; it's definitely true that pdflatex is more normal than dvi->ps->pdf.

I'm fine with respecting local default paper size.

Not really about this PR, but it seems that latex is not run enough times for the text version, and if it were, then maybe pdflatex wouldn't have to be run so many times.   It would also be nice to have targets for the various formats, even if all: just depends on all of them, but again not about this PR.

PR #938, opened 2023-06-19
Another round of docs makefile improvements

tleedjarv:
More details in commit messages. I have kept the commits separate for easier review. The first three commits could easily be separate into a separate PR if needed.

The dvi output is currently removed but I don't see that as a loss. What would be the use case for dvi? If needed, I can add it back but it will again complicate the makefile because it can't be built in parallel with other targets (it conflicts by trying to write to the same aux and toc files).

Edit: I also thought of adding one more commit:
```diff
diff --git a/doc/Makefile b/doc/Makefile
index f406966e..0d4ffc6f 100644
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -28,8 +28,12 @@ unisonversion.tex: ../src/Makefile.ProjectInfo
 # jobname (or, output name for hevea) so that files for non-text versions are
 # not overwritten. Here, suffix -text is used.
 
-%-text-directives.tex:
-       printf '$(TEXDIRECTIVES)\\textversiontrue\\draft$(DRAFT)' > $@
+%-text-directives.tex: TEXTVERSION = true
+%-directives.tex: TEXTVERSION = false
+
+.NOTINTERMEDIATE: %-directives.tex
+%-directives.tex:
+       printf '$(TEXDIRECTIVES)\\textversion$(TEXTVERSION)\\draft$(DRAFT)' > $@
 
 %.dtxt: %.tex %-text-directives.tex
 ifeq ($(HEVEA),true)
@@ -47,9 +51,6 @@ ifeq ($(HEVEA),true)
        ocaml docs.ml < $< > $@
 endif
 
-%-directives.tex:
-       printf '$(TEXDIRECTIVES)\\textversionfalse\\draft$(DRAFT)' > $@
-
 # (pdf)latex must be run multiple times to generate toc and correct references
 
 %.aux %.htoc: %.tex %-directives.tex
```
but decided not to include it because I don't think it really improves anything, yet may be more fragile.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -29,15 +29,18 @@ real:
 .PHONY: once
 once: $(STRINGS) docs.ml prefs.tmp prefsdocs.tmp
 	echo '\def\unisonversion{$(VERSION)}' > unisonversion.tex
+# Text versions: when using different directives then _must_ use different
+# jobname (or, output name for hevea) so that files for non-text versions are
+# not overwritten. Here, suffix -text is used.
 ifeq ($(HEVEA),true)
 	printf '$(TEXDIRECTIVES)\\textversiontrue\\draft$(DRAFT)' \
-           > texdirectives.tex
-	hevea unison-manual.tex
-	(TERM=vt100; export TERM; lynx -display_charset=utf8 -dump unison-manual.html > unison-manual.dtxt)
+           > unison-manual-text-directives.tex
+	hevea -o unison-manual-text.html unison-manual.tex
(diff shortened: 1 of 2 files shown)


gdt:
Well, sometimes people actually look at dvi using xdvi.  I did that a lot in the 90s, before pdf existed, because xdvi worked better than a ps viewer, prepatory to converting to ps (dvips) and printing.   But I don't think it is super important.    It would be nice to have a target for it, even it it's marked non-parallel, and isn't part of "all".  But I am not sure it's worth the work.

PR #939, opened 2023-07-01
Add a more sensible install target in Makefile

tleedjarv:
Provide a basic install target. Not intended to work on Windows but should work on most POSIX-y systems that have a BSD- or GNU-compatible `install`.

@gdt could you review if this works for packagers?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,16 +37,49 @@ clean:
 	$(MAKE) -C man clean
 	$(MAKE) -C src clean
 
+INSTALL ?= install
+INSTALL_PROGRAM ?= $(INSTALL)
+INSTALL_DATA ?= $(INSTALL) -m 644
+
+prefix = /usr/local
+PREFIX ?= $(prefix)
+exec_prefix = $(PREFIX)
+EXEC_PREFIX ?= $(exec_prefix)
+bindir = $(EXEC_PREFIX)/bin
+BINDIR ?= $(bindir)
+datarootdir = $(PREFIX)/share
+DATAROOTDIR ?= $(datarootdir)
+mandir = $(DATAROOTDIR)/man
+MANDIR ?= $(mandir)
+man1dir = $(MANDIR)/man1
+MAN1DIR ?= $(man1dir)
+manext = .1
+MANEXT ?= $(manext)
+
 .PHONY: install
-install:
-	@printf "\n\n=========================================\n\
-To install, copy the files src/unison, src/unison-gui (optional),\n\
-src/unison-fsmonitor (optional) and src/fsmonitor.py (optional,\n\
-if unison-fsmonitor does not exist) to a freely chosen location.\n\n\
-Manual page is at man/unison.1 and user manual is at\n\
-doc/unison-manual.pdf, doc/unison-manual.html and doc/unison-manual.txt\n\
-=========================================\n\n\n"
-	@exit 1
+install: all
+	$(INSTALL) -d "$(DESTDIR)$(BINDIR)"
+ifneq ($(wildcard src/unison),)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -37,16 +37,42 @@ clean:
...
 
+INSTALL=install
+INSTALL_PROGRAM=$(INSTALL)
+INSTALL_DATA=$(INSTALL) -m 644
+
+prefix = /usr/local
+exec_prefix = $(prefix)
+bindir = $(exec_prefix)/bin
+datarootdir = $(prefix)/share
+mandir = $(datarootdir)/man
```
gdt, line 48: Would be great to have these be env-able too, as for pkgsrc, mandir is prefix/man, not prefix/share/man.  Probably on other places too it varies.

Makefile
```diff
@@ -37,16 +37,42 @@ clean:
 	$(MAKE) -C man clean
 	$(MAKE) -C src clean
 
+INSTALL=install
+INSTALL_PROGRAM=$(INSTALL)
```
gdt, line 41: maybe a  755?   maybe that's defined by default, and can be skipped, but if that's what we mean, it's a nice declaration of intent that makes this easier to understand.

Makefile
```diff
@@ -37,16 +37,42 @@ clean:
...
-src/unison-fsmonitor (optional) and src/fsmonitor.py (optional,\n\
-if unison-fsmonitor does not exist) to a freely chosen location.\n\n\
-Manual page is at man/unison.1 and user manual is at\n\
-doc/unison-manual.pdf, doc/unison-manual.html and doc/unison-manual.txt\n\
-=========================================\n\n\n"
-	@exit 1
+install: all
+	$(INSTALL) -d "$(DESTDIR)$(bindir)"
+ifneq ($(wildcard src/unison),)
+	$(INSTALL_PROGRAM) src/unison "$(DESTDIR)$(bindir)/unison"
```
gdt, line 56: bindir is not mkdir -d'd.  I know that sounds funny, but I think it's normal to install a program into its own prefix for testing.

gdt (commented):
Unsaid in micro comments is that this seems otherwise fine.

Makefile
```diff
@@ -37,16 +37,42 @@ clean:
 	$(MAKE) -C man clean
 	$(MAKE) -C src clean
 
+INSTALL=install
+INSTALL_PROGRAM=$(INSTALL)
+INSTALL_DATA=$(INSTALL) -m 644
+
+prefix = /usr/local
```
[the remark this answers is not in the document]
tleedjarv: I think this already works as you'd expect. At least that's my understanding with GNU Make (which is what we currently require). But I'd rather get input from someone who knows better.

I pretty much followed the examples from GNU Make's manual. Not sure if they could be outdated in some cases.

Makefile
tleedjarv: Agree, and that was my intention. As far as I can tell, these are env-able (can easily be converted to uppercase, too, if needed).

Makefile
tleedjarv: I would go after examples from some common projects. Any suggestions as to what those might be?

Makefile
tleedjarv: `$(INSTALL) -d` on line 54 is supposed to do that here. (Also tested precisely as you describe.)

tleedjarv:
Rebased and updated. Now all variables should be env-able. In fact, variables from env will override GNU-style lowercase variables.

jhjourdan:
Is there anything blocking this PR? Should we merge?

It seems like no reviewer raised any issue, and, as far as I can tell, the target is doing what is needed ;)

gdt:
I tested on NetBSD and was able to control PREFIX and MANDIR from the environment, so I'm now ok with this.

PR #943, opened 2023-07-05
Fix a bug with fastcheck and archive properties

tleedjarv:
More details in commit messages. Long story short, the bug has been there since update predicates started being stored in archive properties but so far it has not been dangerous. Most users would likely never have noticed. With the ctime-based fastcheck, there is now an actual risk of missing some file properties. It still requires specific conditions that most users will never encounter, but it's now a real risk that needs to be fixed.


gdt:
I do not expect real review to happen, so given that it passes regression I'm going to just hit merge.

If you are someone reading along who would like a chance to review then 1) please feel free to do so in arrears and speak up and 2) after you do 1, I'll likely slow down.  But I'm not inclined to wait for theoetical people that don't exist, for now.

Issue #959, opened 2023-09-07
Part of the column headers of the comparison results are unreadable


Korb:
It is not clear what is displayed in the title of the two columns with the values "chgd dir", "deleted", "new file" and "props":

![Screenshot_2023-09-07_12-28-30 cropped](https://github.com/bcpierce00/unison/assets/3509797/d0d5ff7b-c66d-48c6-9aec-e2bbbd6f95b4)

Unison File Synchronizer 2.53.3 / unison-gui.exe
Microsoft Windows 10.0.19045.3393 ("use Unicode UTF-8 for worldwide language support" setting in "on")

tleedjarv:
It seems there are two separate issues here.

First, it appears that UTF-8 processing is broken. If that's the case then a fix will be provided shortly.

Second, the titles of the two columns are just too short to be useful. This is a known issue: #506 is one such example, there are probably more.

gdt:
There are multiple things and issues need to be granular.   @Korb could you test #962?   That will almost certainly get merged in a week.   Then, it would probably be good to open issues that say "column says X and it should say Y", assuming that Y is not much longer.  We can't just expand everything.  If it is more complicated than a specific sugestion that is fairly clearly reasonable, please instead open  a discussion on unison-users@.

gdt:
I'm going to declare that this issue is about the UTF-8 mishandling, so it will be closed when that is merged.  That doesn't mean other things aren't wrong, just that we can't handle multiple separable things in one issue.

Korb:
> There are multiple things and issues need to be granular. @Korb could you test #962?

I don't know how exactly to do this, sorry.

gdt:
https://github.com/bcpierce00/unison/actions/runs/6113238365?pr=962 and scroll down to artifacts.   These are builds with the proposed fix.

Korb:
Probably the problem can be considered solved:

![Screenshot_2023-09-18_11-14-53](https://github.com/bcpierce00/unison/assets/3509797/098122d1-788a-4deb-871e-6e94f1fed1b4)

Yes, even if the names of the synchronized folders are the same, it is not obvious which column belongs to which location. But that's another problem.

tleedjarv:
Pushed a small update to the PR, correcting an issue with underscores (a bug that seems to have been there for a long time). If you'd like, you can test the build at https://github.com/bcpierce00/unison/actions/runs/6220749068?pr=962

gdt:
[...]

PR #962, opened 2023-09-07
Use UTF-8 encoding-safe string truncation in UI

tleedjarv:
The names of roots can be very long and are truncated in UI in the list of updates. Truncating by number of bytes does not play well with UTF-8 input. This patch makes the truncation safe on UTF-8 encoded strings while keeping the length of truncated strings roughly as before (and the result should also _look_ correct for Western languages, maybe also for other scripts).

There is one tiny caveat. Since this patch applies to both TUI and GUI (also macOS native GUI), this may have a side-effect of scrambling any non-ISO-8859-1 (or Windows-1252 in GUI) and non-UTF-8 encodings, if such would ever be used. There is no change for the GTK GUI, as it has always expected either ISO-8859-1/Windows-1252 or UTF-8. I don't know if this will actually be a problem because I don't know if Unison has ever worked properly with these encodings. _This caveat only applies to the root names, not to file/dir names. This is a display-only issue, it will not break any syncs!_

This is a partial fix for #959.

### DIFF
--- a/src/.depend
+++ b/src/.depend
@@ -1237,11 +1237,13 @@ ubase/umarshal.cmx : \
     ubase/umarshal.cmi
 ubase/umarshal.cmi :
 ubase/util.cmo : \
+    unicode.cmi \
     system.cmi \
     ubase/safelist.cmi \
     ubase/projectInfo.cmo \
     ubase/util.cmi
 ubase/util.cmx : \
+    unicode.cmx \
     system.cmx \
     ubase/safelist.cmx \
     ubase/projectInfo.cmx \
--- a/src/ubase/util.ml
+++ b/src/ubase/util.ml
@@ -362,11 +362,47 @@ let option2string (prt: 'a -> string) = function
 (*                    String utility functions                               *)
 (*****************************************************************************)
 
-let truncateString string length =
-  let actualLength = String.length string in
-  if actualLength <= length then string^(String.make (length - actualLength) ' ')
-  else if actualLength < 3 then string
-  else (String.sub string 0 (length - 3))^ "..."
+let truncateString s count =
+  (* Truncate a string by counting code points instead of bytes. *)
+  let rec subValidUTF8 ?(extra = 0) s pos len =
+    (* Like [String.sub] but tries to keep the substring a valid UTF-8
+       string (it may not be meaningful in any way but the encoding is not
+       broken). Requires the input string to be valid UTF-8 to work
+       properly.
+       If the initial substring (like a simple [String.sub]) is not valid
+       UTF-8 then it tries to blindly extend (never reduce) the substring
+       until it becomes valid UTF-8. This is a very simple implementation
+       that works without knowing anything about the UTF-8 encoding. *)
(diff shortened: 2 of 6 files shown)


gdt:
I have called for testing on unison-users@ and will merge in 7 days if no adverse comments.

igrep:
I've tested unison-7f147080.ocaml-4.14.0+mingw64c.windows-2022\unison.exe on my Windows 11 22H2 (both when `chcp 932` and when `chcp 65001`). It looks like working without any problem after creating/deleting a `あああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああ.txt`.

tleedjarv:
Thank you for testing. Syncing functionality is not affected by this patch. Only the truncated display of root names as "left" and "right" headings is affected.
Realistically, the potential encoding issue could only manifest in the text UI because the same UTF-8 conversions have always been done in the GUIs; this patch shouldn't change anything there.

igrep:
Ah, sorry. I tried again, then found it looked good:

```
> .\unison C:\t\ああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああああ C:\t\いいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいいい
Unison 2.53.3 (ocaml 4.14.0): Contacting server...
Looking for changes
Reconciling changes
The root of one of the replicas has been completely emptied.
Unison may delete everything in the other replica.  (Set the
'confirmbigdel' preference to false to disable this check.)

Do you really want to proceed? []
Unrecognized command ' ': try again  [type '?' or F1 for help]
Do you really want to proceed? [] y
あああああああああ...   いいいいいいいいい...
         <---- deleted    新規 テキスト ドキュメント.txt  [f] ,
```

tleedjarv:
Thank you for testing! Is this with UTF-8 or some other encoding?

igrep:
Sorry again! I tried when `chcp 932`, and confirmed the result is same with my last comment.

tleedjarv:
> Sorry again! I tried when `chcp 932`, and confirmed the result is same with my last comment (UTF-8: chcp 65001).

I suspect that this will always work properly in Windows, with NTFS at least. It _may_ break (meaning display will be garbled) with FAT when the dirnames are knowingly created in codepage other than UTF-8. But I don't know how that works in Windows.

I think the only real risk case is some Unix/Linux when dirnames on disk are encoding-less blobs (as seen by the OS and the filesystem) and those blobs are neither valid UTF-8 nor Latin 1 (ASCII is of course a subset of both).
(As a side note: When it comes to multi-byte encodings, the display would have always been garbled, no regression there.)

Thank you for testing! At least we know nothing's immediately broken, and even then, any eventual breakage is going to be limited and cosmetic only.

PR #964, opened 2023-09-12
Makefile: Use ocamlfind (if installed) to detect lablgtk3

tleedjarv:
Keep the old libdir-based detection as well because ocamlfind is _not required_ for the build itself.

Diff best viewed with indentation changes ignored and moved blocks highlighted. (The GH diff display of this patch is just useless.)


tleedjarv:
Rebased.

PR #970, opened 2023-10-03
Ignore `force = newer/older` if mtimes are equal

tleedjarv:
`force = newer/older` is defined (and documented as such) only for changes where mtimes are different between the replicas. If mtimes are equal then root 2 is always forced over root 1.

This produces unexpected results for changes that don't change mtime (basically all metadata changes). Since `force = newer/older` is undefined for such changes (and, importantly, is very difficult to define), this patch just ignores the preference ~~for non-conflicting changes~~ if mtimes are equal. ~~For conflicting changes, the current default of forcing root 2 is kept unchanged.~~

There is further discussion and background in the mailing list.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,6 +20,12 @@ Released 2023-xx-yy
     for changed build goals, input variables and build artifact names.
     See INSTALL.md for details.
   * Various fixes in Windows and Cygwin builds.
+  * Preferences "force", "prefer" and related "partial" preferences now
+    work slightly differently with values "newer" and "older". Previously,
+    if mtimes in both replicas were equal then always the second root
+    propagated to the first root (possibly reverting user changes). It
+    is now made explicit that "newer" and "older" only work when mtimes
+    are different.
   * Cleanups in documentation.
   * Bugfixes, minor improvements, cleanups.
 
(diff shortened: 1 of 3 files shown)


gdt:
I object to forcing root 2 if there are conflicting changes.   The user has asked for newer to be forced, but there is no newer, and thus no basis to force at all.  What they are trying to do, and what they actually want, is hard to guess, so I think we should just not.  Yes, I realize this leads to interactive, but I think that's better than doing something unexpected that is basically bug-compatible with before.  Preferring root2 over root1 in  a tie I find bizarre in the first place and it smells like fallthrough instead of intended.

tleedjarv:
I have no issue with your objection and it's only a minor change to the patch but I can't help but be reminded of https://xkcd.com/1172/ (of course, the PR in its current form does the same, so...).

gdt:
The xkcd is amusing.  However, I feel that force=newer is a kludge to start with, and doesn't really belong in a sync tool.  If you want rsync instead of unison, it's much simpler.

gdt:
Thanks.  Will hold off until 16Z on 5 October for discussion and then merge as soon as I realize it's after that :-)

PR #971, opened 2023-10-03
Fix a ctime fastcheck miss, part 2

tleedjarv:
This is the same as #941 but this time for directories (which were missed by the original PR).


gdt:
Will merge as soon as I merge #970.

Issue #996, opened 2024-02-14
Excessive failures logged when using `-repeat`


acsysch:
When syncing continuously using a cmd such as "Unison profile -repeat watch" and a fatal error occurs, unison 2.53.3 issues "restarting in 10 seconds" message and loops back to the fatal error. According to the documentation, one should then issue CTRL+C to exit, while previous versions used to exit in case of an error returning an exit code.
**This new behaviour isn't usable when using this cmd in a background batch** (on Windows in my case).
Could you either restore the previous auto-exit on error OR provide an additional switch on the cmdline in order to exit on error? 

![image](https://github.com/bcpierce00/unison/assets/97226581/797cc32d-8481-429d-8432-f7fc8a4bd923)

tleedjarv:
This is by design. Repeat mode is not intended for use in scripts/batch. What would that even mean?

What is your intention with running in repeat mode (which by definition is non-exiting) in a batch script?

acsysch:
Well, the intention of such batch is to keep in sync a couple of roots in the background. Of course, the sync should stop upon encountering an error.

In the case we use it for, the 2 roots are located on 2 different servers. The aforementioned batch monitors the connection and, upon sync-error, returns to a waiting loop for the remote server to be back online.

 

I don’t see any contradiction in using repeat-mode (i.e. a loop) in a batch as long as I can exit that mode on error. And that worked with version 2.53.0

gdt:
This seems better suited for the mailinglist.

tleedjarv:
You are welcome to clarify the request in the mailing list but I really don't think you need one.

> The aforementioned batch monitors the connection and, upon sync-error, returns to a waiting loop for the remote server to be back online.

It sounds to me that you are using the script to achieve something that is exactly what the repeat mode will do for you. It will continue automatically when the remote end is back online.

> I don’t see any contradiction in using repeat-mode (i.e. a loop) in a batch as long as I can exit that mode on error.

Repeat mode is a never-ending process. You may run it under some process monitor (or init system) that monitors if the process itself is alive and restarts it when necessary.

What you're asking for is exactly like running a Dropbox client (or any other cloud drive client) in a script when in fact the client is start-once never-exit process that will be resilient to connection drops. Or any other permanently running client for that matter. You don't need to monitor the connection yourself, you may only want to monitor if the process has died and restart it.

> Its manual even stated in chapter 6.10 "The exit status could be useful when Unison is invoked from a script"!

[...]

PR #1013, opened 2024-03-27
Suppress logging of repeating errors when retrying automatically

tleedjarv:
This patch improves the logging in case of automatic retry in repeat mode, which is one of the issues pointed out in #996.

Repeating messages are suppressed only in the log, not on stdout when run interactively.

Easiest reviewed commit by commit.


gdt:
retry mode is complicated and has a lot of issues.  For now, I have decided that I don't want to suppress errors, and that instead people who are concerned about too many errors in the log should instead reduce the rate of errors happening.  This may well take a great deal of stepping back and reconsidering many things, but that's ok with me.

tleedjarv:
I agree, this PR no longer makes sense.

PR #1015, opened 2024-04-01
Fix incorrect retry when SIGINT reaches ssh

tleedjarv:
This PR is opened on top of #1013. _Only the last commit_ is unique to this PR and needs to be reviewed here.


tleedjarv:
Rebased on top of #1024

PR #1028, opened 2024-04-27
Improve INSTALL.md

gdt:
This fixes up some previous changes, and makes some new improvements.  I'd like to merge it if it makes things better and does no harm.   I'm open to discussing other things to change, but if it's reasonable to merge this and discuss those separately, I always lean to keeping separable things separate.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -29,14 +29,13 @@ not in the unison issue tracker.
 
 #### Build prerequisites
 
+- A C99 compiler (e.g. gcc, clang)
 - A recent version of OCaml compiler (version 4.08 at minimum)
-  together with a C99 compiler (such as gcc, clang) -- see
-  https://ocaml.org/.  (Note that ocaml upstream says that 5.x is
+  -- see https://ocaml.org/.  (Note that ocaml upstream says that 5.x is
   experimental.  The standard approach is the most recent 4.x ocaml
   release.)
-- A non-ancient C compiler
 - GNU make
-- Basic POSIX tools: sh, sed (optional, for manuals), perhaps more
+- Basic POSIX tools: install, rm, sh
 
 ##### Optional, for the GUI only
 
@@ -49,6 +48,7 @@ not in the unison issue tracker.
 
 ##### Optional, for building the user manual
 
+- sed
 - LaTeX
 - pdf2ps (optional, for PS output; included with Ghostscript, for example)
 - HEVEA (https://hevea.inria.fr/) (optional, for HTML and text formats)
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -115,7 +113,9 @@ To build the documentation, first build Unison without cross-compilation.
 
 #### Build prerequisites
 
-- Xcode Command Line Tools (optional, for the native GUI)
+- Xcode Command Line Tools
+- ocaml (see mostly-POSIX section)
+- optional prereqs from gui and doc sections above
```
tleedjarv, line 118: The commit message states that CLT is needed anyway, but is it fully correct? I don't use macOS myself, so don't know how it works, but even if most installations have CLT installed, couldn't you have gcc installed without any CLT? Maybe it's possible but just exceptionally rare... Either way, I'll leave the decision of CLT being optional or not to you but I'd also consolidate the bullets more (see suggestion below).

```suggestion
- Xcode Command Line Tools (optional, for the native GUI)
- See mostly-POSIX section (common prereqs and text UI, GTK GUI, docs)
```

tleedjarv (approved):
Looks good to me. Just one nit about the macOS section.

INSTALL.md
gdt: The basic issue is that on macs, things traditionally present in Unix have been removed and put in CLT.  It's basically an apple package of compiler and various other things.    So you are right that it isn't strictly needed.   However, the standard approach on a mac is to install either CLT or XCode to get the things you'd expect to be present for buidling programs.   More recently, Apple has removed system headers (/usr/include) and building for the current system is sort of cross, but in a way you don't notice because it's the default.   The headers for the native GUI are almost certainly in this category.  But so are the regular posix headers; a mac running the latest macos does not have /usr/include at all, so even if you had a "C compiler", you still need "POSIX system headers".   I see this as mac oddness, and don't really want to cross into pointing out that you need POSIX system headers on other systems, since it's bizarre to have a C99 compiler and lack them.

So I think I'll rewrite to say that a mac has basically the same prereqs as POSIX, but that as macs lack a C compiler and POSIX header files out of the box, the standard approach is to install CLT or Xcode.  And that the mac GUI requires that one have the appropriate headers/libs, which CLT/Xcode provide.

I'm not really aware of a practice of using macs to build things by installing other compilers and not installing CLT.   Basically installing CLT is what you have to do to make it a regular computer instead of an internet browsing appliance.   For example, homebrew (a mac-specific package manager) says you need CLT or XCode.

Thus, in terms of being helpful for users, telling them they need CLT isn't wrong, but I could reword to say that CLT or XCode is the typical path to satisfying the prereqs.

gdt: reworded and pushed.   Having written that, I feel that I don't want to go down the path of "but you could also install gcc and get headers from someplace else".   People who are going to succeed at that don't need or expect help from Unison's INSTALL.md :-)   Let me know what you think now.

tleedjarv: This is very informative. I'm fine with the change!

Issue #1034, opened 2024-05-07
Choose approach to gdk-pixbuf upstream being unstable


gdt:
They intentionally broke the API in a micro release: https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/merge_requests/169

This ticket is to choose an approach from:
  1) consider this an upstream bug that packaging systems should remediate
  1) change formats so we no longer care about XPM, picking something we hope isn't later deleted
  1) stop using gdk-pixbuf

For now we are on option 1.

gdt:
We have picked 4. implement xpm ourself, and don't ask gdk-pixbuf to deal with it.

That leaves 1 as still a valid opinion, but no longer relevant to unison :-)

Issue #1033, opened 2024-05-05
 Uncaught exception Gpointer.Null


surferbk:
starting unison get the error:
Fatal error

Uncaught exception Gpointer.Null
Raised by primitive operation at Uigtk3.Private.createToplevelWindow.buildPixmap in file "/home/abuild/rpmbuild/BUILD/unison-2.53.4/src/uigtk3.ml" (inlined), line 3283, characters 4-29
Called from Uigtk3.Private.createToplevelWindow in file "/home/abuild/rpmbuild/BUILD/unison-2.53.4/src/uigtk3.ml", line 3287, characters 17-44
Called from Uigtk3.Private.start in file "/home/abuild/rpmbuild/BUILD/unison-2.53.4/src/uigtk3.ml", line 4474, characters 19-42

I did not understand and I could not find the path /home/abuild/... for more information.

If I try to start unison in the Terminal I get the error:
unison:8590): GdkPixbuf-WARNING **: 15:54:17.569: Error loading XPM image loader: Bildlader-Modul konnte nicht geladen werden: /usr/lib64/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-xpm.so: /usr/lib64/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-xpm.so: Kann die Shared-Object-Datei nicht öffnen: Datei oder Verzeichnis nicht gefunden

In the path /usr/..../loaders/ there is no file libpixbufloader-xpm.so

Where can I get this file?

Could you help me please?

My system:
Betriebssystem: openSUSE Tumbleweed 20240503
KDE-Plasma-Version: 6.0.4
KDE-Frameworks-Version: 6.1.0
Qt-Version: 6.7.0
Kernel-Version: 6.8.8-1-default (64-bit)
Grafik-Plattform: X11
Prozessoren: 4 × Intel® Core™ i3-8100T CPU @ 3.10GHz
Speicher: 15,5 GiB Arbeitsspeicher
Grafikprozessor: Mesa Intel® UHD Graphics 630
Hersteller: Micro-Star International Co., Ltd.
Produktname: MS-7B53
Systemversion: 1.0

gdt:
This looks like a request for help debugging what seems like a not-really-right packaging build of unison, rather than a bug report.
Please see https://github.com/bcpierce00/unison/wiki/Reporting-Bugs-and-Feature-Requests

[...]

PR #1035, opened 2024-05-07
Stop using deprecated `GdkPixbuf.from_xpm_data`

tleedjarv:
Re #1033 #1034

Since the XPM format is so simple, the function to parse (subset of) XPM used for the icons is trivial. This approach does not require external resources (like image files or loader plugins) nor embedding binary-format icons in the source.

Working but not heavily tested.

### DIFF
--- a/src/dune
+++ b/src/dune
@@ -1,7 +1,7 @@
 (library
  (name unison_lib)
  (wrapped false)
- (modules :standard \ linktext linkgtk3 uigtk3 uimacbridge test)
+ (modules :standard \ linktext linkgtk3 uigtk3 pixmaps uimacbridge test)
  (modules_without_implementation ui)
  (flags :standard
         -w -3-6-9-10-26-27-32-34-35-38-39-50-52
@@ -27,5 +27,5 @@
  (public_name unison-gui)
  (package unison-gui) ; Dummy: we don't use packages
  (flags :standard -w -3-6-9-27-32-52)
- (modules linkgtk3 uigtk3)
+ (modules linkgtk3 uigtk3 pixmaps)
  (libraries threads unison_lib lablgtk3))
--- a/src/pixmaps.ml
+++ b/src/pixmaps.ml
@@ -253,6 +253,63 @@ let copyBAblack_asym = [|
 
 
 (***********************************************************************)
+(*                          XPM parse function                         *)
+(***********************************************************************)
+
+(* This function is not for universal XPM parsing. It is intended only
+   for parsing the icon definitions above in this file.
+   Do not use [GdkPixbuf.from_xpm_data] as it has been removed from
+   upstream gdk-pixbuf. *)
+let to_pixbuf dat =
+  let colormap = Array.make 256 [| 0; 0; 0; 0 |] in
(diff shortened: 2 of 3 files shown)


src/dune
```diff
@@ -1,7 +1,7 @@
 (library
  (name unison_lib)
  (wrapped false)
- (modules :standard \ linktext linkgtk3 uigtk3 uimacbridge test)
+ (modules :standard \ linktext linkgtk3 uigtk3 pixmaps uimacbridge test)
```
gdt, line 4: commit message says "fix", but I think this is "add module that the previous commit started using".  Always watching out for ways future self might misread....

gdt:
I would really like to see someone review and test.    We just had a release, and I am going to try not to merge anything for a bit until the release and the gdk-pixbuf-is-unstable kerfluffle settle out a bit.

src/dune
tleedjarv: I considered it a fix as the list of modules was incorrect even before the pixbuf change. It's just that it didn't produce an actual error before the pixbuf change, but it wasn't correct. The pixbuf change made the Pixmaps module depend on lablgtk.

ithod:
On openSUSE Tumbleweed I get the following error:
`Uncaught exception (please report a bug): Gpointer.Null
Raised by primitive operation at Uigtk3.Private.createToplevelWindow.buildPixmap in file "/home/abuild/rpmbuild/BUILD/unison-2.53.5/src/uigtk3.ml" (inlined), line 3296, characters 4-29
Called from Uigtk3.Private.createToplevelWindow in file "/home/abuild/rpmbuild/BUILD/unison-2.53.5/src/uigtk3.ml", line 3300, characters 17-44
Called from Uigtk3.Private.start in file "/home/abuild/rpmbuild/BUILD/unison-2.53.5/src/uigtk3.ml", line 4489, characters 19-42`

https://build.opensuse.org/project/show/Archiving:unison

gdt:
Are you testing the branch in this PR, or an artifact built by CI from this branch?  It is known that a release build of unison with a buggy gdk-pixbuf (as contained in openSUSE) does not work.

tleedjarv:
@ithod I'm pretty certain that https://build.opensuse.org/project/show/Archiving:unison does not contain the patch from this PR. This makes your comment a duplicate of #1033.

ithod:
> I would really like to see someone review and test. We just had a release, and I am going to try not to merge anything for a bit until the release and the gdk-pixbuf-is-unstable kerfluffle settle out a bit.

Sorry for my unclear message, indeed the error occurs if the patch is not applied. I wanted to make clear that this patch solves the error given above. When I apply the patch (the 1st, not sure what `src/dune` is), it works for me
https://build.opensuse.org/package/show/home:ithod:updated/unison

tleedjarv:
Thank you @ithod.

> the patch (the 1st, not sure what `src/dune` is)

The 1st is the real patch, yes. The other one is only needed when building with `dune` (and not `make`).

gdt:
We've had one test report, and enough time that anybody that wanted to test has had a chance.

PR #1042, opened 2024-06-02
GHA: CI housekeeping

tleedjarv:


gdt:
I'm ok with dropping testing with 2.48.   However, I think it merits saying something on users@, which I can do, which is "While we don't intend to break compatibility with 2.48, we are no longer testing against it.  Anyone using 2.48 is reminded that it is very very past time to upgrade."

Are we declaring 5.1 irrelevant/broken?  We have been trying to verify on all. are we going to document that unison doesn't work on 5.1?  Something else?

tleedjarv:
> I'm ok with dropping testing with 2.48. However, I think it merits saying something on users@, which I can do, which is "While we don't intend to break compatibility with 2.48, we are no longer testing against it. Anyone using 2.48 is reminded that it is very very past time to upgrade."

I don't think it is necessary but you are free to do so, of course.

> Are we declaring 5.1 irrelevant/broken? We have been trying to verify on all. are we going to document that unison doesn't work on 5.1? Something else?

No, not directly. I don't think there's any point in verifying against _all_ versions. Despite the horror stories, there's normally no random breakage from version to version. That's why I removed most of the 4.x versions, too. With 5.x, it is still experimental ("somewhat experimental" is the phrase used by the upstream) and will probably be for another year (that is the current upstream expectation). Testing against the latest 5.x only seems appropriate and sufficient. This does not mean that earlier 5.x stops working but if a future 5.x makes breaking changes then I see no reason to support any earlier 5.x versions. I'd maintain support starting from the first stable 5.x version (expected in about a year).

Issue #1038, opened 2024-05-18
Add days to ETA display (Nd hh:mm:ss instead of hh:mm:ss) to improve >>1h readability


mvglasow:
[EDIT: dropped because master is hh:mm:ss.  ~~When displaying progress, ETA is currently given in mm:ss. While this is no issue for times up to 60 minutes, it gets hard to read when ETA is several hours~~]

Suggestion: change ETA format to something like `1d 18:56:19`. Days can be omitted if ETA is less than one day. Hours can also be omitted if ETA is less than an hour – or can be left in to make it clear that ETA is shown as mm:ss, not hh:mm.

gdt:
Calling this an enhancement request; seems reasonable.    I would prefer to leave mm:ss as is when sub-hour; the time will be counting down and I don't think it's actually confusing.

mvglasow:
This was observed in unison.log contents, in conjunction with the text GUI. (Both machines involved run Linux.)

Use cases may differ – occasionally I do in fact sync several GB of data over a low-bandwidth connection, which does take several days. So for some cases it is definitely helpful, and shouldn’t be difficult to implement if one knows where to look in the code and is familiar with OCaml. Essentially, it’s just `if hh >= 24 then {dd == hh mod 24; hh == hh div 24} else dd == 0`.

Btw, if different UIs display time differently, this is a sign of code duplication – it’s probably better to move time formatting code into a library function which can be called from wherever it’s needed.

tleedjarv:
> This was observed in unison.log contents, in conjunction with the text GUI. (Both machines involved run Linux.)

Interesting... could you post an example?

> So for some cases it is definitely helpful, and shouldn’t be difficult to implement if one knows where to look in the code and is familiar with OCaml. Essentially, it’s just `if hh >= 24 then {dd == hh mod 24; hh == hh div 24} else dd == 0`.

The code for the latest release is located here: https://github.com/bcpierce00/unison/blob/v2.53.5/src/uicommon.ml#L496-L506

mvglasow:
A few lines from unison.log:

```
 22%  1507:37 ETA
 22%  1511:17 ETA
 22%  1514:05 ETA
 22%  1516:53 ETA
 22%  1519:15 ETA
```

tleedjarv:
I see. This must be some older version then because this should not happen with the current code. Could you try with the latest release?

mvglasow:
I was about to write that this output doesn’t look like it came from the code you linked.

[...]

PR #1043, opened 2024-06-04
Show days in ETA if needed

mvglasow:
As discussed in #1038. This PR adds the following functionality:

If ETA >= 86220 s (which rounds up to 86400 s, or 24 h, for display purposes), it is displayed as:

    0%  0/1  (40.3 MiB of 263.59 GiB)   357 KiB/s    2d 20:25:00 ETA

ETAs below that threshold will get displayed as before.

Since this is my first time messing with any OCaml code, you might want to look at the syntax. Lines 496–501 are truly new, the others just had indentation added. In particular, I am not sure if the parentheses around the tuple members in 497/498 are correct, or if they’re needed at all.

I have run a brief test with ETAs both below and above the threshold, and it builds and works as intended.

### DIFF
--- a/src/uicommon.ml
+++ b/src/uicommon.ml
@@ -493,17 +493,22 @@ let calcETA rem rate =
   let t = truncate (rem /. rate +. 0.5) in
   (* Estimating the remaining time is not accurate. Reduce the display
      precision (and reduce more when longer time remaining). *)
-  let h, (m, sec) =
-    if t >= 3420 then
-      let u = t + 180 in u / 3600, (((u mod 3600) / 300) * 5, 0)
-    else
-      0,
-      if t >= 2640 then ((t + 180) / 300) * 5, 0
-      else if t >= 1800 then ((t + 119) / 120) * 2, 0
-      else if t >= 120 then let u = t + 15 in u / 60, ((u mod 60) / 30) * 30
-      else t / 60, t mod 60
-  in
-  Printf.sprintf "%02d:%02d:%02d" h m sec
+  if t >= 86220 then
+    let u = t + 180 in
+    Printf.sprintf "%dd %02d:%02d:00" (u / 86400) ((u mod 86400) / 3600)
+      (((u mod 3600) / 300) * 5)
+  else
+    let h, (m, sec) =
+      if t >= 3420 then
+        let u = t + 180 in u / 3600, (((u mod 3600) / 300) * 5, 0)
+      else
+        0,
+        if t >= 2640 then ((t + 180) / 300) * 5, 0
+        else if t >= 1800 then ((t + 119) / 120) * 2, 0
+        else if t >= 120 then let u = t + 15 in u / 60, ((u mod 60) / 30) * 30
+        else t / 60, t mod 60
+    in
+    Printf.sprintf "%02d:%02d:%02d" h m sec
 
 let movAvg curr prev ?(c = 1.) deltaTime avgPeriod =
   if Float.is_nan prev then curr else


src/uicommon.ml
```diff
@@ -493,17 +493,23 @@ let calcETA rem rate =
...
-      else if t >= 1800 then ((t + 119) / 120) * 2, 0
-      else if t >= 120 then let u = t + 15 in u / 60, ((u mod 60) / 30) * 30
-      else t / 60, t mod 60
-  in
-  Printf.sprintf "%02d:%02d:%02d" h m sec
+  if t >= 86220 then
+    let d, h, (m, sec) =
+      let u = t + 180 in u / 86400, (u mod 86400) / 3600, (((u mod 3600) / 300) * 5, 0)
+    in
+    Printf.sprintf "%dd %02d:%02d:%02d" d h m sec
```
tleedjarv, line 500: Overall looks correct but I propose to simplify a little since now you don't need the tuple at all. At the same time, I'd consider reducing the precision even more.
Something like this (not tested):
```suggestion
    let u = t + 1799 in
    Printf.sprintf "%dd %02d:%02d:00" (u / 86400) ((u mod 86400) / 3600)
      (((u mod 3600) / 1800) * 30)
```
This makes the precision 30 min (currently rounded fully up, but can be adjusted by changing the constant `1799`). This could be anything, really.

Alternatively, you can still express all variables separately, but still don't need the tuple:
```ocaml
    let u = t + 1799 in
    let d = u / 86400 in
    let h = (u mod 86400) / 3600 in
    let m = ((u mod 3600) / 1800) * 30 in
    Printf.sprintf "%dd %02d:%02d:00" d h m
```
but I'm not sure it adds anything.

mvglasow: Fine with me, but then I would `let u = t + 1800` and make the threshold 84800 (86400 - 1800), to be consistent with the rest of the code.

tleedjarv: You could, but is it the best idea? Those thresholds are not directly related. You can still keep it at 86220 because that goes together with the < 24h threshold. If you make the threshold (86400 - 1800) then you cut the 24h range a bit short (meaning you'll jump from 23:30:00 to 1d 00:00:00). But that's ok too, all of this is completely arbitrary; there is nothing specifying that it should be 5 min precision up to 24h.

tleedjarv: > to be consistent with the rest of the code.

That's not making it consistent. Those thresholds are not related and you can see these differ in the code below. They are arbitrary, true, but they are not the same.

mvglasow: Indeed, I just went by the next lower threshold: for ETAs above roughly an hour, we add 180 s and make the threshold 3420 s.

Looking at the code again, however, I realize that for any `3420 > t >= 2640`, we effectively do the same as for `86400-ish > t >= 3420`, so we could simplify that part as well.

And below that, the thresholds are “true” rather than rounded ETA.

Reconsidering it all, I’d lean towards sticking with 5-minute precision up to `1d 00:00:00` or `24:00:00` (I guess the former format would be more consistent). So that would put us back to 86220...

mvglasow: In short, no further commits on top of your suggested improvements. The rest of the commit history in this PR cancels itself out.

src/uicommon.ml
```diff
@@ -494,10 +494,9 @@ let calcETA rem rate =
   (* Estimating the remaining time is not accurate. Reduce the display
      precision (and reduce more when longer time remaining). *)
   if t >= 86220 then
-    let d, h, (m, sec) =
-      let u = t + 180 in u / 86400, (u mod 86400) / 3600, (((u mod 3600) / 300) * 5, 0)
-    in
-    Printf.sprintf "%dd %02d:%02d:%02d" d h m sec
+    let u = t + 1799 in
+    Printf.sprintf "%dd %02d:%02d:00" (u / 86400) ((u mod 86400) / 3600)
```
gdt, line 498: I am really not a fan of displaying a :00 for minutes, if I follow that. 

I am ok with us deciding that these strings are only for humans.  (If they are for computers, they should just be in seconds, the SI base unit :-)    If so, we could adopt a policy that if there are days, we should only days and hours, and if hours, only hours and minutes, and so on.   But I think showing a 0 in a field when it is not, is not ok.

Overall I lean to simpler, in code but especially in presentation to humans.

tleedjarv:
@mvglasow maybe that's just how GitHub's suggestions work but you don't have to assign any co-authorship to me and can simply squash the commits.

tleedjarv (approved):
I think it looks ok.
Some primitive testing shows:
85919 is displayed as 23:50:00
86040 is displayed as 23:55:00
86218 is displayed as 23:55:00
86219 is displayed as 23:55:00
86220 is displayed as 1d 00:00:00
86221 is displayed as 1d 00:00:00
86399 is displayed as 1d 00:00:00
86400 is displayed as 1d 00:00:00
86401 is displayed as 1d 00:30:00
88199 is displayed as 1d 00:30:00
88200 is displayed as 1d 00:30:00
88201 is displayed as 1d 01:00:00
187800 is displayed as 2d 04:30:00

src/uicommon.ml
tleedjarv: (It's for the seconds in this case.)

I'm not sure I agree with you on this one. It sounds like form over function. How do you avoid ambiguity? You'd have to switch away from positional display then, right?

But this discussion should not extend the scope of this PR. I think this discussion instead goes well together with a discussion of the precision of time display (also separate from this PR).

gdt:
This PR is about changing large numbers of hours into days and hours.  I am fine with that.    I do not want to have a change to display rounded times in the same PR; that's a different discussion that we don't even have an issue about :-)

src/uicommon.ml
mvglasow: > showing a 0 in a field when it is not, is not ok.

For seconds this is what is already happening for any ETA above 2 minutes. Between 1–2 minutes ETA is rounded to the next 30 seconds, and below a minute the exact number of seconds is shown. The only change regarding rounding is that above 24 hours we further reduce precision from 5 to 30 minutes.

I don’t have a clear position on whether precision for ETAs above 24 h should be 5 m or 30 m. I’d suggest merging this PR for now – I’ll leave it up to you guys whether or not to make any changes to precision for ETA > 24 h before merging.

If there is a need to discuss what rounding we should apply when, or whether to display seconds when they are always zero, I would suggest doing so in a separate ticket. For seconds my only requirement is that I don’t want any ambiguity between `hh:mm` and `mm:ss`, even when I just have a single, static value such as a single log line.

gdt:
Agreed that varying between hh:mm and mm:ss is not ok.

I feel that the rounding we already have is overly aggressive.   Agreed that fixing that is separate.   But I'd like to not pile on and make it worse.   If we already have "round to 5 minutes" for times that are most of a day, flipping to 30 minutes at a day seems too much (4%).  (Yes, it might be that the existing breakpoint for 5m is also too much, but we're not changing that.)   If this just doesn't change rounding from how it is (leaving rounding at 5 for multiday), I'm fine with it.

mvglasow:
I’m fine with keeping precision at 5 minutes for ETAs beyond 24 h for now.

gdt:
Great then please adjust, rebase, and make the commit message give rationale for the change.  I don't mean anything super extensive, but it's a wee bit too brief as is

mvglasow:
Just committed. Github should allow a squash merge, which will squash all the commits from my branch into a single one – should be the easiest way. If not, give me a heads-up and I’ll see if I can rewrite my branch history.

gdt:
I do not expect github squash to do a good job with a merged commit message turning into the right message for posterity, and we do tend to have merge commits, so I think you can just do a rebase -i, p/s/s and force-push.

mvglasow:
OK. `git rebase` didn’t do anything as master has not changed, so I did `git rest master`, then re-committed the differences and force-pushed them to my branch. Commit message is the same as the subject of this PR, which should accurately describe what we’re doing (and, afaik, is also what Github uses as the commit message for a squash commit).

gdt:
The commit message is only really understandable once you have the context.  Let me think about how to say enough to orient people who have no idea.    It's allowed to have more lines after the 54-byte summary :-)   I'll wait a day or so for anybody else to comment.

(I meant `git rebase -i origin/master` which will give you a chance to specify how you want things in terms of squashing, fixups and reordering, vs rebase which will not.)   Thanks for pointing out about github and squash, but I basically think it's a bug we are github and don't want to rely on such behavior anyway, trying to be closer to pure git when reasonable.

gdt:
I'm not really 100% happy with the entire time display process, but on re-reading am ok with this being an improvement.   Sorry for it taking so long; I often have unison paged out pretty much entirely.

PR #1045, opened 2024-06-09
[Fix] MacOS build when xcode is not installed

EruEri:
On MacOS, there is two way to install tools you need to develop (make, git, cc, ...). EIther install `Xcode` or something called `Command Line Tools`. But the `Command Line Tools` version doesn't include all the necessary programs to build Apple applications.

Currently the src/Makefile.OCaml assumes that if the `OSARCH` is `Darwin` so it can build the macui but if the xcode version of xcodebuild is not installed, any usage of `xcodebuild` will trigger the following message and stop the build  process with an error.

```
$ xcodebuild
xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory '/Library/Developer/CommandLineTools' is a command line tools instance
```

### DIFF
--- a/src/Makefile.OCaml
+++ b/src/Makefile.OCaml
@@ -72,7 +72,14 @@ endif
 
 .PHONY: macuimaybe
 ifeq ($(OSARCH), Darwin)
-  macuimaybe: macui
+  # If XCode is not installed, xcodebuild is just a placeholder telling that XCode is not installed
+  # and any invocation of xcodebuild results in a non 0 exit code.
+  ifeq ($(shell xcodebuild -version > /dev/null; echo $$?), 0)
+    macuimaybe: macui
+  else
+    macuimaybe:
+	$(info Not building macOS native GUI because XCode is not installed.)
+  endif
 else
   macuimaybe:
 	$(info Not on macOS. macOS native GUI will not be built.)


gdt:
Thanks for contributing!

Could you rebase into fewer (perhaps 1 if there isn't  reason for more) commits and force-push?  That needs to happen anyway, so reviewers might as well look at what you think should go in, vs an intermediate state.  I prefer to wait until then to read the diff, as I don't really have enough cycles for unison maintenance.   (I do have a mac I can test on.)

When rebasing, please ensure that the rationale for how things *are* ends up in comments and the rationale for the *change* is fully in the commit message.   The union of the commit messages doesn't read like that now ;-)   The PR comments will be lost on merge.

Compiler tools on mac are a bit of a mess.  I guess you are talking about having CLT but not the rest.   Please make sure the explanation of the problem can be followed by someone who is a little hazy on exactly how oddly Apple does this.  My belief is that even people with CS degrees who have macs are generally a little unclear on CLT/XCode and how things relate, now and in earlier versions - it seems to keep changing.

Perhaps we should instead change the logic so that the mac gui is not built by default, but only when an explicit target is requested.  I'm not 100% comfortable with "mac gui is built on mac, except when it isn't".

We could also consider a better error message for the gui build, so it fails more cleanly and the invoker rapidly understands.

EruEri:
I only kept one commit. 

Yes I talk about the command line tools without xcode, sorry if it wasn't clear.

EruEri:
Do you think we should change the default and only build the mac gui if it's explicitly intended or provided a better error message when it fails ?

gdt:
I am not really sure.   But I don't really like silently not building it if there's isn't enough xcode support.   I know there is a tradition of enabling options if things are present in build systems, but as a packager that's more of a bug.

So far, we have had "build mac gui if a mac".   So that argues that the lesser change is to keep doing that, and turn a hard-to-read error into "Can't build Unison.app because XCode is not installed:"  or something more useful.

The big question is how normal it is to have a setup where you can build C programs but don't have XCode, and how people in that situation got there.    And also how this relates to following the instructions in INSTALL.txt.
That argues that that INSTALL.txt should say that CLT is enough to build unison/unison-gui and that you need XCode for Unison.app.  And then to unbundle macui from the default build.

EruEri:
But currently it causes the opam install of `unison` (not `unison-gui`) to fail. But we can assume that if we build unison without gui we don't want to build the gui

```
# context     2.1.6 | macos/x86_64 | ocaml.5.2.0 | https://opam.ocaml.org#25082ca0
# path        ~/.opam/default/.opam-switch/build/unison.2.53.4
# command     ~/.opam/opam-init/hooks/sandbox.sh build make NATIVE=true -j 3
# exit-code   2
# env-file    ~/.opam/log/unison-684-e03f6b.env
# output-file ~/.opam/log/unison-684-e03f6b.out
### output ###
# (cd uimac; \
# [...]
# 	 printf "MARKETING_VERSION = 2.53.4\nOCAMLLIBDIR = /Users/nayu/.opam/default/lib/ocaml\nOCAMLLIB_UNIX = -lunix${LIB_SUFFIX}\nOCAMLLIB_STR = -lcamlstr${LIB_SUFFIX}" > ExternalSettings.xcconfig)
# fsmonitor implementation is not available or not configured for this system. fsmonitor will not be built.
# ocamlopt: ubase/umarshal.ml ---> ubase/umarshal.cmx
# ocamlopt -g -I lwt -I ubase -I system -I +unix -I +str -I system/generic -I lwt/generic -c /Users/nayu/.opam/default/.opam-switch/build/unison.2.53.4/src/ubase/umarshal.ml
# ocamlopt: ubase/rx.ml ---> ubase/rx.cmx
# ocamlopt -g -I lwt -I ubase -I system -I +unix -I +str -I system/generic -I lwt/generic -c /Users/nayu/.opam/default/.opam-switch/build/unison.2.53.4/src/ubase/rx.ml
# (cd uimac; xcodebuild -arch x86_64   SYMROOT=build)
# xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory '/Library/Developer/CommandLineTools' is a command line tools instance
# make[1]: *** [macexecutable] Error 1
# make[1]: *** Waiting for unfinished jobs....
# make: *** [src] Error 2

```

EruEri:
One example to get in this situation, is to install homebrew, homebrew requires you to install the `Command line tools`  but after that if you are not interested in build Apple apps (IOS, MacOS, ...) you won't install XCode. Furthermore Xcode takes a lot of place, so I think if you don't need it you won't install it

src/Makefile.OCaml
```diff
@@ -51,6 +51,10 @@ OCAMLLIBDIR:=$(shell $(OCAMLC) -config-var standard_library)
 #
 OCAMLFIND := $(shell command -v ocamlfind 2> /dev/null)
 
+# Any invocation of the non-Xcode version of xcodebuild will trigger an error message
+# and an non 0 exit code. 
+XCODEBUILD := $(shell xcodebuild -version > /dev/null; echo $$?)
+
```
gdt, line 57: Is this being run on non-mac?  (It shouldn't be.)

src/Makefile.OCaml
```diff
@@ -51,6 +51,10 @@ OCAMLLIBDIR:=$(shell $(OCAMLC) -config-var standard_library)
 #
 OCAMLFIND := $(shell command -v ocamlfind 2> /dev/null)
 
+# Any invocation of the non-Xcode version of xcodebuild will trigger an error message
+# and an non 0 exit code.
```
gdt, line 55: This doesn't in my view unconfuse people who don't already understand.  The idea that there is a non-Xcode version or codebuild is surprising.   I'd like to see that explained (so that anyone maintaining the makefile will be enlightened).

src/Makefile.OCaml
```diff
@@ -72,7 +76,12 @@ endif
 
 .PHONY: macuimaybe
 ifeq ($(OSARCH), Darwin)
-  macuimaybe: macui
+  ifeq ($(XCODEBUILD), 0)
+    macuimaybe: macui
+  else
+    macuimaybe:
+      $(info Not the xcode version of xcodebuild, macOS native GUI can not be build.)
```
gdt, line 83: I would suggest
```
Not building macOS native GUI because XCode is not installed.
```
if you think that's accurate of course.

gdt (commented):
I'd prefer to align this with "On macOS, don't build native GUI if XCode is not installed" and leave the "non-XCode version of xcodebuild" confusing detail to the detail.  (The real point of 1st-line commit messages is to have the first 54 characters give enough of a clue that someone can get the big picture and know if they want to read the whole commit or not.   I know I'm being picky, but that's partially for future me looking back, and partially for others :-)

src/Makefile.OCaml
tleedjarv: Is it needed to assign `XCODEBUILD` at all? Just have the `$(shell)` invocation directly in the `ifeq`.

src/Makefile.OCaml
EruEri: Yes it is, a lot nicer that mine

src/Makefile.OCaml
EruEri: > Is this being run on non-mac? (It shouldn't be.)

Ok thanks I will move the check in the case where we are in `Darwin`, `ifeq ($(OSARCH), Darwin) ...`

tleedjarv:
I can't comment on the Mac particulars but overall I'm fine with this patch. Just a couple of nit-picks. Move the comment into the first `if` just next to the second `if`. Double check indentation and tabs vs spaces.

src/Makefile.OCaml
gdt: Yes, that's better.  The basic issue is that many people, including people that build open source code on macs, are unclear on the twisty maze of CLT and XCode and how it targets various SDKs.   It doesn't help that CLT installs an xcode-build, or that you seem to have to install XCode but not install it to get CLT.    Ideally this would be well-explained someplace on the net and we could point to it in INSTALL - that's not about this PR, though.

EruEri:
> ust a coupl

I move the comment, Is it what you meant ??

src/Makefile.OCaml
EruEri: Ok thanks you.

gdt:
Looks good to me - thanks for engaging with review/fixups.   I have enabled the workflow again to run regression tests.

tleedjarv:
> I move the comment, Is it what you meant ?

Yes, thank you.

I've only looked at the patch in GH but indendation and spaces-vs-tabs look broken to me. Please fix before it gets merged.

src/Makefile.OCaml
```diff
@@ -72,7 +72,14 @@ endif
...
 .PHONY: macuimaybe
 ifeq ($(OSARCH), Darwin)
-  macuimaybe: macui
+  # If XCode is not installed, xcodebuild is just a placeholder telling that XCode is not installed
+  # and any invocation of xcodebuild results in a non 0 exit code.
+  ifeq ($(shell xcodebuild -version > /dev/null; echo $$?), 0)
+    macuimaybe: macui
+  else
+    macuimaybe:
+      $(info Not building macOS native GUI because XCode is not installed.)
```
tleedjarv, line 81: While I don't think it is required here, for consistency with other recipes, I propose to replace spaces with a tab here:
```suggestion
	$(info Not building macOS native GUI because XCode is not installed.)
```

src/Makefile.OCaml
```diff
@@ -72,7 +72,14 @@ endif
...
+    endif
```
tleedjarv, line 82: I've only looked at the patch in the web interface but indentation looks wrong here.
```suggestion
  endif
```

src/Makefile.OCaml
gdt: Nevermind, I get it now.   I withdraw the question about style guide.

@EruEri The issue is that in makefiles, the commands to implement a target are indented with a tab character.  The rest of the file does it that way.

tleedjarv: No, and probably not. I only propose a tab here because regular recipes require a tab, so that's how all recipes have been written. Even those that are not 'regular'.

gdt:
There were spaces before the tab ;-)  To just finish I just removed them and forced-push.

gdt:
Thanks again for being cooperative with our software process!

src/Makefile.OCaml
EruEri: Maybe we can indent target with spaces and indent recipes with the same amout of space + the required tabulation ?

EruEri:
> Thanks again for being cooperative with our software process!

You are welcome, it's a pleasure to help.

Thanks you @gdt, @tleedjarv for your time.

src/Makefile.OCaml
gdt: If so, the spaces would have to be after, and I think that's messier.  I called it good enough as I think we've hit diminishing returns....

PR #1046, opened 2024-06-10
Remove FAT-inspired DST workaround

gdt:
In addition, clean up related dead code.

### DIFF
--- a/src/props.ml
+++ b/src/props.ml
@@ -478,9 +478,9 @@ let extract t = match t with Synced v -> v | NotSynced v -> v
 let minus_two = Int64.of_int (-2)
 let approximate t = Int64.logand (Int64.of_float t) minus_two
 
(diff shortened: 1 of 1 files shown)


gdt:
Someone pointed out to me that the time check is also used for update detection, and that this will mean that people syncing to/from FAT will have a full scan/hash every time the timezone changes (DST or a zone change).   Of course, a subset of these cases (changes that aren't 1h) result in rescans now.

gdt:
I don't remember why this hasn't been merged alread, so I'm going to merge it in 24h.     Overall I am increasingly unwilling to impose any negatives to accomodate platforms with problems, such as not having stable timestamps.

gdt:
My analysis is that the failing checks are CI infrastructure  bugs, not problems in this PR.

PR #1072, opened 2024-10-25
Bump version to 2.53.6 and set release date

gdt:


gdt:
This release date is assuming there are no or only minor issues (regresssions) reported from testing.  If there are problems, the PR will be modified and force-pushed.

PR #1081, opened 2024-11-07
feat(src/os.ml,-src/strings.ml): Introduce tempkeepdays parameter to config after how many days should be temporary .unison*.tmp files deleted

LubosKolouch:


gdt:
There's surely a backstory about why you think this is necessary, which is perhaps about other problems.   Please post on unison-hackers@ about the situation, motivation and analysis about relative benefit vs increasing the already-too-large configurablity.

tleedjarv:
`strings.ml` is a generated file and shouldn't be modified directly (it will simply be overwritten anyway).

Adding preferences like this is unfortunately not as simple as that as it would break backwards compatibility with older versions (try it with a server built without this patch).

acolomb:
> relative benefit vs increasing the already-too-large configurablity

Since comments like this come up regularly, I'd like to point out that this assessment is subjective.  Having options is not bad in itself.  If they're well documented and sensibly structured, that's fine.  Sure, I don't use 99 % of them at all, but certainly other users may have a need for them, which we'll never hear about.  And since the code is already there and tested, we shouldn't be pushing hard to remove parts of it again IMHO.  Of course, it's nearly impossible to test a system with a myriad of parameters completely, but we'll need to accept that 100 % test coverage is not feasible, unless we make the whole system much simpler (and thus useless).

Sure, whether this additional parameter is acceptable is to be discussed.  But it looks rather simple and logical to me, so why not accept a patch that someone has already put some effort into?  Provided it integrates well of course...

gdt:
Belatedly setting feedback.    This PR lacks
  - an analysis of usefulness to many vs complexity
  - an explanation of interoperability with servers with and without the option

LubosKolouch:
I'm sorry but your requirements are dumb and impossible to do.

Just close the PR.

🤷🏻‍♂️

PR #1089, opened 2024-12-06
GHA: Add i386 Ubuntu GUI build

tleedjarv:
There was a comment in the workflow file, stating that the 32-bit (i386) GUI build was broken. It works when installing the i386 version of GTK.


gdt:
Is this building an i386-mode unison-gui, on an x86-64 system?   Should we be thinking about doing a full build within an actual i386 environment?  Or has the GNU/Linux world given up on that?

(I'm not opposed to this; just making sure I'm clear on why this vs what I was expecting from 'i386'.)

tleedjarv:
> Is this building an i386-mode unison-gui, on an x86-64 system?

That is correct.

> Should we be thinking about doing a full build within an actual i386 environment?

Not sure if this is possible with GHA. It could be with docker or qemu...

> Or has the GNU/Linux world given up on that?

Not sure, this could also be true. I just went for the path of least resistance (as you can see, the patch is tiny).

PR #1098, opened 2024-12-17
Portable makefiles 8

tleedjarv:
This PR concludes the series aimed at making the makefiles portable. (Build instructions are yet to be updated.)

As a result, building Unison now works with GNU Make, BSD make, Solaris make and NMAKE. GNU tools and a POSIX shell are no longer needed. Additionally, building the manpage and docs works with all of the above, except with NMAKE.

The first commit could theoretically be split into smaller pieces but I chose not to do it for a few reasons. First, I wanted to keep it together for the history. Reviewing it in detail is likely not feasible anyway, and ensuring it does not break anything is a task better achieved via CI and manual testing.

A short summary of what is going on: The series of PRs so far has gradually changed/simplified the makefile contents so that it becomes acceptable for serveral `make` implementations. Any remaining parts that can't be made portable with simple syntax changes have been moved out into a separate script (the first commit in this PR). The script is written in OCaml because OCaml is required for building anyway and this way we can be completely portable. Any remaining shell invocations in the makefiles have been simplified to the degree that they become acceptable for the shell in Windows. This way we no longer depend on GNU tools, a POSIX shell, or otherwise have specific expectations on the environment. The separate script is invoked automatically every time `make` is run to generate configuration files that are included in the makefile. The script uses a syntax that, if you squint really hard, resembles the makefile a little bit.

Nothing changes from the user's perspective, except they no longer have to use `gmake`.

I have tested this with all of the make implementations mentioned above (some manually, some with CI). BSD make has been tested in NetBSD, FreeBSD, OpenBSD and DragonFly BSD. Solaris (d)make has been tested in OmniOS and Oracle Solaris.


gdt:
Builds on NetBSD 10 with gmake, make(bsd), with and without -j32.

PR #1101, opened 2024-12-20
Improve libinotify detection on BSDs

tleedjarv:


gdt:
I tried to read the code and construct the test program build line:
```
ocamlc -ccopt "-I/usr/pkg/include" -cclib "-L/usr/pkg/lib -Wl,-rpath/usr/pkg/lib -linoti" inotifytest__.c && echo OK
```
and then I munged the lib to misspell (pasted above) and it still passed.  It does fail without the -I part.

I think what's going on is that this is compiling the .c, but there  is no link step so the unresolved symbol is not objected to.
In autoconf speak, I think we need AC_TRY_LINK, and not just AC_TRY_COMPILE.

src/make_tools.ml
```diff
@@ -289,16 +289,26 @@ let fsmonitor_dir =
...
+            "{ printf '#include <sys/inotify.h>\\n\
+             void test() { int fd = inotify_init(); }' > inotifytest__.c ;";
+            ocamlc;
+            ($)"CAMLCFLAGS";
+            ($)"CAMLLDFLAGS";
+            libinotify_inc;
+            libinotify_lib;
+            "-o inotifytest__ inotifytest__.c";
+            "> /dev/null 2>&1 && printf true ; } ;\
+             rm -f inotifytest__.c inotifytest__.cm[oix] \
```
gdt, line 307: Why is .cm[oix] being removed?   I don't see that they are generated now.

tleedjarv:
> I think what's going on is that this is compiling the .c, but there is no link step so the unresolved symbol is not objected to. In autoconf speak, I think we need AC_TRY_LINK, and not just AC_TRY_COMPILE.

You're right... but now that I think about it, there's no need for this test at all. I wrote it before relying on pkg-conf. If we require pkg-conf then we don't need to do this test compile/link at all and can just use the output of pkg-conf directly, no?

src/make_tools.ml
tleedjarv: Just a leftover...

gdt:
I looked at two random programs that use pkg-config via autoconf, and they both just believe the output.  Perhaps it is wise to use `--exists` first, and if true, just set flags based on --cflags/--libs.

tleedjarv:
> I looked at two random programs that use pkg-config via autoconf, and they both just believe the output. Perhaps it is wise to use `--exists` first, and if true, just set flags based on --cflags/--libs.

Simplifying sounds like a good idea. Let's do just that. If this backfires in the future, we can always revisit this.

gdt:
Thanks for the multiple iterations.   I have tested via pkgsrc-wip (before the latest if clarification but that's good enough) and it works both ways, with libinotify apparently present and apparently not present.

PR #1113, opened 2025-01-20
More minor GUI tweaks

tleedjarv:
More details in commit messages.


tleedjarv:
(As always, GHA fails are not related to this PR.)

Issue #1118, opened 2025-02-05
unison-2.53.7-ubuntu-x86_64-static.tar.gz is a tar bomb


andreasabel:
The released binary package https://github.com/bcpierce00/unison/releases/download/v2.53.7/unison-2.53.7-ubuntu-x86_64-static.tar.gz is a _tar bomb_, this is its directory:
```
CONTRIBUTING.md
INSTALL.md
LICENSE
NEWS.md
README.md
bin/
bin/unison-fsmonitor
bin/unison
doc/
doc/unison-manual.html
doc/unison-manual.txt
doc/unison-manual.pdf
man/
man/unison.1
```
Not sure what the idea here is.  Is one supposed to extract this to `/usr/local/` or similar?  But then one would also get the `CONTRIBUTING.md` etc. there which does not belong there.

I suppose it would make sense to place everything in the archive under `unison-2.53.7/` or similar so that one does not shoot oneself in the foot when unpacking the archive by accident in the wrong directory.

I have not checked the other released packages whether there are more tar bombs.

P.S.: Thanks for the unison tool, I have been using it for a quarter century now!

gdt:
Unison's issue tracker recently has templates, and it seems they might not be working.  Before I or anyone else considers the content of your issue, I'd like to understand the meta issues.

When you created the issue, were you presented with a menu of 4 choices?  There should be "Defect Report", "Fully-Designed Feature Request" and two mailingilst pointers.   If not, what happened?

If you clicked Defect Report, as I am guessing you did, you should have gotten a template that told you to read https://github.com/bcpierce00/unison/wiki/Reporting-Bugs-and-Feature-Requests .   Did you, or did you get something else?

Assuming you did get the template text, and you read the wiki page, did it contain the text "The CI-provided binaries exist for Continuous Integration and are useful for users as a side benefit. Therefore, the CI binaries not working on a particular operating system is not a Unison bug (there might be Unison bug lurking behind that, but there might not). In general, binaries should be provided by packaging systems." ?

The template said to leave the statement that you read the bug reporting wiki page, but it's missing in your issue as created.  Do you understand what happened here?

andreasabel:
You only gave a formalist answer.

Yet I do think there is actually a chance for improvement, and it is not a high-hanging fruit.

@gdt wrote:

> Therefore, the CI binaries not working on a particular operating system is not a Unison bug (there might be Unison bug lurking behind that, but there might not). In general, binaries should be provided by packaging systems." ?

[...]

PR #1123, opened 2025-02-17
GHA: Include the subdir in tarball

tleedjarv:
Inspired by #1118, I think this was actually a bug in the original CI script, which was then also assumed by the _compat builds. This was very simple to change, so why not go for it.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -548,7 +548,7 @@ jobs:
 
     - name: Package
       shell: bash
-      run: cd '${{ steps.vars.outputs.PKG_DIR }}/' && ${{ steps.vars.outputs.COMPRESS_CMD }} '../${{ steps.vars.outputs.PKG_NAME }}' *
+      run: cd '${{ steps.vars.outputs.STAGING_DIR }}/' && ${{ steps.vars.outputs.COMPRESS_CMD }} '${{ steps.vars.outputs.PKG_NAME }}' *
 
     - uses: actions/upload-artifact@v4
       if: matrix.job.publish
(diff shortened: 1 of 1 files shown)


gdt:
Is the idea that the SHA1 of the sources is in the tarball name?    I think that's important to reduce confusion about which tarball is which later, and I'm pretty sure that's what you would intend.   Looks like the commit is `46dbd65cdbcc60f68cfbb6798565cf3629bacabe`.  But an artifact is https://github.com/bcpierce00/unison/suites/34502718465/artifacts/2605140450 or at least that's the unhelpful link, and it downloads as `unison-git_81cd90ca-ubuntu-x86_64.tar.gz___ocaml-4.14.x.ubuntu_compat-publish.zip` with that hash inside.    That looks like a commit hash, but I can't find it (I have your repo as a remote, so it's been fetched, and `git log 46db` shows me the commit).

tleedjarv:
That's right, the dir name is the same as tarball name (sans .tar.gz).

With an unmerged PR you get a different hash because GHA does a merge for the build (just look at the logs at the checkout step, for example: https://github.com/bcpierce00/unison/pull/1123/checks#step:3:61). This is completely unrelated to this PR.

gdt:
That's quite the GHA bug, making it hard to tie unmerged PR test artifacts back, but it's definitely not a bar to this fix.

tleedjarv:
> That's quite the GHA bug, making it hard to tie unmerged PR test artifacts back, but it's definitely not a bar to this fix.

I think there's a way around it (not sure why it's not the default). I'll try it out and open another PR.

PR #1124, opened 2025-02-19
GHA: Use SHA from PR HEAD commit for pkg name

tleedjarv:
I'm opening this PR for discussion (see https://github.com/bcpierce00/unison/pull/1123#issuecomment-2667888894) but I'm not sure it should be merged, as there are some caveats.

The CI script still builds the merge commit. It is possible to change this and build the PR HEAD, but I'm not sure it is a good idea. The point of CI is to run the build tests before merging and the PR HEAD might be way off from the actual merge result (if it hasn't been rebased recently, for example).

With CI building the merge commit (which I think is a good idea), is it actually a good idea to name the build artifacts with the SHA from PR HEAD? If the PR was not rebased recently then the build artifacts are not based off that HEAD, strictly speaking.

(Note that all of this discussion only matters for open PRs; CI builds on merges, direct pushes and tagging are unaffected.)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -82,12 +82,13 @@ jobs:
         STAGING_DIR='_staging'
         outputs STAGING_DIR
         # parse commit reference info
+        GIT_SHA=${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
         echo GITHUB_REF=${GITHUB_REF}
-        echo GITHUB_SHA=${GITHUB_SHA}
+        echo GIT_SHA=${GIT_SHA}
         REF_NAME="${GITHUB_REF#refs/*/}"
         unset REF_BRANCH ; case "${GITHUB_REF}" in refs/heads/*) REF_BRANCH="${GITHUB_REF#refs/heads/}" ;; esac;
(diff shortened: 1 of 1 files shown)


gdt:
I see your point about the merge commit being necessary.   As someone who believes in vigorous rebasing, I wasn't thinking about the PR commit and the resulting merge having different trees.   And, I also see your point about not labeling an artifact to be about the tip commit on a PR when it is the merge that is built.

The problem is that this makes it hard to tie the artifact back to the sources it came from.

Perhaps put "unmerged-pr-" or something in the artifact name, so we at least know it's off the rails?   That's not strictly necessary, as the sha1 won't be in the repo, but it's a clue that it's a testing artifact vs one from a longer-term branch.

tleedjarv:
Added a commit for testing out the "unmerged-pr-" idea.

gdt:
I checked an artifact and see `unison-unmerged-pr-1124-git_979aeca3-ubuntu-x86_64` which looks good; it has the sha1 of the commit and the unmerged warning. 

I am inclined to hit merge if you think it's ready.

tleedjarv:
I guess so. Might just as well merge.

Issue #162, opened 2018-02-21
Add a .desktop file


FSMaxB:
It would be great to add a .desktop file by default when installing the GUI version. Look [here](https://git.archlinux.org/svntogit/packages.git/tree/trunk/desktop?h=packages/unison) for an example.

This would also require installing an icon to `/usr/share/pixmaps`. Preferably `U.svg`.

bcpierce00:
Happy to consider a PR.

FSMaxB:
I don't understand how `make install` works for unison to be honest. Usually there is `DESTDIR` that can be set for installation directory. But unison doesn't seem to do that. It uses "$(HOME)" instead.

Maybe that should be fixed first. (`DESTDIR` could be set to `$(HOME)` by default to preserve the existing behavior.

bcpierce00:
I don't think that many people use "make install," other than me.  :-|

Your proposed refinement sounds good.

FSMaxB:
It would be easier if:
* unison would build executables with different names for different `UISTYLE`. e.g. `unison-text` `unison-gtk2`
* there was a way to know what has been built when running `make install`, which could just be the filenames from the build step

bcpierce00:
These also seem like reasonable refinements.  One could retain backward compatibility by copying the latest-built binary (unison-text or whatever) to 'unison'.  

The reason we never did something like this is that there are actually just three UI configurations (text, GTK, and the OSX native one); of these, the build and install procedure for the third is separate from the other two, while the GTK configuration supersedes the text one (i.e., it can be used in both ways).

Issue #705, opened 2022-04-20
Reconsider project hosting


gdt:
It's been a long time since the project arrived on github, and thus we should probably evaluate if that's still the best choice.  Missing features are CI for a variety of platforms, mailinglists, and being Free Software.

gdt:
Some possible alternatives are listed at https://sfconservancy.org/GiveUpGitHub/ and concerns about license violations are discussed at https://sfconservancy.org/blog/2022/jun/30/give-up-github-launch/

gdt:
Github CI has required more attention lately, and seems to be withdrawing support for platforms that don't seem super old.  See #445 for more discussion

mvglasow:
Since you mentioned Codeberg in #443 – what are the requirements for a future hosting platform, and what hosting platforms are you currently contemplating?

gdt:
Not fully baked, but my take on requirements is:
  - site runs Free Software
  - site is a charity with a mission towards Free Software, more or less, the point being that it's legally obligated to act in the public interest instead of the interests of shareholders.   Or maybe, the project after getting a fiscal sponsor can buy service with donations.
  - runs git, has wiki, issue tracker, way for people to submit proposed changes, etc.
  - well run, mostly up, etc.
  - has a track record of sanity and one can make a case for future sanity
  - we can somehow do CI including public access to artifacts
    * as part of this, we should be able to do CI on more Free platforms and CPU types
    * Windows and mac are hard (license costs), so those we can consider (and donations if a mechanism), but losing them would be a big change.  I'm personally ok with non-Free platform CI happening only if users pay for CI costs, as long as there is some reasonable mechanism for them to do so.  But I expect others would be upset by that.
  - would be really nice if it had mailinglists, but not a requirement
 
If you get the idea this is not fully baked, you would be correct!

gdt:
Addition to requirements:

  - does not train LLMs on project code (which is a violation of the license)
  - does not suggest that users perform any actions as the result of an LLM, to include issues, PRs, code changes in PRs, or anything else

Relative to that addition, see https://www.theregister.com/2025/09/05/github_copilot_complaints/

mvglasow:
I  have since gotten involved in CoMaps, a relatively new fork of Organic Maps. CoMaps is on Codeberg (OM is on Github), so I am somewhat familiar with the platform.

* Free Software – yes, Forgejo
* Charity – yes, nonprofit Codeberg e.V.
[...]

PR #1128, opened 2025-05-25
add desktop file and install it with unison-gui

mistmist:
Taken from the Flatpak repo
https://github.com/flathub/io.github.bcpierce00.unison and originally based on the Fedora rpm source.


gdt:
Thanks for splitting.    I see the MIT-0 SPDX and given that this is near trivial in creative expression that seems adequate.    Thanks also for pinging the likely author.  I did a build/install to a temp DESTDIR and structurally it seems good.    I wondered why in `make_tools.ml` it doesn't use `build_GUI` but I see what you added is in the same vein as the rest of that function, so that's fine.

Two nits, and I can adjust, or you:

I don't want to encode github in the desktop name.  It's a [bug](#705) that unison is still on github.  Looking at my system (NetBSD/pkgsrc), I have 82 desktop files.  Using the horrifying regexp `'.*\..*\..*'` to find those with a dot in the application name, there are then 25, and *all* of them are domain names belong to the projects, like `gnome.org`, `kicad.org`, `qgis.org`.  There isn't a single `foo.github.io`.   There are 58 that are just the app name, or project-app for desktoppy stuff; an example is `emacs.desktop`.   Given that there could be a desktop-file way to start a background non-gui unison from a desktop file in the future, I'd lean to `unison-gui.desktop`.    Syncthing has `syncthing-start.desktop` and `synching-ui.desktop`.

The comment says `Multi-master File synchronization tool`.   That's not a phrase that i find in unison's documentation.   I would lean to "GUI for Unison file synchronizer".

(I am assuming that packaging systems would be happier to have an upstream-provided desktop file, and drop their local one, seeing that as progress, even if it is a bit different.)

mistmist:
Flathub has a very detailed and inflexible policy about Application ID which also determines the desktop file name, so i had to follow that.

https://docs.flathub.org/docs/for-app-authors/requirements#application-id

https://docs.flatpak.org/en/latest/conventions.html#desktop-files

changing this from io.github.* effectively requires deprecating the existing application and submitting it under the new ID - certainly doable but only after you have actually established the new hosting location.

however, replacing the last part with "unison-gui" is possible:

> Flatpak will export the following desktop filename patterns: $FLATPAK_ID.desktop, $FLATPAK_ID.foo.desktop, $FLATPAK_ID-foo.desktop.

data/io.github.bcpierce00.unison-gui.desktop
```diff
@@ -0,0 +1,11 @@
+# SPDX-License-Identifier: MIT-0
+[Desktop Entry]
+Type=Application
+Exec=unison-gui
+Name=Unison
+GenericName=File Synchronizer
+Comment=GUI for Unison file synchronizer
+Terminal=false
+Icon=io.github.bcpierce00.unison
```
mkrupcale, line 9: Must the `.desktop` `Icon` entry also contain the Flatpak application ID? And does Flatpak then rename the icon files during install / Flatpak creation (e.g. `U.svg` [1] to `io.github.bcpierce00.unison.svg`)?

For the Fedora package, I rename `U.svg` to `unison.svg` and use `Icon=unison` [2,3], consistent with the Icon Theme Specification [4].

1. https://github.com/bcpierce00/unison/blob/master/icons/U.svg
2. https://docs.fedoraproject.org/en-US/packaging-guidelines/#_icon_tag_in_desktop_files
3. https://specifications.freedesktop.org/desktop-entry-spec/latest/extra-actions.html#id-1.12.4.3
4. https://specifications.freedesktop.org/icon-theme-spec/latest/#icon_lookup

mkrupcale:
> 
>     * it would be nice to rename U.svg to unison.svg and Unison.svg to unison-word.svg in the repo, and add a README, but it's not a big deal.

Yeah, that would work fine. BTW the `icons/U.${size}x${size}x16m.png` files should be named consistently with `U.svg` as well to avoid confusion. The SVG format is technically an optional format of the Icon Theme Spec, so we also have to install the png format icons.

On the question of naming `unison.desktop` / `unison-gui.desktop`: I don't have too much preference one way or another (although I'd be curious what intention there is on making some non-GUI/background unison desktop launcher necessitating `unison-gui.desktop` as well as `unison.desktop` in the future), and the Desktop Entry Specification [1] only requires that the name be a valid D-Bus well-known name (a sequence of dot-separated elements containing ASCII letters, digits, dash, and underscore). They do suggest (with the language _should_) that it use the reverse DNS convention, which seems similar to the Flatpak application ID format, but as you pointed out, many applications do not follow that convention.

[1] https://specifications.freedesktop.org/desktop-entry-spec/latest/file-naming.html

gdt:
As for intent, I view unison as primarliy a CUi program and it can run as a daemon.   So the GUI is a special case, and e.g. syncthing has a desktop entry to start it in the background as a daemon, and then another to run the gui.   Hence I lean to make all unison-foo now, to avoid wanting to change. 

Thanks for the png/etc hint.  I did realize that but it's not safe to assume :-)

I get the point of reverse dns as a unique source, but unison simply doesn't have such a name, and isn't likely to get one.

gdt:
I reviewed this again before the release, and my assessment is that a number of issues raised in discussion are not resolved:
  - filename uses github
  - icon strategy is unclear (not installed? naming?)

Thus today this PR, as it stands, is not mergeable.    I added feedback, not as the meaning in issues of request for information, but in the PR sense of waiting for changes.

mistmist:
i completely agree with your assessment, and in any case it's not urgent...

my memory of the comments here is that there would be some discussion on a mailing list (but no links to any results of the discussion), and there are some unresolved questions:

* the ID (and icon/.desktop name) should be "unison" or "unison-gui"? (since neither are acceptable to flathub i have no opinion which one to use)
* this PR seeks to only add a .desktop file, if you like then it could install the icon as well (since it's referenced by the .desktop file), but i'm not sure if this is what you and/or distro packagers want

gdt:
I realize desktop is gui-ish but there is also the notion of running unison as a daemon without the gui.   Looking at my system's collection of files, the vast majority are for programs that are only GUI.   I see syncthing, which has
```
syncthing-start.desktop
syncthing-ui.desktop
```
but of course there are text UIs so I think I land on `unison-gui`.

I see your point about Icons being separable, but I don't see how we can install a desktop file that points to an icon that isn't installed.  We have a bunch of icons in the sources, and it's a little messy.   I would prefer that some grasp/add-README/rationalize that and figure out which icons perhaps should be installed to align with how the broader Free Software world does things.  I think that's putting a few different resolution in share/icons/hicolor.

I lean to using our existing icons if that's at all reasonable, and maybe deprecating the one that is natively in Adobe Illustrator because we shouldn't use non-Free tools.  I think the svg ones were created because of that reason.
But it's a mess, and it would be great if someone cleaned things up.     This really needs to be raised on hackers, as project doctrine is that the project is centered on mailing lists, not github.   I'll send a note.

gdt:
See #1163

tleedjarv:
Merging this PR closes #162

gdt:
Thanks - I think we're either there or super close.

About the icons: this is taking U from sources and putting it in unison-gui, which matches the desktop name.   Probably we should rename U, but that's ok and not this PR's problem.   I'm guessing that having an icon name match the desktop entry name is normal.   Plus, it seems like we could change it later, with minor packager hassle.   Is this a correct take on the icon status?

mkrupcale:
> Thanks - I think we're either there or super close.

Yeah, the current PR looks reasonable to me, although I have not yet tried building it myself.

> About the icons: this is taking U from sources and putting it in unison-gui, which matches the desktop name. Probably we should rename U, but that's ok and not this PR's problem. I'm guessing that having an icon name match the desktop entry name is normal. Plus, it seems like we could change it later, with minor packager hassle. Is this a correct take on the icon status?

Sounds about right to me. If the installed icon name changes, yes, it'll just be a minor change to the package `files` list [1] for me.

1. https://src.fedoraproject.org/rpms/unison/blob/797f3253bc6934d853285e908520fbcae4992593/f/unison.spec#_142

gdt:
It would be great if you are up for building.   CI is running, with spurious failures, looking like it will be ok.

gdt:
Thanks @mistmist for perservering through a lot of little things.

PR #1139, opened 2025-07-29
Fix building with new version of BSD make

tleedjarv:
BSD make now has support for `$^` as specified in POSIX. This breaks the current build logic subtly. Detect the support for `$^` as a workaround (unfortunately, detection can't be done directly in Makefile as it's portable and NMAKE considers any use of `$^` and `$>` a syntax error).


gdt:
Passes CI and tests ok on NetBSD 10 (old BSD make) and NetBSD-current, to be 11 (new BSD make).

PR #1140, opened 2025-07-31
More fixes for BSD make

tleedjarv:
BSD make is still actively adding new features and some of the features are recent enough that they can't be used blindly without testing their presence.

This PR improves feature test for `$^` (added in #1139) and adds a feature test for target-specific local variables (supported by BSD make since NetBSD 10/FreeBSD 13).


gdt:
I tested with make (BSD) and gmake on NetBSD 9, 10 and current, all 6 combinations ok.

PR #1146, opened 2025-10-03
CI: label artifacts with OS versions

gdt:
This is primarily so that it is clear which ubuntu version was used, as builds are unlikely to work on older ones.


tleedjarv:
I'd skip the labels for Windows builds. The GHA runners are based on Windows Server version which is meaningless for most users. Also, Windows version compatibility story is simpler, so not having a label is acceptable. In fact, the actual Windows version constraints come from OCaml runtime, but these are not well documented or officially stated anywhere, to my knowledge (silent compatibility breakage has been the norm).

gdt:
> I'd skip the labels for Windows builds. 

OK, dropped.   We're likely to be dropping windows builds entirely with the sooner-or-later move to Libre hosting....

PR #1152, opened 2025-10-16
Reset preferences to defaults in server process

tleedjarv:
Fix a potential bug when clients of different versions (feature levels) connect to the same server process.


gdt:
This is definitely a bug fix; the only potential about it is if it wil be triggered.  Passes CI and makes sense on review, so I'm just hitting merge.

PR #1158, opened 2025-10-17
NEWS: Add entry for 2.53.8

gdt:
This was generated by examining `git log --reverse --oneline v2.53.7..`, and omitting CI, bugfixes, minor things etc.

Not related to 2.53.8, but I dropped the now-old explanation that NEWS is in NEWS, as there's no trace of the old plan and I've forgotten it.

I carried forward deprecation warnings, and added a caution that the Mac UI is not being maintained.

As for the date, I just added a year from the last release.  I don't feel that we're in a rush, and a bit of testing time seems fine, but it's not cast  in stone.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,10 +8,40 @@ release contains some.  Refer to the documentation for details; this
 file is a terse notice of changes rather than a tutorial about new
 features.
 
-As of 2022, this file (NEWS.md at top level) is used for news.
-
 Some software has a "changelog" file that records all changes, but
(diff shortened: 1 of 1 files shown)


tleedjarv:
Changes in makefiles deserve some further explanation as they were in some way quite invasive and could cause packagers' build scripts to fail. I'll propose something later.

gdt:
Text adjusted for those points (using some of your wording from off-github comments).  I'll leave it pending for a little bit.

PR #1163, opened 2025-11-05
Icon cleanup

gdt:
This PR is a massive, heavy-handed cleanup of the icons directory.  It weighs technical debt more heavily than the risk of deleting something, especially because what is removed remains in git history.

### DIFF
--- a/icons/README.md
+++ b/icons/README.md
@@ -1,0 +1,48 @@
+# Unison icons
+
+This directory contains icons for use with unison.
+
+# Known purposes
+
+## Unix Desktop Environments
+
+The square icon (U.svg) could be installed in share/icons/hicolor for
+use by desktop environments.
+
+The
+[icon theme specification](https://specifications.freedesktop.org/icon-theme/latest/)
+requires a raster format, preferring png, and optionally svg.
+The minimal requirement is a 48x48 icon for hicolor.
+
+## Windows
+
+Windows needs a
+[`.ico`](https://en.wikipedia.org/wiki/ICO_(file_format))file.
+Currently one is present in src/win32rc, but there is no information
+about where it came from or how it was created.
+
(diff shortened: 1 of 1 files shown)


gdt:
CI passed.  If there are no objections, I'll merge after 12 November at 18Z.

tleedjarv:
~~Do not remove U.ico, it is used by src/win32rc/unison.rc (for Windows builds).~~
_Edit_: On second thought, this was wrong. src/win32rc/U.ico is a copy of icons/U.ico so I believe icons/U.ico can be removed.

I would guess it is safe to remove *.bmp *.gif *.jpg (and I believe *.xpm, too).
I would also guess it's safe to remove unison-icon.svg (or U-old.svg).

gdt:
Probably we should:
   - add to icons/README about this
   - define how to generate U.ico from U.svg
   - have src/win32rc reference ../../icons instead of having a copy

mkrupcale:
Looks reasonable to me. As I said at [1], the Icon Theme Spec [2] only requires the png (or xpm for backward compatibility reasons) files with optional SVG support, which the remaining `U.${size}x${size}x16m.png` and `U.svg` provide, respectively.

1. https://github.com/bcpierce00/unison/pull/1128#issuecomment-2907869021
2. https://specifications.freedesktop.org/icon-theme/latest/

gdt:
Thanks.  I added that link to the README and raised the issue of renaming U to unison and unison to unison-word.   Perhaps that step should happen in this branch.

tleedjarv:
The .ico file is apparently a container of some sorts.. Try:
```
icotool -l U.ico
icotool -x U.ico
ls U_*.png
```

gdt:
Thanks; that was a great hint.   Updated with information about the ico file.

PR #1166, opened 2025-11-06
Fix compilation on MSYS2

### LINKED ISSUE #1164: Error when building with MinGW : ` ./copy_stubs.c:369:7: error: request for member 'ByteCount' in something not a structure or union`
I'm the maintainer of the Opam package for unison, and I am trying to publish the package for Unison 2.53.8.

Things seems working, except for Windows, with the MinGW port. I get the following errors: 

```
   # ./copy_stubs.c:369:7: error: request for member 'ByteCount' in something not a structure or union
  #   369 |   dupl.ByteCount.QuadPart = Long_val(len) +
  #       |       ^
  # ./copy_stubs.c:384:31: error: request for member 'TargetFileOffset' in something not a structure or union
  #   384 |   newTargetLen.QuadPart = dupl.TargetFileOffset.QuadPart + Long_val(len);
  #       |                               ^
  # make[2]: *** [Makefile.OCaml:208: copy_stubs.o] Error 2
  # make[2]: *** Waiting for unfinished jobs....
  # make[2]: Leaving directory '/d/opamroot/default/.opam-switch/build/unison.2.53.8/src'
  # make[1]: *** [Makefile:42: all] Error 2
  # make[1]: Leaving directory '/d/opamroot/default/.opam-switch/build/unison.2.53.8/src'
  # make: *** [Makefile:22: src] Error 2
```

(See https://github.com/ocaml/opam-repository/actions/runs/19137603433/job/54693594142#step:14:88)

I have no Windows machine to test, but this seems related to https://github.com/mingw-w64/mingw-w64/issues/90, which added this type definition to MinGW. But looking at `copy_stubs.c`, it seems like it should be handled by Unison if MinGW is too old...

Any idea what's happening? Of cours, I could just disable the support for Windows, but this is lame.

tleedjarv:
(all defines moved up because some caml/*.h includes pull in windows.h)
Fixes #1164

@jhjourdan it would be great if you could test in opam-repository before this gets merged.

### DIFF
--- a/src/copy_stubs.c
+++ b/src/copy_stubs.c
@@ -15,6 +15,25 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+#ifdef _WIN32
+
+/* Minimum Windows API needed for getting the definition of
+ * struct DUPLICATE_EXTENTS_DATA with MinGW headers under MSYS2.
+ * As of 2025-11, MSYS2 by default targets Windows 7, which is
+ * not sufficient in this case. */
+#define WINVER 0x0603
+#define _WIN32_WINNT 0x0603  /* Windows 8.1 */
+
+#ifndef UNICODE
+#define UNICODE
+#endif
+#ifndef _UNICODE
+#define _UNICODE
+#endif
+
+#endif /* _WIN32 */
+
+
 #include <caml/memory.h>
 #include <caml/threads.h>
 #include <caml/unixsupport.h>
@@ -243,13 +262,6 @@ CAMLprim value unison_copy_file(value in_fd, value out_fd, value in_offs, value 
 
 #elif defined(_WIN32)
 
-
-#ifndef UNICODE
-#define UNICODE
-#endif
-#ifndef _UNICODE
-#define _UNICODE
-#endif
 
 #include <windows.h>
 #include <winioctl.h>


src/copy_stubs.c
```diff
@@ -15,6 +15,24 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+#ifdef _WIN32
+
+/* Minimum Windows API needed for getting the definition of struct
+ * DUPLICATE_EXTENTS_DATA with MinGW headers. This is only really
+ * needed with MSYS2, which otherwise gets it wrong. */
+#define WINVER 0x0603
+#define _WIN32_WINNT 0x0603
```
gdt: What does 0603 mean?  It's a magic number.

Is there a bug in msys2, or is i that msys2 uses the windows sdk, and that is documented to need to be told the oldest api, and if you don't, it provides a really old api, and this definition isn't part of that?  I'm still having trouble understanding if this is unison doing what it should, or if it's a workaround.

Setting this is presumably declaring the min version, and thus the resulting binary may not (probably does not) run on older.   I'm ok with that.   But this says API, and there's ABI. 
 
It could be that I'm being too difficult here, but I don't even understand today, and surely will not in a year or so.   I want wisdom to be in checked-in comments, not in discusssions, so future self will see it.

It might be that @jhjourdan should steal the PR and write the comments I want, or the ones I would want if I understood, since I think @tleedjarv does not really want to hack on Windows :-)

gdt:
And yes, testing would be great.

jhjourdan:
See https://github.com/ocaml/opam-repository/pull/28851: all lights are greens, the tests passed.

src/copy_stubs.c
jhjourdan: >  I've written Windows-specific code such that it will run fine under older Windows versions, it just needs to be compiled with newer headers present. I have also verified this by testing built binaries in Windows 7 (as the magic number here specifies Win 8.1).

Interesting: how is it possible that `FSCTL_DUPLICATE_EXTENTS_TO_FILE` is recognized by `DeviceIoControl` on Windows 7? In my understanding, this API appeared on Windows 8.1.

gdt:
It's great to see progress figuring this out.   My goal are simply that 1) the comments that are checked in (PR discussion doesn't count for posterity) explain everything to the future and 2) if this is a bug in msys, we at least note that, and perhaps file.  But it appears to be a choice, and maybe the bug is just that they don't document how to change it.

tleedjarv:
The thing is, this `WINVER` / `_WIN32_WINNT` is a Microsoft thing and normally defaults to whatever SDK you're building with (I think if you have recent toolchain that means the Windows version you're building on). You actively set the value _to target an older version_. This is nothing to write home about.

I don't know how MinGW sets the default value, but it appears that MinGW in Cygwin and MinGW in MSYS2 use a different default. The issue is that MSYS2 limits the API _for you_ and you then set the value _to target a newer version_. Which to me appears backwards and not what Microsoft intended. That's just what I think, I may be totally wrong, I've no experience with any of this.

I have now recorded all of this in the comment.

gdt:
Thanks -i think future self will be happy with this!

Issue #1122, opened 2025-02-14
Unison (fastcheck=true) doesn't notice new files (written in exFAT by another computer running Windows)


Artur2048:
# Instructions

----------------------------------------
I have thoroughly read the [Reporting Bugs and Feature Requests](https://github.com/bcpierce00/unison/wiki/Reporting-Bugs-and-Feature-Requests) wiki page.

# Meta

Running Unison to sync local drive with removable USB SSD.

# Environment

  - unison version 2.53.7
  - Linux Mint 21.3
  - ocaml 4.14.2)
  - issue affect both - GUI and CLI
  - both - local and SSH
  - use of fsmonitor: no

# Reproduction Recipe

After performing initial sync with Unison, I've unmounted and removed my USB SSD, plugged it to another PC and added some files and directories.
I have then plugged it back to my computer and started Unison again. After rescanning both folders it wrote that "Everything is in sync" (which is not true - there are new files there)

**Interesting fact - if I create additional file in exactly the same directory as the "undetected new files" on my local computer and rescan with Unison - it will detect all new files - including files created on another PC which were previously undetected**

Additional information - **my SSD is formatted as ExFAT**, and the profile PRF is:

```
root = /media/artur/Extreme Pro/Test1/
root = /home/artur/Test1/
fat = true
times = true
log = true
logfile = sync_test1.log
```

Steps to reproduce:

0. both roots (home and media) are in sync
1. unmount and remove "media"
2. add some files to some subdirectory on ANOTHER PC - Windows PC:
  for example add file "test1" in already existing directory "Archive"
  and also add "test2" in the already existing directory "Documents"
3. bring back the USB SSD and mount it on local PC, start Unison again and resccan (either via GUI or CLI)
4. Unison will report that everything is in sync and no changes has been made to either replica
5. now on LOCAL PC - create additional (empty) file on "media" in the Archive directory (let's say test3)
6. hit RESCAN on Unison GUI (or start it again via CLI):
7. Unison will now detect the "test3" file correctly as well as "test1" file ...but "test2" remains undetected
8. after clicking GO - it will correctly copy test1 and test3 to the replica in "home".

# Expected behavior vs actual behavior

Expected behavior - Unison should be able to detect files modified to removable drives on external PCs

Additionally:
- if i test using above steps on very small replica set (just 2 directories with few empty files) - SOMETIMES Unison detects them and sometimes it doesn't. (but if it doesn't - then running additional rescans will not change anything - files remain undetected)
[...]

Issue #1130, opened 2025-06-05
Add fastcheckdir preference (workaround for Windows bug)


gdt:
# Motivation and Background

Multiple users report that at least on ExFAT, Windows systems do not update the directory mod time when adding a file, more or less.   See #1122 for lengthy discussion, and pointers to standards.  So far, there are no reports of other operating systems having this bug.

With fastcheck true, a directory with the same mod time is presumed unchanged -- not surprising, given the specifications.  Thus, people running unison with an ExFAT filesystem (and maybe FAT32) that was written by Windows would want to disable at least this part of fastcheck.  Test reports indicate that fastcheck=false results in correct synchronization.   However, disabling fastcheck is a severe performance penalty, and even just disabling it on directories appears significant.

Currently, directory fastchecking is disabled on Windows.  This seems incorrect, without evidence that Window's NTFS and other filesystems (including CIFS hosted on Unix) have the same bug.

This is marked `windows` because while people will want to use the feature on non-Windows, it is fundamentally about accomodating Windows.

# Scope

The proposed behavior has been crafted to be useful to those synchronizing files on exFAT that have been written by Windows, regardless of the operating system being used for unison, while not imposing performance penalties on anyone not suffering from the not-updating-timestamp bug.    The scope includes removing a previous workaround because the new workaround is sufficient and more precise.

# Will you write code?

I am not going to write this; it solves a problem I don't have.   I've created the feature request to be able to close a lengthy issue.  Thus, this awaits a PR.

# Operation of the new feature

Add (and document) a new preference fastcheckdir, defaulting to true.    If fastcheck is false, then all fastcheck optimizations will continue to be omitted, just as now.   Assuming fastcheck true, fastcheckdir will control the optimization for directories.  Thus, setting fastcheckdir=false will skip `update.ml:dirContentsClearlyUnchanged`.

Remove the automatic disabling of fastcheckdir on !unix.  Alternatively, document behavior of Windows on other kinds of filesystems also (including remote CIFS hosted on Unix and NFS) if there is a case that fastcheckdir never works.  In that case, add finer-grained checks for enabling the workaround.

# Complexity vs Benefit

While new prefererences are basically bad, this isn't confusing and will benefit a lot of users.

Artur2048:
I've created fork, where this issue is fixed: https://github.com/Artur2048/unison
Pull Request: https://github.com/bcpierce00/unison/pull/1168

PR #1168, opened 2025-11-30
Fix bug #1122 - mishandling of ExFAT filesystem

Artur2048:
### Pull Request Description

**Problem**  
Unison running on Linux currently fails to detect file changes on ExFAT filesystems when they are modified by a different operating system than the one running Unison.  
This happens because ExFAT implementations across operating systems (Windows, cameras, etc.) do not consistently update directory modification times when files are added or changed.
Unison’s `dirFastcheck` mechanism relies on directory modification times, which works reliably only on Unix-style filesystems. On ExFAT, this leads to false “everything is in sync” reports even when the remote replica has new or modified files.

A partial workaround already exists in Unison: `dirFastcheck` is automatically disabled when Unison is running on Windows. However, no such protection exists on Linux, even though ExFAT volumes mounted under Linux are very often shared with other systems (USB drives, SD cards from cameras, external disks used by Windows machines, etc.).
(the currently existing workaround only fixes the issue if Unison is run on Windows, leaving Linux installation vulnerable)

**Solution**  
Disable `dirFastcheck` for ExFAT filesystems on all platforms, not just Windows.

This change:
- Guarantees correct change detection on ExFAT regardless of which OS is currently running Unison.
- Has almost no performance impact (`fastcheck` mechanism is still used, only `dirFastcheck` is disabled)
- Avoids exposing a new configuration option `dirFastcheck`, which would make Unison configuration unnecessarily complicated.

**Real-world impact of the bug**  
Without this fix, users syncing an ExFAT-formatted USB stick, external HDD, second partition of internal HDD, or camera SD card from a Linux machine will frequently see Unison report “no changes” after new photos or files have been added by another device or OS, leading to silent data loss in one direction.

**Testing**  
Verified on Linux with an ExFAT-formatted USB drive:
- Before the patch: adding files from Windows → Unison reports replicas in sync.
- After the patch: changes are correctly detected and propagated.

This brings ExFAT behavior in line with the existing Windows workaround and makes Unison safe to use with typical cross-platform removable media.

ref. #1122 and #1130

### DIFF
--- a/src/update.ml
+++ b/src/update.ml
@@ -2628,7 +2628,8 @@ let t1 = Unix.gettimeofday () in
       (* Directory optimization is disabled under Windows,
          as Windows does not update directory modification times
          on FAT filesystems. *)
-      dirFastCheck = useFastChecking () && Sys.unix;
+      (* A.Frycze: dirFastCheck is disabled on exFAT filesystems *)
+      dirFastCheck = useFastChecking () && Sys.unix && not (Prefs.read Globals.fatFilesystem);
       dirStamp; rescanProps; archHash = archiveHash fspath;
       showStatus = not !Trace.runningasserver }
   in


gdt:
As I read your diff, it does not implement the approach discussed in #1130 -- but your comment in #1130 says that the issue is 'fixed", in a way that I intended to convey was not acceptable.  In #1122 I said "without impairing performance for those with non-defective systems".

So far, it seems no one has reported this bug to MS.

If you think I'm confused in reading your patch, and that it really does meet the intent of #1130, please explain.

If you think I'm not confused in understanding what your PR does, and that we simply have a disagreement about which parties in the software world should be accomodated at the expense of which other parties, that's something else.  For me it is a fundamental philosophical point that Free Software users should not have any negatives imposed on them because of problems with proprietary software.

(Of course, you are welcome to run whatever modified code you wish, to distribute it under unison's license, and to point it out to others on the users list.   I would ask only that it be clear to recipients that it's different.)

PR #1170, opened 2026-01-15
wip: internationalization: languages module (PoC)

kkrolczyk:
This is initial translation framework idea.

See src/languages/README.md for details, TODOs, known BUGS ect. 

See discussion on mailing list here: https://lists.seas.upenn.edu/pipermail/unison-hackers/2026-January/002166.html


gdt:
I granted permission for CI to run (which is withheld for new PRs, or new people, idk, to avoid resources being used for spam or malicious submissions).    But that's just me pushing one button, more or less "Approve workflow run for new contributor" and then it runs as if you had never had CI denied.  So there's no bearing on the result, and I'm just telling you so "why did I get a failure notice days later" is less mysterious.

All builds are failing, and it looks like the CI config doesn't install ocaml-gettext.     I don't mean to demand you fix this fast, as the list discussion involves a lot of bigger-picture issues about dependencies.  Also CI is flaky, but I think this failure is real.

Also, a comment that how CI runs and how packaging is done are not necessarily the same.  The purpose of CI is to verify that a reasonable build plan works, not to check that all packagers will be ok with the situation.   Meaning more or less, that "we can build with opam whatever" is not the same as "nobody should have an issue with this".  In theory we could set up a  NetBSD CI runner with a build script that installs deps from pkgsrc instead of opam, to be aligned with packaging.  It's just that github offers very few OS choices, nobody has set up a self-hosted runner, and we haven't e.g. run NetBSD under qemu on Linux.

EDIT: I see you added ocaml-gettext to CI, but apparently it didn't work.  I hope you can debug it from the workflow run results.  Probably if you force push it will run again, or ping me via offlist email and I can re-approve.

kkrolczyk:
Hi @gdt - thanks. I will certainly look into that. Indeed it looks as if the step to actually install opam stuff was not executed, at least in one the builds. I'll go over them within a couple next days at most; I have a couple of TODO items on my list.

tleedjarv:
My recommendation is to ignore everything related to build system for now. Just have enough (hard-coded, if need be) so that CI runs and it works for you locally. We'll come back to build system in a later stage.

gdt:
I'm ok with ignoring "build system", but we need to have one of two believable stories:
  - ocaml-gettext is so ok and widespread that we can tell people who complain that they can't build unison because they don't have ocaml-gettext, to just go get ocaml-gettext and we don't think this is a bug
  - some way to build without ocaml-gettext (surely w/o i18n) and a plan for this not to be troublesome.

At this piont the the first item seems not plausible.     So if there is any kind of switch, even if it's manually editing something, that can be automated later, that's good enough. 

If somebody wants to make the case that ocaml-gettext is baked enough to be mandatory, please do so, on -hackers.
I don't find it on any BSD, homebrew, macports.
https://repology.org/project/ocaml%3Agettext/versions

tleedjarv:
I meant integration with the build system and only until the functional PoC is considered a success and desire to move forward is established; definitely must get it sorted out before anything gets merged.

> * some way to build without ocaml-gettext (surely w/o i18n) and a plan for this not to be troublesome.

I consider this a must and don't count on it being too complicated.

gdt:
> > * some way to build without ocaml-gettext (surely w/o i18n) and a plan for this not to be troublesome.
> 
> I consider this a must and don't count on it being too complicated.

It's not build-system complicated, but it bears on how the sources are written with respect to i18n, tags for strings, etc.   I am not clear on if there is existing art for building with/without.    I'm not claiming this is really hard, just that I have it on my lists of risks for this effort, even if out of not being clueful about i18n mechanisms.

tleedjarv:
> It's not build-system complicated, but it bears on how the sources are written with respect to i18n, tags for strings, etc. I am not clear on if there is existing art for building with/without. I'm not claiming this is really hard, just that I have it on my lists of risks for this effort, even if out of not being clueful about i18n mechanisms.

For now I consider it a non-issue because we have used and use a mechanism to support a wide range of compiler standard library versions where one-code-compiles-with-all-of-them would otherwise not be possible. Simplistically, the build system will replace all references to gettext with dummy stubs that pass through the strings.

kkrolczyk:
I meant to reply to Tõivo's message on the group, but i think here is also fine.
I've briefly reviewed the old branch solution in the fork. Actually it is the same approach (mostly).
It also uses gettext, but with gettext-stub (we also discussed that this is desired, due to camomile being old and possibly abandoned). The branch also makes it default if ocamlfind finds the library, I actually prefer manual trigger (env variable, so the user has a choice).

The focus and complication here is, that there are two parallel build systems used, and while one might be "more important" / widely used in the project, i think both should actually work. 

Regardless. Second issue which i haven't yet figured out is how to provide one entry point, "same name" which, depending on library inclusion or not - would either translate or do fallback substitution. Modules/libraries and their naming in OCaml is giving me some hard time there, and i need to postpone digging into it for a couple days, till i have some more time.

I think, given that makefile "seems" to work (i'd be happy if someone found some time to test on their end too), because control over what is linked and what not is pretty much manual, and thus easy. With Dune... well, after a couple hours I had to go for a walk, nuff said ;)

> ocaml-gettext is so ok and widespread that we can tell people who complain 

I remember i had it planned to do small research if anything in OCaml environment actually does i18n and how, if so. But the switch is there - if it doesn't work - i messed sth up :)   

**but as far as i can tell some builds succeeded and mind you - i had forgotten to actually include in CI the switch - meaning add a matrix variant or sth with `GETTEXT_ENABLED=1`, so everything what build - build because switch worked (not including anything gettext-related), and those which failed - failed due to me adding some dependencies to opam's CI - and those deps didn't manage to install** 
(or so i think, after a brief look. Will have another in a couple days)

One another fail [build (ubuntu-22.04, 4.14.x)](https://github.com/bcpierce00/unison/actions/runs/21129950728/job/60791156663?pr=1170#logs) - is silly, i added `$(MAKEFLAGS)` to the call, to be able to pass `make GETTEXT_ENABLED=1` (instead of passing it as an env arg. It might not really be needed, but anyways this broke the build, because for some funny reason it passes "w" (how, why, no clue)...


I had done some comparisons like

- my-os-ocaml-package 5.3 and os pkgs, they could build with upstream gettext version 5.3, failed to do some linking (at which point i didn't dig deeper)
- 2x Opam switches 4.0.8 (i think? or 4.12?) - worked, and the 5.3 switch also had troubles but surprisingly different than above. But i think it might be possible.

tleedjarv:
> Second issue which i haven't yet figured out is how to provide one entry point, "same name" which, depending on library inclusion or not - would either translate or do fallback substitution. Modules/libraries and their naming in OCaml is giving me some hard time there, and i need to postpone digging into it for a couple days, till i have some more time.



> added `$(MAKEFLAGS)` to the call, to be able to pass `make GETTEXT_ENABLED=1` (instead of passing it as an env arg. It might not really be needed, but anyways this broke the build, because for some funny reason it passes "w" (how, why, no clue)...

This is why I recommend ignoring integration with the build system for the moment and focus on the functional PoC. If that works out and is deemed feasible, if the translations workflow is figured out and interest from translators is there, only then should we move on to making the builds work on all platforms, in all environments and with our supported compiler versions. This is somewhat involved and I will help with it (especially when it comes to non-Linux platforms).
(Also, adding `$(MAKEFLAGS)` is not needed.)

Speaking of compiler versions, we support >= 4.08 and I noticed that ocaml-gettext now requires >= 4.14 (seemingly because dune requires it?). This by itself does not have to be a problem (we just disable i18n support with older compilers) but it's something to keep in mind (it will complicate setting up opam environment for CI).

tleedjarv:
Before starting any reviews, could I ask you to clearly separate (by subdirs or by some other way) files that are related to runtime (what is needed by the user of the program to compile and run with i18n support; for example, files like lang_bin.ml and lang_fallback.ml) from files that are only required for development and translating (so anything related to string extraction, po file compilation, what have you; also anything that could be safely left out of a tarball)? Or, you could also say that it's not easy to do or the request makes no sense because there is no such clear separation.

tleedjarv:
I took some time to experiment with build system integration. The result I think is good enough to ignore build related questions for now and focus purely on the functional PoC, including string extraction, translation and runtime. The experiment is at https://github.com/tleedjarv/unison/tree/refs/heads/gettext-test @kkrolczyk if it works for you then you can base your PoC on it. If it doesn't then let me know what fails.

Gettext functions are automatically in scope for all modules and can be used directly.

**Notes**

1. I enabled gettext for some CI builds and not for others to demonstrate that the same code works for both cases. Even though in CI opam and ocamlfind are used, the builds work without opam and ocamlfind, too.

2. I also had to pin ocaml-gettext to 0.4.2 because the current latest release (0.5.0) is broken due to dune-sites and even though the issue has been fixed upstream for almost a year now, there is no new release.

3. The build failures are currently as follows:
   - dune. I didn't even try to make dune work.
   - OCaml 5. ocaml-gettext 0.4.2 does not support OCaml 5.
   - macOS on ARM. ocaml-gettext appears to be broken on macOS ARM.
   - macOS on Intel. TUI and GTK GUI seem to work, native UI fails with error "Undefined symbols for architecture x86_64:" (list of gettext stubs). Not sure what's going on here.
   - Native Windows builds. ocaml-gettext is broken on Windows.

4. The fact that so many builds are broken is not really an issue going forward. Gettext builds must be explicitly requested by the user and _the user is reponsible for having ocaml-gettext properly installed_.

kkrolczyk:
@tleedjarv  - sorry - i didn't have the time to reply sooner. I actually changed quite a bit of logic which might have been confusing and about which you asked msg before. It was my futile attempt to fight with dune, and have one dune invocation building conditionally. Frankly, i've tested like tens of ideas, mostly sane, but insane ones too = like for example have just rule, running shell script, running copy & replace and then running matrioshka dune build. However dune sandbox thwarted most of those sandbox escape attempts. For a good reason i think anyways, although maybe there is a way to modify a src dir between calls.


But anyways - so far i verified that:
1. builds without gt work
2. gettext work in make and dune with camomile (that is - builds, runs, and translates)
3. build with gettext-stub builds only in dune - and even though unison runs, it does not translate, i'm guessing lib is not found just stubs or god knows.

PS.i came up with somewhat already sophisticated "test script" ;) in form of `.justfile` due to the fact i wanted to test "sys-libs" / opam env 4 / opam env 5, camomile and stubs. Doesn't really make sense to include it here in the commit, but i guess i can push to my fork on some spam branch. This shows that there is hope for ocaml 5 build with upstream gettext (which partially addresses concerns you mentioned in the last msg)

PS 2. Thanks for the commit on fork - i will need some more time to look at it as it is something more than just enabling CI build (i took a quick look, hoping it was just a quick patch i can pull into tree, but unfortunately no)

tleedjarv:
> Thanks for the commit on fork - i will need some more time to look at it as it is something more than just enabling CI build (i took a quick look, hoping it was just a quick patch i can pull into tree, but unfortunately no)

Sure, take your time. The code in that branch is complete replacement for your code in the makefiles and lang_entry files. Your current code couldn't be merged because it is not portable (our makefiles work with GNU Make (also the ancient version provided by macOS!), BSD make, Solaris make and NMAKE; that's why the changes to support something seemingly simple become a bit involved). I also added an interface to the gettext modules to limit what is brought into the global namespace when opening the modules; yet another bit of complexity.

kkrolczyk:
> code in that branch is complete replacement for your code in the makefiles and lang_entry files

@tleedjarv fine by me! I don't have any pressure to push my solution, as long as the translations (or framework & means to do them) are there - i'm happy, and the non-English-speaking person i was doing this for, will be as well. I guess this makes this PR obsolete, but I'll wait until you finalize merge of your stuff before closing this PR, just in case you needed anything.

tleedjarv:
Well, no, it does not make this PR obsolete and I don't intend to open a PR from my branch. There is still a lot to do and you will get full credit when the code is merged. I wanted you to ignore integration with the build system and focus on everything else for the reason that making the integration portable is not a task that's immediately obvious and can't even be completed without testing in all supported environments.

The way to look at my branch is as a review feedback on following:
 * minimal _portable_ makefile integration for gettext _runtime only_ (and fallback when no gettext available);
 * automatically opening the Gettext module for all other modules (this is strictly not necessary but at the moment I think it is desired over explicitly opening within every module; maybe we need to rethink this if it makes building with dune difficult (but it shouldn't, really)).

What is not included in my branch:
 * support for Camomile (not sure if we even want to have that);
 * marking strings for translation;
 * string extraction (this does not have to be portable and can work with GNU Make only as long as it does not touch the main makefiles in a non-portable way);
 * managing translations/new languages (not sure if any code is needed or it's going to be separate workflow/conventions only);
 * compilation of translation files;
 * installation of translations (this should be somewhat portable);
 * everything else I can't think of at the moment.

kkrolczyk:
@tleedjarv indeed, there would be more to cover whole scope. I'm not worried about the credit, that's not my motivation. I wonder if we could sketch the rough plan, how to do it. By that i mean:

 - perhaps one single PR is not too efficient to cram everything into, splitting work into smaller chunks is definitely easier for reviewers
 - can we address the questions in a structured way. I think there were some, last for example: "do we want to have camomile support or not"

My understanding, and my soft preference is yes. It adds little burden to the setup, and leaves the user to decide what do they want.
I'm in favor of respecting user's choice, if this doesn't complicate the code too much.

Perhaps collecting the questions like above in a doc somewhere helps tracking. We could utilize either text (i had readme with todo section for that), try to use github's built-in project tasks list, or free style text in comments / mails in news - harder to do sync, pun intended ;)

With regards to integration; I am sure you have reasons, not willing to open the PR and merge this separately, or simply it's about free time & processing powers; so I see two paths moving forward.

1. some effort required, to clean the fork's branch and the code. If it's ready, tested/documented the way you'd like - then it could be used as a rebase origin. Otherwise it is quite difficult to make assumptions about someone else's code - from my perspective it would be kind of like building a PoC based on a PoC, there would be quite a lot to change, and with no clear knowledge if the provided base is stable, or will undergo changes, given it's not merged. One also has to understand the changes fully, in order to be able to base on them.

2. if you want, I can apply the changes you did to mine PR, which i can easily reason about, and adjust my testing framework easily too; if you prefer option (1) that's fine, just will take me some more time.

tleedjarv:
I don't think there is a shortcut here, it will take time either way. My experimenting in the separate branch was in preparation to reviewing this PR. It quickly turned into a working and portable setup, which is a good thing in my view. Instead of getting to a similar result bit by bit, comment by comment, I consider it a drop-in replacement and spare you from having to deal with issues you'd possibly otherwise have no interest in. With or without the experimental branch, I would still request changes in your PR until I can recommend merging the result. Unfortunately, this is not going to be a fast process (not saying anything about this PR, but some PRs take years before being merged). It is just the reality of available resources, not lack of interest.

In general, I am fine with either (1) or (2), I don't think the speed of overall progress is going to differ much. It's just that with (1), you have the option of not concerning yourself with questions you otherwise wouldn't want to. (I'm pretty sure I'd continue experimenting in my branch anyway, just to aid in my review of your work, so I'm not going to spare that effort.)

For now, I need some more time (I hope it'll be quicker but it may take a few weeks) so that I can revisit all currently existing work and information, including the readme file and the emails, to form an opinion of what needs answered/solved. It's perfectly fine if you keep pushing updates, should you continue your work meanwhile.

kkrolczyk:
In working dir (not yet pushed) i have solution which:

builds both:
- `make`
- `dune`
for both
- `ocaml 4.12` and
- `ocaml 5.3.0` (i presume, that it would build for ocaml 4.02 too, 4.12 was simply chosen... well i don't even remember how, let's consider it a lucky dice roll ;)
NEARLY for both
- camomile
- gettext-stub.

Only version that doesn't build is `make` + `camomile` + `5.3.0` due to `dune-sites`.
Also, `make` + `gettext-stub` + `5.3.0` requires pinning `gettext` to master, `opam pin gettext https://github.com/gildor478/ocaml-gettext.git#master` - it seems that the fix was committed, but release wasn't made, so, yeah. Dune build on the other hand, worked even with unpinned opam gettext upstream.

Anyways - given what i see, I changed my mind, and agree that there is little point in supporting camomile. While it is nice - by moving dependencies from system to opam, but supporting both in makefile bloates it unnecessarily for 3 variants (camomile.0.4.2 + camomile.0.5.0 + gettext-stubs).

There is a small caveat, as usual.
With `5.3.0` search paths for `unison.mo` files are kinda broad (which is good, various distros have various paths)
With `4.x` it seems that only place it looks for the file is inside `/usr/local/share/locale/**` which kind of sucks, given that not all distributions even create this directory. During install one could create it of course, but I'm not very happy with it. I think, given that with camomile it searches way more paths - this behavior must be coded inside `gettext-stubs.0.4.2`.

Solution is compatible with totally disabling gettext too. It boils down to running 
`./src/dune.prepare` (both for make/dune builds) in case gettext should be enabled for build (by default it is kept disabled)

I'll try to push it tomorrow (with partially adopted changes for CI and compatibility), given that I'll be afk for a week or so.

tleedjarv:
I haven't had a chance to properly look at the pushed changes yet. I wanted to note down some of my thoughts and findings; just ignore if they're no longer relevant in light of your progress.

I propose we don't need Camomile at all, as it's only used for charset conversion. If we don't do charset conversion then we can use the pure OCaml gettext implementation _without Camomile_ (I've tried it, it works). We can eliminate the need for charset conversion under following assumptions: a) the input (PO and MO files) is always UTF-8 -- I'd say this is true regardless, but in any case it is under our full control; and b) the output is always UTF-8 -- with GTK using UTF-8 anyway, this is already covered for GUI. This only leaves the TUI. I'm willing to go out on a limb and say that in 2026 you're highly likely to have set up your console/terminal with UTF-8 (+ in Windows we already do it programmatically since many years).

There is a bug in ocaml-gettext < 0.5.0 which breaks translation of format strings. This leaves us with only the undesirable option of having to patch ocaml-gettext (either fix 0.4.2 for format strings, or patch 0.5.x to compile with older compilers -- in my understanding the compiler requirement comes from dune and not from ocaml-gettext itself).

> With `5.3.0` search paths for `unison.mo` files are kinda broad (which is good, various distros have various paths)

The pure-OCaml implementation, unlike system gettext, uses a list of search paths. The native gettext does not have this feature, this is not related to ocaml-gettext. This is arguably not a useful feature to us, since we don't want to pick up files from random locations and we don't want to install them to random locations either (and we don't depend on third party OCaml libraries installing translations to random locations).

> With `4.x` it seems that only place it looks for the file is inside `/usr/local/share/locale/**` which kind of sucks, given that not all distributions even create this directory. During install one could create it of course, but I'm not very happy with it. I think, given that with camomile it searches way more paths - this behavior must be coded inside `gettext-stubs.0.4.2`.

As you've already identified, translation not working with gettext-stubs is caused by a bad default in ocaml-gettext, but this is merely a configuration issue and should be fixed by overriding the correct directory in the config. This is not a direct problem with pure OCaml implementation as it uses a list of dirs, even though this feature in itself is not directly useful but can cover a list of widely used locations as fallback.

The correct directory should still be set at build time (and this will be different for portable use vs installed -- in the latter case it most likely needs to be under control of packagers anyway (for native gettext it may even work by not defining any dir (that is, remove the default in ocaml-gettext), which should then use whatever the native libc/libintl gettext uses)).

gdt:
I'm ok with expecting UTF-8, as long as a system configured with ascii or iso-latin1 that uses the C locale is ok too.   I think that  simply means that our untranslated strings remain ASCII.   Avoiding an unmaintained ocaml charset library sounds like a big win.

agreed that unison should look in only one place at runtime, the correct place in the prefix where translation files are stored on that system, figured out/configured at build time.   This follows a general pattern of build-time binding for reliable operation.


Also, if it's at all reasonable to first get this working without the charset lib, then especially if it would be a sane incremental change to add the charset lib later, I think it's better to have minimal viable patch landed first and we can see how it goes.   I'm a big fan of Gilb and Graham's Evolutionary Delivery, making the smallest useful change, without being prickly about smallest.

tleedjarv:
Yes, the C locale will remain untranslated ASCII in any case.

kkrolczyk:
Hey, i just got back from my holidays - please don't worry about reviewing just now - its sill a wip, it'll likely change a few more times before it's ready. I'll just need to refresh my memory, what was it all about though :)

I feel like there is a consensus about ditching Camomile; latest commit actually did that already, so that's on-track.

I also agree that UTF-8 is a standard, and cannot think of any platforms where Unison might be used which are not following it. Also worth noting, while "po" files are/would be indeed text encoded utf-8, resulting "mo" file is a binary file. BOM in utf-8 text ("po") files is not required, not recommended, and as far as i've seen - usually not used. Not sure if i understand you fully: the consideration is about systems with no utf-8 support, configured to use C locale, trying to build unison locally from sources, thus processing utf-8 text "po" file to binary "mo" file? I don't see any dangers here, but i might be missing something.

> The pure-OCaml implementation, unlike system gettext, uses a list of search paths. The native gettext does not have this feature, this is not related to ocaml-gettext. 

> agreed that unison should look in only one place at runtime

I understand that you'd like to rely fully on ocaml-gettext binary to do the po-mo conversion, so not to use native C lib at all? I might be remembering stuff wrong after the pause, but I think this is a bit different? I haven't yet checked if pure OCaml binary is doing all the work on its own, or does it explicitly call to native lib too, but from a quick glance it seems to call native lib always. So if i get it right - it's not the unison searching for the mo file (it's not directly controllable), but gettext C lib. 
About the second issue, file lookup - maybe I remember it also wrong, i thought it was something like this:
`ocaml-gettext` package would use C library (gettext) through `ocaml-gettext-stubs` to read "mo" file, but also provides binary `ocaml-gettext` which is - according to my understanding - also using native C lib to process "po" to "mo" files, so kind of a replacement for typically available binaries used to convert "po" to "mo".
So I'm not sure, how could one instruct the dependency - gettext native lib - to search different paths without recompiling it? Maybe I'm misunderstanding some things here.

Overall this does seem somewhat chicken and egg problem. If some distros are not following FHS, and for sure some do not, as some are not even Linux ;) should compiled mo files location be left to the system, to install wherever it thinks it needs to be? It does sound better than forcing creation of a root based directory tree, just to keep one file, but - like above, I'm not sure if it is possible to configure pure OCaml gettext wrapper (wrapping native gettext lib) to search path the file was installed to, in runtime. I'll need to do some experiments and see.

To reiterate - there are two steps here: one during build when "po" is compiled to "mo" and installed to some path in the system, and the second is in runtime searching some path for "mo" file. While might be somewhat easier if this is the same machine doing both steps, @gdt already mentioned that it's not that simple when doing the packaging.

gdt:
> I also agree that UTF-8 is a standard, and cannot think of any platforms where Unison might be used which are not following it.  Not sure if i understand you fully: the consideration is about systems with no utf-8 support, configured to use C locale, trying to build unison locally from sources, thus processing utf-8 text "po" file to binary "mo" file?

I think in 2026 there is no need to support building or running on a *system* that lacks UTF-8 support.   But I think that running unison under the C locale should continue to work, outputting the untranslated ASCII.

This does mean the build system will have to set locale variables and not assume that the build environment locale is ok, probably.

> I understand that you'd like to rely fully on ocaml-gettext binary to do the po-mo conversion, so not to use native C lib at all? 

I don't understand well enough to be precise.

What I meant is that when unison is installed to some prefix, and then run from that prefix, it should open files from the specific place that they were put during the build, and not look in some  list of other places.    This is a general packaging thing of build-time binding of paths, so that one gets deterministic behavior, as long as unison's prefix isn't messed with incorrectly.

If unison is using an existing OS-native (or equiv)  gettext, and the way unison deals with paths with that native gettext is the same as how every other program deals with it, it's probably fine.

> Overall this does seem somewhat chicken and egg problem. If some distros are not following FHS, and for sure some do not, as some are not even Linux ;) should compiled mo files location be left to the system, to install wherever it thinks it needs to be? It does sound better than forcing creation of a root based directory tree, just to keep one file, but - like above, I'm not sure if it is possible to configure pure OCaml gettext wrapper (wrapping native gettext lib) to search path the file was installed to, in runtime. I'll need to do some experiments and see.

Distro and FHS is much too narrow scope.  This is not about Linux; it is about any reasonable mostly-POSIX system, plus having a story for WIndows.  (Actually, I'd be ok with i18n on POSIX and untranslated on windows, as better than what we have now.)

As an example, pkgsrc is typically configurd with /usr/pkg, but /opt/pkg on macOS (because SIP), or any other prefix the user wants to use, e.g. $HOME/pkg for unpriv personal builds.   mo files go in $prefix/share/locale.  glib2 has (snippet from middle!)
```
/usr/pkg/share/locale/en_CA/LC_MESSAGES/glib20.mo
/usr/pkg/share/locale/en_GB/LC_MESSAGES/glib20.mo
/usr/pkg/share/locale/eo/LC_MESSAGES/glib20.mo
/usr/pkg/share/locale/es/LC_MESSAGES/glib20.mo
/usr/pkg/share/locale/et/LC_MESSAGES/glib20.mo
/usr/pkg/share/locale/eu/LC_MESSAGES/glib20.mo
/usr/pkg/share/locale/fa/LC_MESSAGES/glib20.mo
/usr/pkg/share/locale/fi/LC_MESSAGES/glib20.mo
/usr/pkg/share/locale/fr/LC_MESSAGES/glib20.mo
/usr/pkg/share/locale/fur/LC_MESSAGES/glib20.mo
/usr/pkg/share/locale/ga/LC_MESSAGES/glib20.mo
/usr/pkg/share/locale/gd/LC_MESSAGES/glib20.mo
/usr/pkg/share/locale/gl/LC_MESSAGES/glib20.mo
/usr/pkg/share/locale/gu/LC_MESSAGES/glib20.mo
```
That's where the build puts and installs them (when configured --prefix=/usr/pkg and maybe with some other autoconf options), and that is I believe the sole place the programs/libs look.   All I meant is that random other paths should not be searched.

> To reiterate - there are two steps here: one during build when "po" is compiled to "mo" and installed to some path in the system, and the second is in runtime searching some path for "mo" file. While might be somewhat easier if this is the same machine doing both steps, @gdt already mentioned that it's not that simple when doing the packaging.

It can't be the same machine; we have to work with packaging.    This is a solved non-issue in the rest of the free software world.    `make` creates mo and `make install` puts them in place.  packaging carries bits from where install put them to the corresponding place in the system that has the package.

I think we are trying to simply do what every other program does.

kkrolczyk:
> If unison is using an existing OS-native (or equiv) gettext, and the way unison deals with paths with that native gettext is the same as how every other program deals with it, it's probably fine.

Well, yes, or - depends. Obviously if other programs install their mo file in prefix "/xyz" and gettext is configured to look there and there only (unlikely, but here to demonstrate the situation) - then one has to know what is the prefix where unison should install its translation files. If there are multiple search paths, then of course some more flexibility is given, where to install. For example the one you've given `/usr/pkg/share/locale/` - if there is a way to get during the install the default directory - then I guess this would be preferred method. 

> It can't be the same machine; we have to work with packaging. 

We seem to have misunderstanding. I think it can be the same machine, if you are the dev and want to test on your dev machine. On top of that - it has to be able to be a separate machine, so that you can prepare a package to be installed in a different system/machine.
While every other program does that, they are rarely written in OCaml ;) it adds a bit of complexity due to the ecosystem - opam and its relationship with dependencies/libraries ect. Currently we make assumption those tools are installed during build phase, and perhaps afterwards? It's a bit fuzzy, but i assume they might be part of the packaging process - I don't know how it is currently done. It's clear that the input of the packaging process should be compiled binary and all dependencies, and the output, well a package which can be installed in another machine, providing it with both unison and its dependencies.

I went through above's compilation errors, and to sum up:

1. build (ubuntu-22.04, 4.14.x, true) - looks like gettext dependency was not found. I think this one is relatively simple to fix, it appears that machines used to build did not come with gettext native - I've now added it to steps which install it.

2. Few windows builds seem to fail due to: `export OCAMLPATH := $(shell opam var lib)` - there has to be a way to make this conditional, compatible with make syntax (so i guess no "nmake-specific": `!IF`) - for now windows builds are NOT configured to build gettext at all, so if this and (1) are solved, then all builds should be green. It will not build for all architectures/systems, but that was not planned for now anyways.

I think i've fixed them (will push in a second). Also - i've applied most if not all items from  @tleedjarv 's branch. There are two large code changes there, which i simply don't understand, so i cannot commit them on his behalf, sorry. I'll check later today what else can i do next.

PS - i already see that there was a commit changing CI.yaml, so I'll rebase and resolve this conflict later today

gdt:
Thanks for the comments - I think on packaging we agree.   As for where to install -- setting ocaml being odd aside -- when you build/install a program, you specify a prefix, and you can specify paths for various things, defaulting to the normal place within the prefix.    For any given operating system, there is one normal place per prefix, and we should line up with that.    I do understand that ocaml thinks it is its own world, but the interface of how files are laid out is a convention that should be met.     I'm not super clear on the details, just that in pkgsrc there are a vast number of packages that install po files, and nobody talks about where, so I gather it is settled and not exciting - which is good!

tleedjarv:
@kkrolczyk sorry, I have not been able to focus on this, so you'll have to bear with me for a while longer.

kkrolczyk:
@gdt - ok, one could consider specifying this prefix. This would mean, that one cannot just simply "make" anymore, and would be required to pass a prefix (when willing to build an internationalized version. I guess it could be still paramless if "English+no support for gettext" variant was built by simple 'make' invocation), so at least one - but in practice several parameters more, as opposed to simple default build. Would that be the preference?

Alternatively, either such prefix (or list of different prefixes) could be hardcoded somewhere in build-system files, and maintained for different supported systems, or it would need to be dynamically resolved during the build. Both cases have pros and cons. For hardcoding approach - I don't know if such list exists. While it is possible that "for any given operating system, there is one normal place per prefix, and we should line up with that" - but it would require collecting them beforehand, I suppose. For the second, dynamic resolution - I don't know if this is feasible.

I'll try to dig a bit into pkgsrc you mentioned - maybe one can gain some insights, but I guess we agree that most Linux distributions not only have their own package manager (so is not using pkgsrc binary) but also their own repositories, and often patch the deliverables before offering them to users.

gdt:
I think what's happening is that *in general* the concept of "just type make", while appealing, is unworkable.   Programs that open their own data files, write to temporary storage system wide, link with their own libraries, etc. need to have the prefix baked in.   It's somewhere between lucky and a legacy of past limiting choices because of the build system that unison can be built this way.

In my view, build systems need to have the equivalent of autoconf's `--prefix=/usr/foo`.   For unison, it might be that the env var PREFIX is set, and it defautls to /usr/local if not.    I'd rather just tell people they need to set it than bend over backwards to accomodate not setting it.

There can't be a list of prefixes based on OS, because the prefix is where the user (builder) intends to install it.  Yes, there's a normal place for the native packaging system (/usr on GNU/Linux, /usr/pkg on NetBSD, /usr/local on FreeBSD), but any particular build should perhaps be someplace else.     What I meant is that on most operating systems, once you are given a prefix, there seems to be a normal place within that prefix for po files.  On my system, there is
```
> l /usr/pkg/share/locale/fr/LC_MESSAGES/l*
-rw-r--r--  1 root  wheel    1138 Mar 15 22:00 /usr/pkg/share/locale/fr/LC_MESSAGES/lasem-0.6.mo
-rw-r--r--  1 root  wheel    9855 Mar 15 22:50 /usr/pkg/share/locale/fr/LC_MESSAGES/libadwaita.mo
-rw-r--r--  1 root  wheel  125655 Sep 19  2025 /usr/pkg/share/locale/fr/LC_MESSAGES/libexif-12.mo
-rw-r--r--  1 root  wheel   37883 Mar 13 14:55 /usr/pkg/share/locale/fr/LC_MESSAGES/libextractor.mo
-rw-r--r--  1 root  wheel    1409 Mar 15 21:26 /usr/pkg/share/locale/fr/LC_MESSAGES/libgnome-keyring.mo
-rw-r--r--  1 root  wheel   22649 Feb 28 20:55 /usr/pkg/share/locale/fr/LC_MESSAGES/libgpg-error.mo
-rw-r--r--  1 root  wheel   14255 Mar 15 21:58 /usr/pkg/share/locale/fr/LC_MESSAGES/libgsf.mo
-rw-r--r--  1 root  wheel   18769 Mar  1 10:25 /usr/pkg/share/locale/fr/LC_MESSAGES/libhandy.mo
-rw-r--r--  1 root  wheel    8950 Sep 20  2025 /usr/pkg/share/locale/fr/LC_MESSAGES/libidn.mo
-rw-r--r--  1 root  wheel    6812 Oct 21 08:54 /usr/pkg/share/locale/fr/LC_MESSAGES/libidn2.mo
-rw-r--r--  1 root  wheel   35934 Feb 28 23:43 /usr/pkg/share/locale/fr/LC_MESSAGES/libpq5-16.mo
-rw-r--r--  1 root  wheel    1709 Mar 15 22:34 /usr/pkg/share/locale/fr/LC_MESSAGES/libsecret.mo
-rw-r--r--  1 root  wheel    4483 Mar 22 12:54 /usr/pkg/share/locale/fr/LC_MESSAGES/libsoup-3.0.mo
-rw-r--r--  1 root  wheel    4620 Mar 15 22:51 /usr/pkg/share/locale/fr/LC_MESSAGES/libsoup.mo
-rw-r--r--  1 root  wheel   21344 Mar 15 22:03 /usr/pkg/share/locale/fr/LC_MESSAGES/libwnck-3.0.mo
-rw-r--r--  1 root  wheel   20543 Mar 15 22:32 /usr/pkg/share/locale/fr/LC_MESSAGES/libxfce4ui.mo
-rw-r--r--  1 root  wheel    6752 Mar 15 22:07 /usr/pkg/share/locale/fr/LC_MESSAGES/libxfce4util.mo
-rw-r--r--  1 root  wheel    1919 Mar 15 22:33 /usr/pkg/share/locale/fr/LC_MESSAGES/libxfce4windowing.mo
-rw-r--r--  1 root  wheel  149096 Feb 10 13:14 /usr/pkg/share/locale/fr/LC_MESSAGES/lynx.mo
```


I agree that all packaging systems patch things to make them fit, or call configure programs with args or env vars.  What we as an upstream should be doing is making that easy.  The wrinkle is that I'm not expert at i18n.!

gdt:
Unison already has prefix/destdir support.  See top-level Makefile and src/make_tools.ml

tleedjarv:
I've now spent more time on this. What you have here in the WIP is still not portable and overall very early in the process, seems to me. It's very hard for me to have any constructive feedback and it's more like flying blind. So instead, I went back to my own experiment, to know what works, what doesn't and what are the remaining issues. Seeing as it was basically complete, I went ahead and polished it to something reviewable and mergeable. See https://github.com/tleedjarv/unison/tree/refs/heads/gettext-test
I invite both of you to review what I've done in that branch. Given the state of that branch, I'm now also ready to finish it to a pushable PR state, should you want me to.

I tried to split all the changes into somewhat small commits that could be reviewed independently. This is what I recommend, too - view commit by commit, otherwise it will quickly become overwhelming. All commits that could be merged have some description in the commit message.

Comments and caveats:
 - You can ignore a) the very first commit; it sets the stage and is not directly related to gettext; b) the dune upgrade commit; c) the commit moving Windows GTK code around in GHA CI.
 - I've included a selection of pure OCaml gettext backend (w/o Camomile, so UTF-8 only) or syslib gettext backend.
 - There is no automatic detection of gettext (I don't think we should have it); the user/builder must opt in to building with gettext.
 - There is search logic for translations (MO files) for both backends. The primary search dir is `$(dirname executable)/../share/locale` followed by defaults configured in gettext (usually /usr/share/locale and /usr/local/share/locale). This should work in all situations, including on Windows and from any location (e.g. run from a USB stick, if you will). No build-time or run-time configuration needed.
 - Builds work with make and with dune, on all platforms and with all make implementations
 - dune files required a newer version (due to use of env vars in `enabled_if`).
 - I intended the POT file to be updated only by Unison developers, and committed to the repository.
 - The translators only update/commit PO files. Due to this, the po/Makefile is very simple and could be simplified further (by reducing po/POTFILES to a make dependencies file).
   - They only need to `make lang_code.po` to merge with the template.
   - To start a new language, copy `unison.pot` to `lang_code.po` and add it to repository.
   - `make translations` in the toplevel or in `src/` to build the MO files.
 - ocaml-gettext is unfortunately not suitable for string extraction; the implementation is just not complete enough. Must use GNU gettext instead.
 - The `LINGUAS` env var is considered when building and installing translations.
 - ocaml-gettext includes a few command line arguments that could be used. I've decided to parse those but not display in the usage message nor include in any documentation. I think these arguments are only useful for debugging purposes and should never be needed by users.
 - I also marked `strings.ml` for translation but this is not optimal. Ideally, the manual (TeX) would be translated instead. How to do that?
   - Also not translated:
     - `.desktop` file. Nothing special needed - translators edit directly.
     - The man page. Not sure if and how man pages are translated.
 - The two commits marked DEBUG are to be removed as soon as you've convinced yourselves that correct backends are built and linked in and translations work.
 - In the CI, I've tripled all GHA builds for now. This will not remain so, of course.
   - ubuntu-24 builds work but seem to not display translated strings (unlike ubuntu-22) - this is a Ubuntu/GHA issue. When downloading the build artifacts, everything works correctly.
   - There is only one failing build: mingw 32-bit with the syslib backend. The local selftest works, selftest with a server crashes with no meaningful error message. I tested the build artifacts in a Windows VM and everything works correctly.
   - **To make the GHA builds work reasonably, I had to use the upstream unreleased ocaml-gettext version and patch it for compiler support and bugs.** This means that it could be a bit of a struggle for packagers/users building themselves.
 - Almost forgot the biggest caveat of them all: OCaml format strings do not support out-of-order arguments. This likely means that many format strings are not easy, and could be impossible, to translate to various languages.

I did some _very crude_ performance testing. I used ocaml-gettext 0.4.2 where translation of format strings is broken; not sure if it has any impact on the results. The results may no longer apply with the latest upstream ocaml-gettext. The test was simply a listing of 400 000 updates.

* The dummy backend (no translations, but with marked strings) was ~1-3% slower than 2.54.0.
* The pure OCaml backend
  * without translations: ~3-5 % overhead
  * with translations: ~15% overhead (13-17)
* The syslib backend: ~15% overhead (13-17..20)
  * without translations _slightly_ faster, no significant difference

Considering that Unison is normally primarily I/O- and CPU-bound, I think these results are acceptable. Even with 400 000 updates, as in this test, we are talking about milliseconds up to 1..1.5 seconds worst case.

I also measured the sizes of built binaries.

* The binary sizes are ~the same with the dummy implementation (slight increases were offset by slight reductions that resulted from changes when marking strings).
* With the pure OCaml backend, binaries are ~600-700 kB larger (unstripped), which means +6-10% size increase, and +18% for fsmonitor (+28% stripped).
* With the syslib backend, binaries are ~200-300 kB larger (unstripped), or +3-5% size increase (+10% for fsmonitor).

kkrolczyk:
@tleedjarv from your comment i understand that you consider the version you have on your branch superior to this one. I recon it is for some people harder to collaborate and instead they prefer to write something of their own - I can understand that; In this case I care only about actual technical capability which was (and at the time I'm writing it - still is) missing from the software. 

I will deprecate this PoC in such case, and hopefully feature will someday become available - from what you write - you consider it ready, or almost ready.
As for review, I unfortunately do not see this feasible, for several reasons mentioned in above, so I'll abstain from repeating myself here.

PR #1187, opened 2026-04-17
Fix crash when current directory is passed as first argument

### LINKED ISSUE #1186: Uncaught exception `Sys_error("Is a directory")`
I have thoroughly read the [Reporting Bugs and Feature Requests](https://github.com/bcpierce00/unison/wiki/Reporting-Bugs-and-Feature-Requests) wiki page.

#Environment

- Unison version: 2.53.8
- OS: Arch Linux x86_64 (6.19.11-arch1-1)
- CPU: AMD Ryzen 5 4500
- OCaml: 5.4.0
- UI: TUI
- fsmonitor: no

# Reproduction Recipe

Just run a command without any options/flags and with tree arguments, and the current directory (`.`) must be the first argument. 

# Expected behavior vs actual behavior

Expected behavior: The cli should print the usage or help text in TUI.

Actual behavior: there is an unexpected error, for example command `unison . ~ /`:

```plaintext
Error: Uncaught exception (please report a bug): Sys_error("Is a directory")
Raised by primitive operation at Stdlib.input_line.scan in file "stdlib.ml", line 453, characters 12-32
Called from Prefs.readAFile.loop in file "/build/unison/src/unison-2.53.8/src/ubase/prefs.ml", line 385, characters 19-36
Called from Prefs.readAFile in file "/build/unison/src/unison-2.53.8/src/ubase/prefs.ml", line 413, characters 10-24
Re-raised at Prefs.readAFile in file "/build/unison/src/unison-2.53.8/src/ubase/prefs.ml", line 413, characters 56-63
Called from Prefs.loadTheFile in file "/build/unison/src/unison-2.53.8/src/ubase/prefs.ml", line 490, characters 27-40
Called from Uicommon.initPrefs in file "/build/unison/src/unison-2.53.8/src/uicommon.ml", line 1037, characters 5-24
Called from Uitext.Body.start in file "/build/unison/src/unison-2.53.8/src/uitext.ml", line 1674, characters 4-78
```

But when run command `unison <not_dot> <arg2> <arg3>`, cli printing the usage guide.

P.S: I created current report cause in error was a sentence `please report a bug` and I thought I can help somehow :).

Fakhriddin3040:
When a directory (e.g. `.`) is passed as the first argument, Unison may treat it as a profile name and attempt to read it
as a preferences file, leading to:

`Sys_error("Is a directory")`

This PR:
- prevents directories from being treated as preference files
- improves CLI argument handling consistency

Now:
- multiple positional arguments are treated as roots
- directories are no longer passed to readAFile

If the first argument is `.`, it is no longer interpreted as a profile and is handled as a root path.

Fixes #1186

### DIFF
--- a/src/ubase/prefs.ml
+++ b/src/ubase/prefs.ml
@@ -490,8 +490,17 @@ let processLines lines =
 
 let loadTheFile () =
   match !profileName with
-    None -> ()
-  | Some(n) -> processLines(readAFile n)
+  | None -> ()
+  | Some(n) ->
+      let path = profilePathname n in
+      try
+        let stat = System.stat path in
+        if stat.Unix.LargeFile.st_kind = Unix.S_REG then
+          processLines (readAFile n)
+        else
+          ()
+      with _ ->
+        ()
 
 let loadStrings l =
   let rec loop n out = function


tleedjarv:
Thank you for the report and fix. Could you please disclose if any LLM tool was used for the PR or code?

Fakhriddin3040:
Sure. Cause I have no any experience with OCaml, I used ChatGTP only for adding a check: 

Is profile a file?

If yes, behavior the same as it was.

Otherwise, just ignore the profile.


There were two ways of behavior when the given profile is not a file:

1. Just ignore 
2. Print help or guide and exit

I choose the firts one.


And now, I asked for detailed explanation one more time for this method, and LLM sad that the function catching any error and It's wrong behavior on my opinion. 

I think, there should be only check for a file and file permission, and don't handle other kinds of error.

tleedjarv:
Thank you for clarifying. Normally the project can't accept LLM-generated contributions. Since it is so short, I reviewed it anyway. Unfortunately, the fix is not optimal because it is located separately from other error checks, it doubles function calls to pathname and stat (as they would be repeated immediately in the readAFile function). Catching all errors is usually not a good practice but here it is made worse by then being silent about it. I think it is better to report an error rather than just ignore and continue.

Fakhriddin3040:
Understand. Thank you too for your review. I'll learn about the Unison's architecture and OCaml basics. After will try by myself. I'll be back after few days if someone will not fix it. Thanks for attention.

tleedjarv:
I have a fix lined up. You are of course very welcome to study the code contribute any time!

Fakhriddin3040:
Okay. I hope that could help somehow 😅. Thanks for attention. Wishing you all the best.

PR #1189, opened 2026-04-25
Drop ocamlfind dependency

tleedjarv:
`ocamlfind` is another bit that is no longer well maintained. There is actually no _need_ for ocamlfind but it will still be used if it's installed and found. That way, nothing changes for repositories that needed ocamlfind for builds anyway.


gdt:
I will need to test this against pkgsrc, which builds for a prefix that isn't /usr or /usr/local, as well as really read the diff harder.  It's beyond my pre-release comfort level.

tleedjarv:
Rebased, now that CI scripts are working again.

gdt:
See also https://github.com/ocaml/ocamlfind/issues/124

PR #1197, opened 2026-06-06
Update unison-manual.tex: the `mergebatch` preference has been removed.

OnkV:
Fixes a minor inconsistency in the manual.


gdt:
Wow, a preference removed in 2.18!

Unison contains `src/strings.ml` which needs frequent regeneration.   How to do this is described in CONTRIBUTING.md in the Strings section.   You should end up with a strings.ml commit which changes just what you are changing, but it is possible locale issues will become pesky.     I'm happy to do this and add it to your commit, or you can do it if you want.

gdt:
Thanks for your contribution.  I did the strings regen and merged it.

PR #1198, opened 2026-06-22
Support mid-sync abort for external GUIs

bcourbage:
The OCaml to Cocoa bridge in uimacbridge.ml doesn't expose a way for the macOS Cocoa frontend to abort an in-progress sync. This change registers an abortAll callback that calls the existing Abort.all, so Cocoa frontends can hook into it. This change is purely additive. No behavior change for existing callers.

### DIFF
--- a/src/uimacbridge.ml
+++ b/src/uimacbridge.ml
@@ -682,6 +682,9 @@ let unisonSynchronize () =
 ;;
 Callback.register "unisonSynchronize" unisonSynchronize;;
 
+let unisonAbortAll () = Abort.all ();;
+Callback.register "abortAll" unisonAbortAll;;
+
 let unisonIgnorePath pathString =
   Uicommon.addIgnorePattern (Uicommon.ignorePath (Path.fromString pathString));;
 let unisonIgnoreExt pathString =


gdt:
I am unclear if our CI does any useful tests, so I wonder what the state of the in-tree mac gui was, and if you have run it after.   (I get it that one can reason it must not be harmed; I'm asking if we know that's true.)   I myself do not use the unison GUIs at all.

I enabled the CI run; sorry for  not hitting that earlier but I didn't realize it wouldn't run (for some definition of "new contributor", it only runs if I hit "approve running CI").

bcourbage:
Thanks for enabling CI.

I built and ran the app from this branch. Built `Unison.app` via `make macui` (OCaml 5.4.1 / Xcode 26.5): clean build, codesigns and `codesign --verify --deep --strict` passes. Launched it and ran an actual sync. Files propagate correctly, archives are written, and the app exits cleanly. The interactive reconcile window renders normally and shows queued changes as expected. The change is purely additive (registers a new `abortAll` callback; the in-tree Cocoa frontend never calls it), so existing behavior is unchanged. 

I also looked into the CI failure. Both red jobs are the Windows builds, failing at the `lablgtk install` step (building GTK/gettext from source under MSVC: `GUI library lablgtk not found`), before this PR's code is reached. This change touches only `src/uimacbridge.ml`, which compiles solely on macOS. All macOS and Linux jobs passed, including `macOS: Build and package Unison.app`. The Windows break looks pre-existing on `main` rather than introduced here.

gdt:
The CI setup is a bit flaky; jobs randomly fail to fetch packages that must be installed.   I just usually rerun them.   Thanks for testing/reporting.

gdt:
Windows CI is failing reliably, and it's not about your change.

PR #1207, opened 2026-09-03
Add dry run option

### LINKED ISSUE #1203: Add --dry-run command line option to non-interactively detect changes
Purpose: letting a script decide whether running other tools to create backup snapshots etc is useful.

Exit code 0 for nothing to do, exit code 1 for something to mirror, exit code xyz for error xyz.

tleedjarv:
This is the simplest dry run implementation possible. A new preference `dryrun` is added (I chose to name it so that it maches other preferences, but the very common `--dry-run` and `-n` also work).

Updates scanning and reconciliation are run as normal, but the synchronization step is simply skipped. The exit codes remain the same as they've always been.

All other preferences continue working as they otherwise would. In particular, this means that the UI is interactive by default. In scripts you want to combine dry run with `-silent` and/or `-batch`.

The simplest implementation I've opted for here is to simply skip the synchronization step wholesale. This means that you can see the updates that are detected, what actions are suggested by reconciliation, the pre-sync summary (in TUI) and the process exit code remains meaningful.

On the other hand, this does not simulate the syncing action for each individual path. I don't know if this would be useful or not. It is certainly possible and I even tried it for a quick PoC but I'm not sure it is worth the additional code complexity (increased maintenance burden is a slight worry, but without major safeguards it would pose the risk of some future code change accidentally ignoring the dry run restriction).

Closes #1203

### DIFF
--- a/src/uicommon.ml
+++ b/src/uicommon.ml
@@ -57,6 +57,25 @@ let auto =
      ^ "to move to the next, it will skip over all non-conflicting entries "
      ^ "and go directly to the next conflict.)" )
 
+let dryRun =
+  Prefs.createBool "dryrun" false
+    ~category:(`Basic `Syncprocess) ~local:true
+    ~send:(fun () -> false)
+    "dry run, do not change replicas"
+    "Scan and reconcile updates as normal, but do not synchronize. \
+     Allows checking for updates and seeing what actions would be taken \
+     to synchronize the replicas without making any changes in replicas \
+     (note that the archive files and cache files used by Unison may still \
+     be updated).\
+     \n\n\
+     Process exit code will signal if there were updates to synchronize. \
+     See \\sectionref{exit}{Exit Code} for more information. \
+     Dry run may be used together with \\texttt{batch} and/or \\texttt{silent} \
+     to check the exit code in scripts."
+(* Provide commonly-used alternatives *)
+let () = Prefs.alias dryRun "-dry-run"
+let () = Prefs.alias dryRun "n"
+
 (* This has to be here rather than in uigtk.ml, because it is part of what
    gets sent to the server at startup *)
 let mainWindowHeight =
@@ -701,6 +720,7 @@ let transportFinish () = Transport.logFinish ()
 
 let transportItems items pRiThisRound makeAction =
   if Abort.isAll () then () else
+  if Prefs.read dryRun then () else
   let waiter = Lwt.wait () in
   let outstanding = ref 0 in
   let starting () = incr outstanding in
--- a/src/strings.ml
+++ b/src/strings.ml
@@ -1103,6 +1103,7 @@ let docs =
       \n\
       \032 How to sync:\n\
       \032  -batch              batch mode: ask no questions at all\n\
+      \032  -dryrun             dry run, do not change replicas\n\
       \n\
       \032 How to sync (text interface (CLI) only):\n\
       \032  -auto               automatically accept default (nonconflicting) actions\n\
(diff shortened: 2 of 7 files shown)


src/uicommon.ml
```diff
@@ -57,6 +57,25 @@ let auto =
...
+     to synchronize the replicas without making any changes in replicas \
+     (note that the archive files and cache files used by Unison may still \
+     be updated).\
+     \n\n\
+     Process exit code will signal if there were updates to synchronize. \
+     See \\sectionref{exit}{Exit Code} for more information. \
+     Dry run may be used together with \\texttt{batch} and/or \\texttt{silent} \
+     to check the exit code in scripts."
+(* Provide commonly-used alternatives *)
+let () = Prefs.alias dryRun "-dry-run"
```
gdt, line 76: I would prefer not to add aliases for what I see as typos.  The proper user flow is to read the docs and use the right option!   We have too many preferences, and long term my goal is to reduce them, when it doesn't hurt (not that this is really feasible or happening).   I think `-n` is important to add, as it's the way rsync etc. spells it.

Regarding dry-run vs dryrun, I'd lean to consistency with unison's multi-word prefs.

gdt:
Scanning existing prefs, it's overwhelmingly in favor of `dryrun`, as you (unsurprisingly) picked.

The diff looks fine to me.   It's interesting that this change happens in all the UIs and not in core, but I'm guessing there's a reason that is harder or uglier.

I will hold merging, until the people that said this was important have tested it and described their test experience here.

tleedjarv:
> It's interesting that this change happens in all the UIs and not in core, but I'm guessing there's a reason that is harder or uglier.

I had some reservations about it myself. There is a common check at https://github.com/bcpierce00/unison/pull/1207/changes#diff-de63a7009add9b35cb4eef338122bb702820f34049d01f2f1037289a14d62587R723 but that simply makes the sync function a no-op. It could instead raise an exception or return an error result but that would mean larger changes and would still mean changes in all the UIs, so I decided against it. For now, at least. Maybe I'll revisit that decision. If it turns out in future that we need to improve then we'll do just that.

gdt:
This is waiting on the people that said this was important performing testing and reporting, and resolving the alias for a typo.

src/uicommon.ml
tleedjarv: My reasoning for adding aliases was to cover common spellings, even though they wouldn't be published in any of the documentation. So the usefulness is indeed pretty limited and I can easily remove them if they're seen as unnecessary. But these are intentionally not typos. I chose to have `-dryrun` as the primary because it matches the other preferences. Then there is the Unix-style `-n`, which I think is very commonly used. And finally, there's the GNU-style `--dry-run`, which I believe is also very common.

The aliases do not increase the preference count, of course, but once they're out there, clawing them back later can become nearly impossible.

If we wan't to keep `-n`  (which I think is good) then the question is, why not `--dry-run`? (Note that it is not a typo, it's prefix is two dashes, not one as all other prefs). A strong (and perhaps the only relevant) argument against all these aliases is that they wouldn't be documented.

tleedjarv:
@carlospgmat03 and @jjYBdx4IL could you grab a test build and see if this kind of dry run is what you'd expect? The build artifacts can be found at https://github.com/bcpierce00/unison/actions/runs/33742163028

carlospgmat03:
I do not know the protocol to comment on issues, but thanks for starting to deal with our comments!!! I would expect that if there is something to do, it exits with an error, without user interaction. Likewise, if there is nothing to sync, 

So, with changes:
```
/tmp/unison-unmerged-pr-1207-git_ee593409-ubuntu-22.04-x86_64/bin                  06442209
> echo $status
1
```
without any changes:
```
> ./unison -dryrun -batch -silent  work_home_fun
/tmp/unison-unmerged-pr-1207-git_ee593409-ubuntu-22.04-x86_64/bin                  06442209
> echo $status
0
```
Thanks!!!!

PR #1208, opened 2026-09-15
Fix an edge case in error path of directory copy

tleedjarv:
Under specific circumstances, the dirfastcheck mechanism can miss unsynced files.

It's a bit tricky to explain but the pre-requisites are that (1) a directory is copied, (2) dir itself is not updated, (3) some children fail the copy, and (4) the failed children were previously synced (known by the other replica). In this case, the failed, now unsynced, children will remain unsynced and will not be discovered by the next updates scan.

The fix is very simple: disable dirfastcheck for the affected directory.

This reproducer fails without the patch and works with the patch:
```sh
mkdir -p testa/sub
cat /dev/null > testa/sub/file.txt
# Sync twice to make sure all archives are up to date.
unison -batch testa testb
unison -batch testa testb

# Cause a reading error in one replica...
chmod a-r testa/sub/file.txt
# ... and remove the dir in other replica...
rm -R testb/sub
# ... and copy it back (this will partially error)
unison -batch testa testb -force testa -confirmbigdel=false

# The next run should identify that file.txt is not synced.
# Without patching, it does not!
unison -batch testa testb

# Even after removing the artifical error, the unsynced file is still missed.
chmod u+r testa/sub/file.txt
unison -batch testa testb

# Only once fastcheck is disabled (dirfastcheck specifically in this scenario),
# the missing file is discovered.
unison -batch testa testb -fastcheck=false
```


tleedjarv:
Improved wording in code comment.

